PR #1, opened 2018-06-30
docker support

sorrowhill:
개발환경 구축하기가 어려워서 만들어 봤습니다.

이미지는 2개로 나뉘는데요. 하나는 라이브러리 이미지(ome_base),
다른 하나는 코드를 받아서 빌드하는 이미지입니다.

일단은 기본 동작만 확인하고 외부 포트, 설정 파일 위치만 확인해서 
노출시켰습니다.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: required
 before_install:
-  apt-get update && apt-get install -y build-essential
-  apt-get bc nasm autoconf libtool zlib1g-dev pkg-config curl
-  chmod +x $TRAVIS_BUILD_DIR/Docker/ubuntu/ome_base/*.sh
+  - apt-get update && apt-get install -y build-essential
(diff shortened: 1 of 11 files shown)


JehoenHan:
We are very pleased to receive the first pr and thank you. We have plans to make a docker image. And we want it to be based on alpine linux because we want image size is small. If you can help to solve this problem, we would appreciate it. If it is difficult, please wait until we update the docker image. Thanks again.

JehoenHan:
And we are using c++ 14.

sorrowhill:
I can work with it, of course. since gcc needs at least 100 MB, decreasing the size of these libraries will be the biggest problem. it may affect the performance of your application. I have to close the first request anyway because the pull request comes from the master branch and I'm going to revoke them all.

JehoenHan:
I do not think gcc is needed. I think the docker image version should be built with just binaries. I know this is a difficult task. Thank you for your contribution.

PR #57, opened 2019-11-22
Bug/stream deletion crash

rubu:
PR for issue https://github.com/AirenSoft/OvenMediaEngine/issues/56.

Basically what this does it increments the iterator in the for loop to the next valid, so that the call to RemoveSession which will underneath trigger erasing the stream from the map does not invalidate it (this was not clear previously since the `for(const auto &x :` hid the underlaying iterators.

### DIFF
--- a/src/projects/webrtc/rtc_application.cpp
+++ b/src/projects/webrtc/rtc_application.cpp
@@ -59,9 +59,11 @@ bool RtcApplication::DeleteStream(std::shared_ptr<StreamInfo> info)
 	}
 
 	// 모든 Session의 연결을 종료한다.
-	for(auto const &x : stream->GetAllSessions())
+	const auto& sessions = stream->GetAllSessions();
+	for(auto it = sessions.begin(); it != sessions.end();)
 	{
-		auto session = std::static_pointer_cast<RtcSession>(x.second);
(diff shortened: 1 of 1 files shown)


getroot:
Thank you very much 👍

rubu:
No problem, I really like the feature set you guys have and I'd be interested in contributing in long term - but since I am on macOS I'd then need you guys to accept my changes for macos support, since I'll be developing on a mac. I'll open a PR for that and let's see if that stuff seems acceptable:)

src/projects/webrtc/rtc_application.cpp
```diff
@@ -59,9 +59,11 @@ bool RtcApplication::DeleteStream(std::shared_ptr<StreamInfo> info)
 	}
 
 	// 모든 Session의 연결을 종료한다.
-	for(auto const &x : stream->GetAllSessions())
+	const auto &sessions = stream->GetAllSessions();
+	for(auto it = sessions.begin(); it != sessions.end();)
 	{
-		auto session = std::static_pointer_cast<RtcSession>(x.second);
+		auto session = std::static_pointer_cast<RtcSession>(it->second);
+		it++;
```
dimiden: ```++it``` seems to be better than ```it++```

PR #59, opened 2019-12-09
Support for macOS

rubu:
So this commit contains changes so that OvenMediaEngine can be built and run on macOS. The code has been modified only as much as it was needed to compile and run (seems like there is shady stuff which gcc passes through happily, but clang does not). A quick explanation of the changes:

1) misc/prerequisites.sh: 
*)Added OS name, version and CPU core detection
*) clang does not support `-rpath=` so use `-rpath,`
*) libvpx needs to have a proper darwin target specified
*) Install the needed brew packages

2) src/core/config.mk: CPU core detection

3) Clang does not support gcc style variable length compile time  arrays so where possible use const/constexpr size initializers, where not possible use memset

4) src/projects/base/ovlibrary/byte_ordering.h: byte swap function definitions for macOS

5) src/projects/base/ovlibrary/converter.h: add `ToString` overload for size_t which was missing

6) src/projects/base/ovlibrary/dump_utilities.cpp: std::max was called with off_t and size_t, since assigning to size_t cast the first arg to size_t

7) src/projects/base/ovlibrary/log_write.cpp: sstream is used but not included, clang fails to compile

8) src/projects/base/ovlibrary/path_manager.cpp: 
*)Add __mode_t as an alias of mode_t
*) Add support for macOS in PathManager::GetAppPath

9) socket.h / socket.cpp: my implementation of epoll on top of kevent, i was unable to map some epoll
functionaluty to kevent, but added asserts if any of that is used, so far I did not encounter such scenarios while running OME

10) src/projects/base/publisher/application.cpp: this seemed like a bug since an object of type ov::String is passed to "%s" and GetName was used to print id. This seems like a correct fix

11) src/projects/config/config_manager.cpp: patch structure field name differences between macOS and Linux

12) src/projects/http_server/interceptors/web_socket/web_socket_frame.cpp: Again std::min with different types - size_t and uint64_t (clang was not happy), cast the result of Remained() to uint64_t

### DIFF
--- a/misc/prerequisites.sh
+++ b/misc/prerequisites.sh
@@ -13,9 +13,15 @@ VPX_VERSION=1.7.0
 FDKAAC_VERSION=0.1.5
 FFMPEG_VERSION=3.4.2
 
-OSNAME=$(cat /etc/*-release | grep "^NAME" | tr -d "\"" | cut -d"=" -f2)
-OSVERSION=$(cat /etc/*-release | grep ^VERSION= | tr -d "\"" | cut -d"=" -f2 | cut -d"." -f1 | awk '{print  $1}')
-NCPU=$(nproc)
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    NCPU=$(sysctl -n hw.ncpu)
+    OSNAME=$(sw_vers -productName)
+    OSVERSION=$(sw_vers -productVersion)
+else
+    NCPU=$(nproc)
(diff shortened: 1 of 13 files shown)


rubu:
Also, maybe let's not ignore tasks.json ? Because at least for macOS it is important to use gmake, and set DYLD_FALLBACK_PATH to /opt/ovenmediaengine/lib, thus having that stuff preconfigured would make sense

dimiden:
@rubu 
If you can share your tasks.json file here or elsewhere, I'll put it in the github wiki or ```misc```.

PR #67, opened 2020-01-30
Fix macOS compilation

rubu:
*) sigemptyset is a macro on macOS thus cannot be prefixed with ::
*) HOST_NAME_MAX is not available on macOS but _POSIX_HOST_NAME_MAX is
*) do not include <linux/sockios.h> on macOS


dimiden:
Thank you for the contribution!

PR #71, opened 2020-02-11
Cleanup and implementation for rtsp provider stream names

rubu:
Cleanup and implementation for rtsp provider stream names

*) bit_reader.h: remove unneeded variable
*) sdp.cpp: fail when codec id's cannot be deduced (ffmpeg will convert h264 to mpeg4 w/o -c:v copy) and for example mpeg4 is useless here since there is no decoder and neither can it be bypassed to any kind of currently supported publisher
*) other files - route domain from the actual rtsp connection to rtsp provider so that app name lookup can be performed

### DIFF
--- a/src/projects/base/ovlibrary/bit_reader.h
+++ b/src/projects/base/ovlibrary/bit_reader.h
@@ -12,8 +12,7 @@ public:
     BitReader(const uint8_t *buffer, size_t capacity) : buffer_(buffer),
         position_(buffer),
         capacity_(capacity),
-        bit_offset_(0),
-        mask_(0x80)
+        bit_offset_(0)
     {
     }
 
@@ -88,5 +87,4 @@ private:
     const uint8_t* position_;
     const size_t capacity_;
     int bit_offset_;
-    uint8_t mask_;
 };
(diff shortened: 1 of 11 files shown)


rubu:
Tested this by:
1) adding 127.0.0.1. my-amazing-stream.com to /etc/hosts
2) Adding the domain to conf :
```
<VirtualHosts>
		<VirtualHost>
			<Name>default</Name>

			<!-- Settings for multi domain and TLS -->
			<Domain>
				<Names>
					<Name>my-amazing-stream.com</Name>
				</Names>
				<!--
				<TLS>
					<CertPath>path/to/file.crt</CertPath>
					<KeyPath>path/to/file.key</KeyPath>
					<ChainCertPath>path/to/file.crt</ChainCertPath>
				</TLS>
				-->
			</Domain>

			<!-- Settings for applications -->
			<Applications>
				<Application>
					<Name>app</Name>
					<!-- Application type (live/vod) -->
					<Type>live</Type>
```
3) Pushing rtsp stream with ffmpeg to my-amazing-stream.com:
```
ffmpeg -i o.mp4  -c:v copy  -f rtsp -rtsp_transport tcp rtsp://my-amazing-stream.com:8554/app/live
```

The app name was correctly resolved to #default#app so I assume this is the proper way to do it?

PR #193, opened 2020-09-12
Add srt pkg-config in ovlibrary, dtls_srtp, monitoring and orchestrator

erdnaxe:
When only SRT and FFMpeg libs are installed in `/opt/ovenmediaengine/`, compilation fails saying that `srt/srt.h` is not found. This is due to missing SRT pkg-config call in some projects. It seems to be an unwanted oversight.

One solution to that problem is to add SRT pkg-config to `AMS.mk` when a project requires it.

Regards,


dimiden:
@erdnaxe Thank you for your contribution!

PR #212, opened 2020-11-03
Fixed typo in log_internal.cpp

OlegGavrilov:
Can't build without this fix.


dimiden:
Oh, I only tested it in Debug mode and missed it. Thank you very much!

PR #247, opened 2021-01-01
fix check if scheme of ovtStream is equals OVT.

basisbit:
The code threw an error if the scheme of an OVT stream is `OVT`, but it should only throw the error if the opposite is the case.

This replaces a `==` by `!=` and also makes the check case-insensitive.

### DIFF
--- a/src/projects/providers/ovt/ovt_stream.cpp
+++ b/src/projects/providers/ovt/ovt_stream.cpp
@@ -140,7 +140,7 @@ namespace pvd
 		}
 
 		auto scheme = _curr_url->Scheme();
-		if (scheme == "OVT")
+		if (scheme != "OVT")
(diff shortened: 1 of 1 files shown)


getroot:
Oh!
This code you found hasn't found any problems in the meantime because the internal module requests the ovt scheme in lowercase when generating the ovt url. If you modify the code as follows and make a pull request again, I will merge this code.

if (scheme.UpperCaseString() != "OVT")

basisbit:
updated the PR.

basisbit:
updated the PR.
The internal module requests the ovt scheme as specified in Origin.Pass.Scheme from the edge config xml. So if that contained `OVT`, as specified in some of the samples from the documentation, then the exception was thrown.

PR #254, opened 2021-01-15
Update Docker "Getting Started" instructions

Dave3o3:
I updated the `Getting Started` instructions for the `docker run` command and included the use of named volumes.


getroot:
Thank you

PR #257, opened 2021-01-22
Fix typo: publihser -> publisher

command-tab:
This PR fixes typos where `publisher` was misspelled as `publihser` (note the `s` and `h` transposition), both in comments and code.


getroot:
Thank you!

PR #262, opened 2021-01-30
Use tls_address for HTTPS ThumbnailPublisher

command-tab:
I believe this is the same situation as #261, where `tls_address` should have been used instead of `address`, but it's in ThumbnailPublisher instead of APIServer.


kwonkeukhan:
Thank you. really helped me a lot.

PR #297, opened 2021-02-28
Create signed_policy_url_generator.php

bchah:
Simple PHP script to allow your web server to sign OME URLs on the fly. Takes a GET request consisting of the BASE URL and optionally, a stream policy. Returns the full signed URL.


getroot:
Thanks for your contribution!

PR #307, opened 2021-03-18
Add apt-get update before every apt-get

manosamy:
apt-get update must be included with every apt-get, else, Docker build will use previous cache and will fail to get new packages correctly.
Build failed for me without it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ ENV     OME_VERSION=master \
         PCRE2_VERSION=10.35
 
 ## Install build utils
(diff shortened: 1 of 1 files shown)


getroot:
@manosamy 
Thank you very much for your PR.

But I'm curious. If I put the "apt-get update" command, does it ignore the cache and build a new one? I know docker caches images based on the same command. I am not sure about this yet. If the user wants to ignore the previous cache from a certain line, he can rebuild from that point by entering a simple command like RUN mkdir /tmp/aaa The reason you didn't use the old cache when you put an apt-get update may be because the command in that part has changed.

If the user does not want to use the old cache when doing docker build, this can be solved by giving docker build the --no-cache option.

PR #310, opened 2021-03-19
set TLS 1.2 as min version for HTTPS (HLS/DASH/API)

basisbit:
* sets TLS 1.2 as min version for HTTPS

Tested using the ssllabs.com scanner, showed no problem with access from any typically used device, resolves some potential downgrading attack vectors and allows for further cleanup of the ciphers list + performance improvements.
Edit: used this in production for an event with 2500 viewers and no problems reported so far.


basisbit:
Once this is merged, I'll create a second PR which reduces CPU requirements on typical KVM/VMWare based edge server by roughly 30% - 50%, when HLS is used.

getroot:
Thank you very much for posting your PR!

PR #313, opened 2021-03-23
Disable TLS1.3 and improve performance for TLS encrypted HTTP

basisbit:
### DIsable TLS 1.3
Disable TLS1.3 because it is not yet supported properly by the HTTP server implementation (HTTP1.1 & HTTP2 support, Session tickets, persistent connections, ...)
This also allows for using less secure cipher suites for lower CPU requirements when using HLS/DASH/LL-DASH streaming
Once the HTTP server is improved, the max TLS protocal version should be changed to TLS1_3_Version and then the new way of setting tls1.3 cipher suites should also be added, so that only fast ciphers are used.

### Improve performance for TLS encrypted HTTP
For most use cases, high performance is much more important than using only modern cipher suites. The HTTP server implementation currently does not set a preferred TLS version anyways (this gets fixed by only having one supported cipher suite) and TLS 1.3 ignores this cipher suite filter by design (it has a new function for setting tls1.3 cipher suites and adds its own default if this function is not used).

The TLS_RSA_WITH_AES_128_CBC_SHA (AES128-SHA) cipher suite is still considered not yet broken, but does not offer Perfect Forward Secrecy support because only RSA is used. This would theoretically allow an attacker to decrypt recorded transmissions, if the attacker gets the private key from the webserver later on. However, when rotating used keys typically every 3 months for example and then deleting the old private key, this risk is further reduced. 

Most important for typical use cases however is, that RSA is faster (requires less CPU) than the previously used ECDHE for initial key exchange - this is especially important because the HTTP server implementation does not yet support HTTP 1.1 persistent connections or HTTP2 and thus a new TCP+TLS connection is established for each chunk file download.

## Additional notes
In my initial tests, this reduced CPU load on typical KVM or VMWare based virtual machines with AES accelerated CPUs by roughly 30 to 50 percent when using HLS/DASH based live-streaming edge servers, which allowed for cheaper operation and better scaling of the live video streaming service.


dimiden:
@basisbit 
Thank you for your contribution!

As far as I know, TLS 1.3 is not supported in OpenSSL 1.1.1, so I didn't have to specify ```SSL_CTX_set_max_proto_version```. However, I accept your PR because I believe it's a good idea to explicitly designate OpenSSL in case of future upgrades.

The cipher suites currently set in OME is set to the "Intermediate Compatibility" recommended by [Mozilla](https://wiki.mozilla.org/Security/Server_Side_TLS).
Changing cipher suites is hard to decide because I don't have the best practice, so I thought it would be better to use AES128 as you suggested and upgrade OME to set up in ```Server.xml``` for users to change cipher suites later.

basisbit:
OpenSSL supports TLS 1.3 and the docker build of OME HTTP server even announced its support, but actually negotiating tls 1.3 connection failed, clients then automatically fall back to 1.2. also see https://wiki.openssl.org/index.php/TLS1.3
This caused some extra delay at every chunk file download which is now reduced thanks to `SSL_CTX_set_max_proto_version`

dimiden:
Oh, that's something I didn't know. Thank you for the information.

PR #322, opened 2021-03-31
README.md: remove trailing space in Docker command

BaronKhan:
Trying to copy this command into a terminal for convenience won't work due to the trailing whitespace after the escape character. Figured this was a simple enough change to make and convenient for users :)


getroot:
Thanks! This is useful.

PR #339, opened 2021-04-15
Detect and clean up dead pull streams after 15 seconds

basisbit:
In some cases (for example HLS edge pulls from origin over OVT and the rtmp-source "died" without proper disconnect), OME did not properly clean up dead streams and thus only responded the same old playlist.m3u8 and old .ts chunks to hls / dash clients, which then immediately retry to get new chunks, eventually run out of buffer and then just show the loading animation.

Without this change, the only way to be able to reuse the stream url was to manually restart OME on all the edge servers, even when the stream source reconnected to the origin server over rtmp and started pushing new media data again (which the edge server never receive).

### DIFF
--- a/src/projects/base/provider/pull_provider/application.cpp
+++ b/src/projects/base/provider/pull_provider/application.cpp
@@ -86,6 +86,14 @@ namespace pvd
 									stream->GetApplicationInfo().GetName().CStr(), stream->GetName().CStr(), stream->GetId(), elapsed_time_from_last_recv);
 						}
 
+						// Abort pull stream when no packets arrived for more than 15 seconds. Most likely the stream source died, thus give OME a chance to correctly create a new stream + playlistis
+						if(elapsed_time_from_last_recv > 15)
+						{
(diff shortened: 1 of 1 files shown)


basisbit:
Tested this in "production" and seems to do its job as workaround for Edge servers which are connected over OVT not detecting that a stream died.

getroot:
PR Thank you. I'll review this.
I'm busy with other tasks these days, so I need some time.
I will comment again after review.

basisbit:
converted this PR to a draft, because after thinking a bit about it, this should not be merged. Better behavior would probably be, if the server is started as edge, to try to kill the a dead connection to the origin and then try to reconnect plus continue delivering chunk files to clients so that they can continue playback.

getroot:
Yes I tried to reconsider this after fixing all the Origin-Edge issues you reported. Thank you for your continued contribution. I'm wondering if the recently committed fixes solve your problem.

basisbit:
Closing this, because it seems like OME now recovers or completely stops and restarts the OVT connection on certain types of errors. (requires a current version of OME on the origin server and on the edge server).

PR #381, opened 2021-05-25
small Build fixes and enable CentOS 8 building

caveman99:
- ffmpeg configure won't find libmpx cause it is in ${PREFIX}/lib64/pkgconfig
- CentOS 8 doesnt need devtools and glibc-static
- Auto-Install libdrm-devel libX11-devel (the -y was missing, thus prompting)

### DIFF
--- a/misc/prerequisites.sh
+++ b/misc/prerequisites.sh
@@ -180,7 +180,7 @@ install_ffmpeg()
     mkdir -p ${DIR} && \
     cd ${DIR} && \
     curl -sLf https://github.com/FFmpeg/FFmpeg/archive/refs/tags/n${FFMPEG_VERSION}.tar.gz | tar -xz --strip-components=1 && \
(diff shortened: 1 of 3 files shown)


caveman99:
with these changes i got to build AND run the master branch on Centos 8

getroot:
Thank you very much for your contribution. I can't test it because I can't install centos 8 right now, but I trust you and merge this.

Does the latest master branch work well in your environment and centos 8? There have been reports that the streaming of the current master branch is unstable compared to the previous version. I wonder if it played well when tested in your environment.

caveman99:
it was running fine with the default config pushing a few MPEG streams. I am now setting up RTSP pull from a Sat>IP server so the satellite input is only activated on demand. I was testing in local net only. The binary basically is the same as on other platforms, since all of the prerequisites are compiled the same way. The changesets are just the tweaks to get the compile going through. Compiler is the stock GCC 8.4.1.

getroot:
Thank you for your confirmation. I asked because I wanted to find an environment where https://github.com/AirenSoft/OvenMediaEngine/issues/376 is reproduced.

PR #416, opened 2021-06-15
fixing memory leak issues in codec instances.

mehdi-qencode:
Hi,

I'm a member of Qencode team. We have tested OvenMediaEngine and found some memory leak issues. I have fixed some of them and making this pull request. 

Kind regards,

Mehdi Mammadov

### DIFF
--- a/src/projects/transcoder/codec/decoder/decoder_aac.cpp
+++ b/src/projects/transcoder/codec/decoder/decoder_aac.cpp
@@ -38,6 +38,9 @@ bool DecoderAAC::Configure(std::shared_ptr<TranscodeContext> context)
 
 	if (::avcodec_open2(_context, _codec, nullptr) < 0)
 	{
+        // close codec context to prevent definetly memory leak issue
+        ::avcodec_close(_context);
+
 		logte("Could not open codec: %s (%d)", ::avcodec_get_name(GetCodecID()), GetCodecID());
 		return false;
 	}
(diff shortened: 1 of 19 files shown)


getroot:
Thank you for your contribution. Please explain in a little more detail.

OME releases the transcoder object in the destructor below. Are there any objects that are not freed even in the destructor below? And is the memory leak reproducible?

https://github.com/AirenSoft/OvenMediaEngine/blob/7758b524623bc607edd128e6caeb7e9b5adbee27/src/projects/transcoder/transcoder_decoder.cpp#L33

mehdi-qencode:
Sorry for the late reply. We have tested OvenMediaEngine for transcoding performance for long periods of time (24, 36, 72 hours) and with various types of incoming streams. Memory leaks occur when incoming streams are unstable, when they are often terminated. There are leaks under normal conditions, but they are not significant. But in case of unstable streams, these leaks lead to the OvenMediaEngine crash. I analyzed with Valgrind and found the points where these leaks occur. The improvements in this pull request fixes these errors.

Keukhan:
@mehdi-qencode 

Sorry for the late reply.  

Thank you for your precious PR. but, referring to the link below, avcodec_free_context is being called, and avcodec_close is called within the function.

https://github.com/AirenSoft/OvenMediaEngine/blob/3635d74e27e9321e1c1b1e8228dc856dc8f0110d/src/projects/transcoder/transcoder_decoder.cpp#L43
https://github.com/AirenSoft/OvenMediaEngine/blob/3635d74e27e9321e1c1b1e8228dc856dc8f0110d/src/projects/transcoder/transcoder_encoder.cpp#L48

PR #423, opened 2021-06-21
Conditional serperator if Base URL has "?"

Thunder80:
The javascript policy generator had a bug. It didn't have the conditional seperator, i.e. "&" or "?". It failed for cases like wss://host:port/app/stream?direction=send, where the Base URL already had a question mark.


dimiden:
@Thunder80 Thank you very much!

PR #425, opened 2021-06-25
Stop setting the `display_aspect_ratio` on AVStreams

lf-:
This was causing a compilation failure on ffmpeg 4.4 on Arch Linux.

This field was moved out of AVStream into a private structure in this commit:
http://git.videolan.org/gitweb.cgi/ffmpeg.git/?a=commit;h=c1b916580ae92abca583d9afa2f9f64165292dd8

Based on the message there, it was not intended to be exposed to begin
with, so we can just stop setting it.


kwonkeukhan (approved):
It has been confirmed that there is no problem in operation even if the code is deleted.
Thank you.

genofire:
Nice - thx for merging ;)

Issue #456, opened 2021-07-29
Improve video quality: reencode preset


basisbit:
When using OME to reencode video streams, it currently uses ultrafast and zerolatency, which both together cause video streams at lower and medium bitrates to look like potato quality with lots of blocks and other artifacts, once there is lots of movement/change. Thus, I'd suggest changing the default preset to `faster`, which seems to get a good compromise on require CPU time vs resulting quality. See for example https://streaminglearningcenter.com/blogs/cut-clients-encoding-costs-75-30-minute-phone-call.html for some comparison charts.

Alternatively, please at least offer a config option to overwrite the default behavior, because this is probably a common thing where new users of OME install it, configure stuff as needed, and then they compare the result, and some manager says "this video looks better, let's pick that solution".

The choice of encoder preset does not influence playback stability in any webbrowser and thus does not result in any playback issues like stuttering or similar.

The code for this can be found at https://github.com/AirenSoft/OvenMediaEngine/blob/14e7923fe75ee9f1eb253ce38bd994b367a2c0e1/src/projects/transcoder/codec/encoder/encoder_avc.cpp#L77

If you are okay with changing the default to `faster` instead of `ultrafast`, then I'll create a PR for this.

Keukhan:
Hi @basisbit 

Thanks for the good suggestion.
I was planning to change to the default encoding option which is optimal considering the encoding quality and CPU resource usage.
If you PR, I will merge after testing.

thank you.

PR #457, opened 2021-07-29
Improve video quality: Use ffmpeg preset faster for reencoding video

basisbit:
See #456 . The ultrafast preset delivers too low video quality for a given bitrate. `faster` seems like a much better compromise between speed and quality, especially when using profile baseline for compatibility reasons.


Keukhan:
Thank you @basisbit

PR #525, opened 2021-09-29
Fix prerequisites.sh

bchah:
The install function calls appear to have been commented out recently 😅
Uncommenting them allows the make commands to proceed 👍


getroot:
Thank you! It's my mistake

Issue #577, opened 2021-11-30
Starting rtmp push in edge server.


trapasso:
**Describe the bug**
When no client plays a particular stream, on the edge server the startPush API fails to start a stream if this should come from a connection with an origin server. 

**To Reproduce**
Launch the startPush API on an edge that has not yet started a connection with a remote origin. (through the OVT protocol in my case) 

**Expected behavior**
I would have expected the edge server, automatically start the provider, on which the stream is configured. 

**Logs**
NO LOG

**Server (please complete the following information):**
 - OS: Ubuntu 18.04
 - OvenMediaEngine Version: 0.1.28
 - Branch: master

**Player (please complete the following information):**
no player involved 

**Additional context**
I would propose this patch, it seems to work fine. Thanks in advance and greetings. 
[rtmp-push-edge-server.txt](https://github.com/AirenSoft/OvenMediaEngine/files/7627483/rtmp-push-edge-server.txt)

getroot:
Great idea, great work.
Could you please make a PR?

PR #578, opened 2021-12-01
Rtmp push edge

trapasso:
In reference to problem #577. I forward the PR.
To build on ubuntu 18.04 with gcc 7.50, it was necessary to modify a constructor.


getroot:
Thank you!!

Issue #583, opened 2021-12-06
Bad crash, when running in debug version.


trapasso:
**Describe the bug**
Discovered the bug running the debug version.
Bud crash, with related stack trace log and server restart.

**To Reproduce**
1. Run debug version, with the level set to debug (level="debug").
2. Connect a player.

**Expected behavior**
NO CRASH

**Logs**
NO LOG.

**Server (please complete the following information):**
 - OS: Ubuntu 18.04
 - OvenMediaEngine Version: v0.12.8
 - Branch: master

**Player (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser Firefox
 - Version 94.00

**Additional context**
 Bad buffer overflow in segment_stream_private.h.
In the attached file a patch.
Thanks in advance. Greetings.
[crash-debug.txt](https://github.com/AirenSoft/OvenMediaEngine/files/7659886/crash-debug.txt)

trapasso:
Thank you so much.

PR #584, opened 2021-12-06
Patched bad crash, when running in debug version.

trapasso:
Reference to #583.


dimiden:
@trapasso 
Thank you! You found what I missed. Thank you!

Issue #598, opened 2021-12-15
REST API connections won't decrement when hiding the streamkey via the AdmissionWebhooks


chnagy:
**Describe the bug**
The REST API connections won't decrement at
`https://<host>/v1/stats/current/vhosts/default/apps/app/streams/<key>`

It work correctly as long as I won't add the AdmissionWebhooks.

**To Reproduce**
1. I pulled the latest docker image of ovenmediaengine
2. Uncomment at Bind/Managers/API to enable the REST API in the Server.xml
```
...
        <Bind>
                <!-- Enable this configuration if you want to use API Server -->
                <Managers>
                        <API>
                                <Port>${env:OME_API_PORT:8081}</Port>
                                <WorkerCount>1</WorkerCount>
                        </API>
                </Managers>
                <Providers>
...
```

3. Add the AccessToken in the Server.xml

```
...
        </Bind>

        <Managers>
                <Host>
                        <Names>
                                <Name>*</Name>
                        </Names>
                </Host>
                <API>
                        <AccessToken>secrettoken</AccessToken>
                </API>
        </Managers>

        <VirtualHosts>
...
```

4. Add the AdmissionWebhooks in the Server.xml

```
...
                        <AdmissionWebhooks>
                                <ControlServerUrl>http://<host>/auth</ControlServerUrl>
                                <SecretKey>secret</SecretKey>
                                <Timeout>3000</Timeout>
                                <Enables>
                                <Providers>rtmp,webrtc,srt</Providers>
                                <Publishers>webrtc,hls,dash,lldash</Publishers>
                                </Enables>
                        </AdmissionWebhooks>

                        <!-- Settings for applications -->
                        <Applications>
...
```

5. In the Control Server I replace the given streamkey based on a map to hide the actual streamkey as suggested in the docs.
6. Stream with OBS
7. Recieve with ovenplayer the webrtc stream.
8. Check API

```
HTTP/1.1 200 OK
Server: openresty
Date: Wed, 15 Dec 2021 16:59:02 GMT
Content-Type: application/json;charset=UTF-8
Content-Length: 445
Connection: close
X-Served-By: <host>

{
  "message": "OK",
  "response": {
    "connections": {
      "dash": 0,
      "hls": 0,
      "lldash": 0,
      "ovt": 0,
      "webrtc": 1
    },
    "createdTime": "2021-12-15T16:58:03.180+0000",
    "lastRecvTime": "2021-12-15T16:59:02.647+0000",
[...]

PR #612, opened 2021-12-26
Fixed (Cannot find stream) error when using new_url

Amirhasanioox:
Fixed (Cannot find stream) error when using new_url (Admission webhook) and client stops playing webrtc.

#598 

I fixed it for webrtc publisher and stats api shows number of connections correctly.


getroot:
Thank you very much!!

PR #613, opened 2021-12-27
Updated versions of OpenSSL, SRTP and SRT

basisbit:
Resolves some critical security / denial of service issues.
Starting with OPENSSL_VERSION 3.0.0-beta2, the output was written to `/opt/ovenmediaengine/lib64` instead of just `../lib`. The best temporary workaround I found so far was a symlink.


getroot:
Thank you very much!

PR #630, opened 2022-01-11
Updated versions of VPX, FDKAAC and FFMPEG and sync library versions between prerequisites.sh, Dockerfile, Dockerfile.nv

Lorekin:


Lorekin:
I tested rtmp+webrtc with transcoding.
I did not find any errors in the translation and in the logs.
If you are worried about compatibility, I can create a pull only to synchronize library versions between prerequisites.sh, Dockerfile, Dockerfile.nv

getroot:
I don't think there's a problem either. Since the master branch is the development version, I can merge this. If you find something strange while using it, please let me know again.

Issue #634, opened 2022-01-12
TCP RELAY, different external server port and local server port.


trapasso:
**Is your feature request related to a problem? Please describe.**
It would be interesting to add a parameter for the Tcp Relay in order to separate
the local binding (address and port) with the one that is sent as a candidate to the peer.
In a nutshell it would be interesting to have a local port different from the external port.

**Describe the solution you'd like**
One solution would be to use an additional parameter, in the `<IceCandidates>` tag.
The parameter could be `<TcpRelayLocal>:3434</TcpRelayLocal>`.
The external address would be `<TcpRelay>` and the internal address `<TcpRelayLocal>`.
Of course this implies that the external server does a NAT to the locale server address:port.
Above value could be expressed as `<TcpRelayLocal>ADDRESS:PORT</TcpRelayLocal>`.
ADDRESS: local ip address. PORT: local port.
If ADDRESS is empty then the address of the `<IP>` tag will be used.
If ADDRESS is equal to * all network interfaces will be used.
If `<TcpRelayLocal>` is not used then the behavior will not change compared to the
previous behavior with only the `<TcpRelay>` tag.

**Describe alternatives you've considered**
NO ALTERNATIVES

**Additional context**
I have build this patch [turn-nat.txt](https://github.com/AirenSoft/OvenMediaEngine/files/7856198/turn-nat.txt) which seems to work fine.

PR #635, opened 2022-01-12
TCP RELAY, different external server port and local server port.

trapasso:
Reference #634

### DIFF
--- a/src/projects/config/items/bind/common/webrtc/ice_candidates.h
+++ b/src/projects/config/items/bind/common/webrtc/ice_candidates.h
@@ -19,6 +19,7 @@ namespace cfg
 			protected:
 				std::vector<ov::String> _ice_candidate_list{"*:10000-10005/udp"};
 				ov::String _tcp_relay;
+				ov::String _tcp_relay_local;
 
 				int _tcp_relay_worker_count{};
 				int _ice_worker_count{};
@@ -29,6 +30,7 @@ namespace cfg
 
 				CFG_DECLARE_REF_GETTER_OF(GetIceCandidateList, _ice_candidate_list);
 				CFG_DECLARE_REF_GETTER_OF(GetTcpRelay, _tcp_relay);
+				CFG_DECLARE_REF_GETTER_OF(GetTcpRelayLocal, _tcp_relay_local);
 
 				CFG_DECLARE_REF_GETTER_OF(GetTcpRelayWorkerCount, _tcp_relay_worker_count);
 				CFG_DECLARE_REF_GETTER_OF(GetIceWorkerCount, _ice_worker_count);
@@ -38,6 +40,7 @@ namespace cfg
 				{
 					Register<Optional>("IceCandidate", &_ice_candidate_list);
 					Register<Optional>("TcpRelay", &_tcp_relay);
+					Register<Optional>("TcpRelayLocal", &_tcp_relay_local);
 					Register<Optional>("TcpForce", &_tcp_force);
 
 					Register<Optional>("TcpRelayWorkerCount", &_tcp_relay_worker_count);
--- a/src/projects/modules/ice/ice_port.cpp
+++ b/src/projects/modules/ice/ice_port.cpp
@@ -100,7 +100,7 @@ bool IcePort::CreateIceCandidates(const std::vector<std::vector<RtcIceCandidate>
 	return succeeded;
 }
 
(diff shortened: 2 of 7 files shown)


getroot:
good idea. But there are a few things I want to ask you.

- This complicates the setup. In contrast, what benefits can it provide to users?
- The term <TcpRelayLocal> is unlikely to be easily understood by users. How about explicitly calling it <TurnServer> instead?

```
<WebRTC>
  <Signalling>
	  <!-- <Port>13333</Port> -->
	  <TLSPort>13334</TLSPort>
	  <WorkerCount>1</WorkerCount>
  </Signalling>
  <TurnServer>*:13478</TurnServer>
  <IceCandidates>
	  <TcpRelay>192.168.0.160:13478</TcpRelay>
	  <TcpForce>true</TcpForce>
  
	  <IceCandidate>192.168.0.160:11000/udp</IceCandidate>
	  <TcpRelayWorkerCount>1</TcpRelayWorkerCount>
  </IceCandidates>
</WebRTC>
```

trapasso:
`<TurnServer>` is fine.

An advantage is that you can leave free choice to configure the external ports as opposed to the internal ones, in short, not being forced to use the same ports as the media servers also on the router.
This is a big plus.
Another advantage is that of being able to isolate the interface on which the turn server is listening.
It is true this is a slightly advanced configuration but if a user does not understand what `<TurnServer>` is, he can safely omit the parameter and the server behavior will remain exactly the same as before.
If you think it might be interesting, I have no problem changing the tag name from `<TcpRelayLocal>` to `<TurnServer>` and move it out from `<WebRTC>`.

thank

getroot:
Thanks for the detailed explanation.

There is something I still don't understand.

If TurnServer is bound to 192.168.0.100:3478, in fact, isn't there an alternative to using a port other than 192.168.0.100:3478 for TcpRelay?

In one usage example, TurnServer binds to *:3478, and TcpRelay provides 192.168.0.100:3478 or 211.100.2.33:3478, so it seems possible to provide another IP. However, it seems that other ports cannot be used.

Am I misunderstood?

trapasso:
Let's proceed for examples, valid after applying the patch of this Pull Request.
```
<TcpRelay>211.100.2.33:3478</TcpRelay> local BIND '*:3478' remote CANDIDATE '211.100.2.33:3478'
<TcpRelay>211.100.2.33:3478</TcpRelay><TcpRelayLocal>192.168.1.155:3478</TcpRelayLocal> local BIND '192.168.1.155:3478' remote CANDIDATE '211.100.2.33:3478'
<TcpRelay>211.100.2.33:443</TcpRelay><TcpRelayLocal>192.168.1.155:3999</TcpRelayLocal> local BIND '192.168.1.155:3999' remote CANDIDATE '211.100.2.33:443'
<TcpRelay>211.100.2.33:443</TcpRelay><TcpRelayLocal>*:3999</TcpRelayLocal> local BIND '*:3999' remote CANDIDATE '211.100.2.33:443'
<TcpRelay>211.100.2.33:443</TcpRelay><TcpRelayLocal>:3999</TcpRelayLocal> local BIND '<IP>:3999' remote CANDIDATE '211.100.2.33:443'
```
Where `<IP>` is the `<IP>` tag of the `<Server>` tag.
I think, this can clear up any doubts. Refer to this #634.

Thank

getroot:
`<TcpRelay>211.100.2.33:443</TcpRelay><TcpRelayLocal>192.168.1.155:3999</TcpRelayLocal> local BIND '192.168.1.155:3999' remote CANDIDATE '211.100.2.33:443`

I have no idea how this will work.
The actual bind is at 192.168.1.155:3999, how can the client (player) connect to 211.100.2.33:443? Port 443 is not open. Isn't it?

I'm not suspicious of you. :-)
I'm asking because I don't understand how this PR works.

getroot:
Oh, I found the descriptions you wrote on #634.

`Of course this implies that the external server does a NAT to the locale server address:port.`

trapasso:
yeah .. of course this implies NAT

getroot:
I now understand how this PR works. And this PR is acceptable as it can also be used with existing configurations.

However, in my opinion, the term TcpRelayLocal and the location of the setting don't seem to fit this function perfectly. I think TcpRelayBind or TurnServer etc are good.

If you come up with a better term than TcpRelayLocal (including my suggestion), please do a PR that applies it. And then, please update the manual as well.

Thank you very much for your contribution.

PR #667, opened 2022-02-06
Updated version of openh264, FFmpeg and nasm

Lorekin:


basisbit:
I'd suggest waiting for 5.0.1 or 5.0.2 to avoid the typical major bugs that new ffmpeg versions (or third party libraries who interact with it) tend to have.

getroot:
Thanks for the PR!
However, major version updates of ffmpeg should be very careful (may contain compatibility issues or bugs), so we'll look into this.

PR #669, opened 2022-02-09
Mpegts push

trapasso:
I propose this Pull Request to push on the mpegts protocol. The basic idea is this: through mpegts (udp or tcp) you can bring out the streams to OBS or ffmpeg for example, filter them and reacquire them in input, always with the mpegts protocol.

Thanks in advance.


getroot:
That's great work. I'll review the code.
Could you please update the manual so that other users can use it easily?

trapasso:
I can try. I would only edit the api part with an example.

```
{
  "id": "stream-9998-udp",
  "stream": {
    "name": "stream-00",
    "tracks": "" 
  },
  "protocol": "mpegts",
  "url": "udp://192.168.122.1:9998",
  "streamKey": ""
}
```
Thanks

getroot (changes requested):
This commit does not fit the topic of the PR, so please revert it. Also, too many files are modified at once without content, so I will work on it myself in the future. (There are many things in the project's source code that do not match the coding convention, so whenever the code is updated, I am correcting it by setting the convention.)
Thanks for letting me know it's set to 755.

getroot (changes requested):
I'm not familiar with git either, so I'm not sure, but I think you need to delete that commit. This change doesn't serve the original purpose of not wanting to change many files without content, since it's just one more commit added.

trapasso:
Ok .. deleted the commit that changed file permissions.
Thanks

getroot:
Thank you very much for your contribution.

PR #724, opened 2022-04-06
Updated version of openssl, openh264, FFmpeg and nasm

Lorekin:


src/projects/transcoder/codec/codec_utilities.h
```diff
@@ -14,6 +14,8 @@ extern "C"
 #include <libavformat/avformat.h>
 #include <libavutil/opt.h>
 #include <libavutil/pixdesc.h>
+#include <libavutil/channel_layout.h>
+#include "libavutil/cpu.h"
```
basisbit, line 18: Any reason for those includes without using any new functions from these header files?

PR #775, opened 2022-05-24
added python signed policy generator

Nictec:
I implemented the signed policy generator as a python class. The python file is directly callable with the same arguments as the js implementation but also importable in a python project


getroot:
Thank you very much!

PR #841, opened 2022-08-04
Fix epoch.

edudebacker:
1. Fix bug in generate() where timestamp was saved in seconds which cause bad policy formatting, changed to milliseconds.
2. Fix correct storing of 'allow_ip' from policy["allow_ip"] = allow_ip to policy_dict["allow_ip"] = allow_ip.
Script working fine now.


getroot:
Thank you for contributing!

PR #856, opened 2022-08-16
Replace broken docs link

Firerouge:
The rtc_signalling p2p code has moved under the modules subdirectory, previous link is a 404


getroot:
Thank you!!!

PR #892, opened 2022-09-16
Go support for generating signed policy URL

ivanngyb:
Added the support for generating signed policy in Golang! Included read me for examples.


getroot:
Thank you so much!!

PR #912, opened 2022-10-10
Tip before run manual installation

lonasozo:
On Google Cloud VM with Ubuntu18+, without running at the first start `sudo apt-get update` the prerequisites.sh run fails.  So I've added an hint to try run it before going on with manual installation.


getroot:
Thank you!

PR #928, opened 2022-10-27
Add ability to disable custom library paths at build

hashworks:
When one builds OME against shared system libraries they might not use `/opt/ovenmediaengine/lib` at all. In that case not disabling it completly can be a security risk.


hashworks:
Bump.

dimiden:
@hashworks 
Thanks for the PR!!! 👍

PR #934, opened 2022-11-02
Update OpenSSL version

basisbit:
Updated OpenSSL version.
Resolves a Denial of Service issue.


getroot:
Thank you
Does OME work well with OpenSSL 3.0.7?

basisbit:
It compiles and starts. Besides that I only did very short/basic testing.
(I am still on my old fork from more than a year ago, and am now trying to replace that old fork with default OvenMediaEngine from this repository here)

getroot:
Since the API has not changed in 3.0.7, it is better to merge it first and try it together. thank you

PR #942, opened 2022-11-08
Use local code files instead of re-downloading it

basisbit:
Simplify development environment setup

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,11 +15,8 @@ ARG     GPU=FALSE
 ENV     PREFIX=/opt/ovenmediaengine
 ENV     TEMP_DIR=/tmp/ome
 
-## Download OvenMediaEngine
-RUN \
-        mkdir -p ${TEMP_DIR} && \
-        cd ${TEMP_DIR} && \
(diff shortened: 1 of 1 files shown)


getroot:
Since it is also used by an external automatic build system, it should be possible to build with only one Docker file.

basisbit:
What build system are you using, any details available, so that we can find a solution that works well for both?
One of the reasons for this here pull request was to be able to use Github Actions to easily build and "test" any commit to main branch or any open pull request, to simplify development.

getroot:
We pass a single Dockerfile to Jenkins, or often a single Dockerfile to other users so they can build and use the Image.
If so, how about adding one more Dockerfile.local file that copies the local file? Or, it would be good to receive it as an option from the existing Dockerfile.

getroot:
Our automatic build system downloads https://raw.githubusercontent.com/AirenSoft/OvenMediaEngine/{version}/Dockerfile to build airensoft/ovenmediaengine:xxx , builds it and pushes it to Dockerhub.

One of our customers uses AWS Fargate and their automated build system uses the same method.

When the Dockerfile is changed, all of these systems must be modified.

basisbit:
Okay, I'll create a new pull request for the Dockerfile.local

PR #943, opened 2022-11-08
Change docker-compose version to add support for Ubuntu 20.04's default docker-compose

basisbit:
Changed docker-compose version to add support for Ubuntu 20.04's default docker-compose.

This docker-compose file doesn't make use of any docker-compose 3.7 features, thus setting it to 3.6 doesn't have any disadvantages, but increases range of supported docker-compose versions.


getroot:
Thank you!

PR #945, opened 2022-11-10
add signed_policy_url_generator java version

motolies:
add signed_policy_url_generator java version

### DIFF
--- a/misc/signed_policy_url_generator.java
+++ b/misc/signed_policy_url_generator.java
@@ -80,10 +80,6 @@ public class OvenMediaEngineKeygen {
         if (StringUtils.isNotBlank(allow_ip))
             policyMap.put("allow_ip", allow_ip);
 
-        if (policyMap.size() < 1)
-            throw new RuntimeException("At least one parameter is required.");
-
-
         String policy = gson.toJson(policyMap);
         String encodePolicy = encode(policy);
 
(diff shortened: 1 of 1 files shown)


getroot:
Thanks for your contribution.
- Please change the comments in the code from Korean to English.
- Is the main function for example? But could you please change it to the way it uses args[]?

motolies:
Changed comments to English and changed to use arguments when running the program.
Please check again.

basisbit:
imho the previous version with the easy to use "generate" function was much better. Java developers in 99% of the use cases will copy the function and call it within code, instead of using the untyped `main(String[] args)`.
I'd suggest keeping it like at the initial commit, so that it can easily be used as an example for other Java software devs.

getroot:
@motolies What do you think of basisbit's comments? I'm honestly not a Java user so I didn't know.

motolies:
@getroot In my personal experience, I have used a lot of frameworks such as spring rather than console applications using java.
So, it seems that many people will be able to use it more easily by providing it as a utility class rather than receiving an argument to the main function and using it.

If you want to do this, I'll change the comments to English and commit again.

@basisbit  Thanks for the good answer.

getroot:
@motolies Please re-commit like your last comment. sorry for bothering you.

motolies:
@getroot When I first sent a pull request to open source, I thought that each project had its own character.

Don't be sorry. I got a lot of help.

- Annotations changed to English
- Removal of unnecessary exception check due to provision of essential parameters

Please check again.
thank you

getroot:
Thank you!

PR #1019, opened 2023-02-07
Add feature for settings storage.

trapasso:
What I would like to propose with this P.R. it is the saving of the dynamic settings.

The design is really very simple.
At startup, the following configuration parameters are read:

```
<Server>
    <Managers>
        <API>
            <Storage>
                <Enabled>true</Enabled>
                <Path>Storage.xml</Path>
            </Storage>
        </API>
    </Managers>
</Server>
```
                            
those proposed above are the defaults.
These parameters already exist but are not documented nor used.

If the `<Enabled>` parameter is true then the configuration file with name `<Path>` is read.
The file contains a list of virtual hosts in xml format, identical to the `<VirutalHosts>` tag of the Server.xml file (of course it could also not exist or be empty).

The virtual hosts are then added to the other hosts (those of the normal configuration) but with the write flag enabled.
The virtual hosts inserted in this way can be accessed through the APIs, those for dynamic settings, e.g.: change virtual hosts, add apps, change apps, etc.
The total of virtual host addition or deletion operations is stored in a set of strings that contains "the names of the active virtual hosts".
It is not necessary to keep track of changes because the virtual hosts considered "added" will be saved complete with all their parameters.
This set is updated, thanks to the Storage::AddVhost and Storage::DelVhost functions.

As regards writing to the file (actual saving), it occurs when the server is closed or through the API `http://server.local:8081/v1/vhosts/storage`.

For more details you can see the source files of the P.R.
I found this feature really useful, I think and hope it will be of interest to others as well.

I remain available for any observations and/or comments.
Thanks in advance.


dimiden:
@trapasso
Thank you for your PR. However, I'm sorry that this PR has two issues and cannot be accepted.

1. ```include``` and ```${env:}``` are not functioning properly.
OME's ```Server.xml``` provides the ```include``` keyword to load other XML file, and ```${env:}``` syntax to obtain environment variables such as ```${env:PATH}```.
However, if you serialize XML as it's in the PR, there is a problem that ```include``` and ```macro``` values no longer exists.

2. The API URL ```/v1/vhosts/storage``` may confuse the user.
The API URL provided by OME is designed in the most "RESTful" style.
According to this design, the ```/v1/vhosts/storage``` API points to a virtual host named ```storage```, which can confuse users.

trapasso:
@dimiden

Very thank for taking the time to review this P.R.
I'll answer now point by point with examples taken from some tests I've done.

1) I try to insert a vhost with the following json:

`[{"distribution":"statistics.local","host":{"names":["desktop-origin.custom"],"tls":{"certPath":"${env:OME_CERTS_PATH}/desktop-origin.local.cert","keyPath":"${env:OME_CERTS_PATH}/desktop-origin.local.key"}},"name":"custom"}]`

as you can see above there are 2 `${env:OME_CERTS_PATH}`, these are read from the server with the right path (follows OME log):
```
Certified | certificate.cpp:35 | A certificate has been created for VirtualHost [custom]:
    Cert file path: /develop/OvenMediaEngine/src/bin/conf/origin/certs/desktop-origin.local.cert
    Chain cert file path:
    Private key file path: /develop/OvenMediaEngine/src/bin/conf/origin/certs/desktop-origin.local.key
```

After storage operation, the saved file (`Storage.xml` for default) contains the following xml:

```
<?xml version="1.0" encoding="utf-8"?>
<VirtualHosts>
	<VirtualHost>
		<Name>custom</Name>
		<Distribution>statistics.local</Distribution>
		<Host>
			<Names>
				<Name>desktop-origin.custom</Name>
			</Names>
			<TLS>
				<CertPath>${env:OME_CERTS_PATH}/desktop-origin.local.cert</CertPath>
				<KeyPath>${env:OME_CERTS_PATH}/desktop-origin.local.key</KeyPath>
			</TLS>
		</Host>
	</VirtualHost>
</VirtualHosts>
```

The file contains the same environment variables as those entered with json.
As regards the `include`, it is not in the storage's interest to include multiple files, at least for the way this P.R. is conceived.

2)  The problem you pose on the RESTFull API is true but it is at the same time very simple to solve. Just point the API under the path `http://server.local:8081/v1/vhosts-storage` instead of `http://server.local:8081/v1/vhosts/storage`.
Of course I tried it and it works fine.

Now, after the last commits, the `src/projects/api_server/api_server.cpp` file reports a conflict, but I can fix it in a short time.
Thanks in advance.

getroot:
I'm not the reviewer for this PR, but I'd like to make one comment.

As per the design guidelines of OME and REST APIs, this function is not a resource, so it is correct to implement it as an action. Same as HLS Dump. https://airensoft.gitbook.io/ovenmediaengine/rest-api/v1/virtualhost/application/stream/hls-dump

Therefore, this function should be implemented as below.

`POST http://server.local:8081/v1/vhosts:store`

trapasso:
Following @getroot's comment I changed the API call, `POST POST http://server.local:8081/v1/vhosts:store`.

Thank you all for your cooperation.

dimiden:
@trapasso 
I apologize for the delayed response.
After receiving this PR, I realized that some additional work is needed and I need time to prepare for it.
However, I'm currently busy with NAB preparation, so I may not have enough time to review it right away.
Thank you for your understanding, even if it's late.

trapasso:
@dimiden
there is hope that one day you can reach a verdict on this P.R.

Spes contra spem. (Latin proverb)
very thanks

phsm:
A really good idea to make OME support dynamic configuration. However, how to be if there's a fleet of OME origins/edges? With this implementation, you need to walk over all the servers in your fleet and update them. 

I think supporting "cluster configuration managers" such as zookeeper, etcd, consul would yield more benefit.

Or take an old school approach: implement live reload support for the daemon:
1. A configuration management system (ansible, chef, puppet etc.) updates the config file.
2. A signal is sent to OME daemon (e.g. SIGHUP).
3. Daemon performs graceful reload, just like nginx.

PR #1033, opened 2023-02-20
Signed Policy URL Generator - Parameter issue solved for c# 

umeshshende:


getroot:
Could you elaborate on what problem this solved and how?

umeshshende:
**Issue:** When generating sign in policy through cs code i found that during broadcast with this url `wss://serverurl:3334/app/stream1234?transport=tcp` default policy and signature were append to url using `"?"`

`var stream_url = string.Format("{0}?policy={1}", base_url, policy_base64);`
so my url become 
`wss://serverurl:3334/app/stream1234?transport=tcp?policy={1}`
to prevent this 

**Solution:** Checked if parameter present in the url if yes then use `"&"` else use `"?"`
```
 if (base_url.Contains("?")) { base_url = base_url + "&"; }
 else { base_url = base_url + "?"; }
```

getroot:
Thank you!

PR #1058, opened 2023-03-06
AdmissionWebhooks - Add client's user-agent information

choigilhoon:
The purpose of this P.R. is to add client's user-agent information in the AdmissionWebhook's request message.

**Supported protocols**

[Providers]
- webrtc

[Publishers]
- webrtc
- llhls


getroot:
A simple but useful feature. Could you please update the manual as well?

PR #1130, opened 2023-03-13
Add new_url in AdmissionWebhook's closing request message

choigilhoon:
image


getroot:
Good job!

getroot:
It's not a bug, but there is one problem with LLHLS.
AdmissionWebhooks will send opening status to Control Server multiple times if playback is attempted through multiple players over HTTP/2 on one computer. But even in that case, this code seems to send a single closing status.

This is actually close to the fundamental problem of HTTP streaming. But can we improve this?

PR #1147, opened 2023-03-21
Added SignedPolicy and AdmissionWebhooks to WHIP and Thumnail publisher

choigilhoon:
The purpose of this P.R. is to add AccessControl to the WHIP and Thumbnail Publisher.

### DIFF
--- a/src/projects/providers/rtmp/rtmp_stream.cpp
+++ b/src/projects/providers/rtmp/rtmp_stream.cpp
@@ -128,12 +128,12 @@ namespace pvd
 		}
 
 		// Send Close to Admission Webhooks
-		if (_url && _remote && _publish_url)
+		if (_requested_url && _remote && _final_url)
 		{
 			auto remote_address { _remote->GetRemoteAddress() };
 			if (remote_address)
 			{
-				auto request_info = std::make_shared<AccessController::RequestInfo>(_url, remote_address, _url->ToUrlString(true) == _publish_url->ToUrlString(true) ? nullptr : _publish_url);
+				auto request_info = std::make_shared<AccessController::RequestInfo>(_requested_url, remote_address, _requested_url->ToUrlString(true) == _final_url->ToUrlString(true) ? nullptr : _final_url);
 
 				GetProvider()->SendCloseAdmissionWebhooks(request_info);
 			}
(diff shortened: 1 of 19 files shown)


src/projects/providers/rtmp/rtmp_stream.cpp
```diff
@@ -128,12 +128,12 @@ namespace pvd
 		}
 
 		// Send Close to Admission Webhooks
-		if (_url && _remote && _publish_url)
+		if (_requested_url && _remote && _final_url)
```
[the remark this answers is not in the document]
getroot: Ah I realized that this is pvd::Stream::_final_url . Make _final_url a private member, and access it through get/set functions. I'm confused when reading the code.

getroot (commented):
Check it out please

getroot:
And it is better to use info::Stream::GetMediaSource() to make client info in closing request. 
In publisher GetLinkedStream()->GetMediaSource()

src/projects/providers/webrtc/webrtc_provider.h
```diff
@@ -102,6 +102,10 @@ namespace pvd
 		std::shared_ptr<Certificate> CreateCertificate();
 		std::shared_ptr<Certificate> GetCertificate();
 
+		bool AddStream(const std::shared_ptr<WebRTCStream> &stream);
```
getroot, line 105: AddStream / DeleteStream 

The function name conveys the misleading meaning of disposing of the Stream from the system. It is used to search the stream mapped to the session key in WHIP, so change it to another appropriate name.

src/projects/providers/webrtc/webrtc_provider.h
```diff
@@ -123,5 +127,9 @@ namespace pvd
 		std::shared_ptr<Certificate> _certificate = nullptr;
 
 		std::mutex _stream_lock;
+
+		mutable std::shared_mutex _streams_guard;
+		// Key: stream_key / Value: WebRTCStream
+		std::map<ov::String, std::shared_ptr<WebRTCStream>> _streams;
```
getroot, line 133: The name _streams is also not very clear.
This is a table to quickly find the stream mapped with the whip session key, so change it to such a variable name.

getroot (changes requested):
Check it out please

getroot (approved):

Issue #1172, opened 2023-04-05
Memory usage when no framerate specified for encoder


bchah:
If you do not specify a Framerate option in the Video encode settings, transcoding to H264 does not succeed and starts to overwhelm the input queue. This continues until the OME utilizes all available RAM then restarts. This happens even if the input is H264, not just HEVC. I believe the intended behaviour is that the source framerate would be used if no Framerate option was specified.

This is how the input queue keeps building, along with the RAM usage. Let me know if you need any more info to track this down, thank you 👍 

```
[2023-04-04 19:27:38.080] W [Rescaler:951] ov.Queue | queue.h:268  | [0x7f8849a01818] Input queue of Encoder. track(0) codec(h264/27) size has exceeded the threshold: queue: 152, threshold: 120, peak: 152
[2023-04-04 19:27:43.121] W [Rescaler:951] ov.Queue | queue.h:268  | [0x7f8849a01818] Input queue of Encoder. track(0) codec(h264/27) size has exceeded the threshold: queue: 303, threshold: 120, peak: 303
[2023-04-04 19:27:48.124] W [Rescaler:951] ov.Queue | queue.h:268  | [0x7f8849a01818] Input queue of Encoder. track(0) codec(h264/27) size has exceeded the threshold: queue: 453, threshold: 120, peak: 453
[2023-04-04 19:27:53.132] W [Rescaler:951] ov.Queue | queue.h:268  | [0x7f8849a01818] Input queue of Encoder. track(0) codec(h264/27) size has exceeded the threshold: queue: 602, threshold: 120, peak: 602
[2023-04-04 19:27:58.135] W [Rescaler:951] ov.Queue | queue.h:268  | [0x7f8849a01818] Input queue of Encoder. track(0) codec(h264/27) size has exceeded the threshold: queue: 753, threshold: 120, peak: 753
[2023-04-04 19:28:30.162] W [Rescaler:951] ov.Queue | queue.h:268  | [0x7f8849a01818] Input queue of Encoder. track(0) codec(h264/27) size has exceeded the threshold: queue: 1122, threshold: 120, peak: 1122
```

_Originally posted by @bchah in https://github.com/AirenSoft/OvenMediaEngine/issues/1169#issuecomment-1496498780_

Keukhan:
@bchah 

Thanks to report.

If no framerate is set, the intent is to encode at the same framerate as the original. 

If the problem is that the queue of the encoder is increased,  The problem is predicted.

[...]

PR #1174, opened 2023-04-05
Include framerate calculation fix in decoder_hevc

bchah:
This same calculation method was applied by @Keukhan to the other decoders, but not decoder_hevc where the memory issue #1172 is still occurring.


Keukhan:
many thanks :)

PR #1246, opened 2023-05-20
Added disableAudio option and fixed SETUP request for pulling rtsp streams

krsvital:


getroot:
Thanks for the PR. However, a detailed description of PR (motivation, purpose, method of use, related issues, etc.) is required.

And fixing the SETUP request URL bug in this PR is good. But I think the disableAudio option is bad because it is not applicable to OVT and other protocols that will be added in the future. Also, for the disableAudio option to be provided, disableVideo must also be provided, and must also be provided in a way that considers multi-track.

In your use case, it might be nice to have the option to configure the stream only with supported codecs, rather than terminating the stream when an unsupported codec is received.

Therefore, I would appreciate it if you PR only the bug fixing part first and write a detailed explanation. And the issue of ingesting streams with unsupported audio tracks deserves further discussion.

krsvital:
We use broadcasting from video cameras via webrtc, the cameras have  G711 audio codecs, they are not supported in the html5 player, so it became necessary to disable them. Therefore, disableAudio feature is a useful feature in some cases.

krsvital:
I added 2 new commits for fixing repos and rtsp sessions.
After OvenMedia sends PLAY camera stops streaming if haven't messages for a long time (1-2 minutes).
Therefore, they need to be sent periodically.

getroot:
I agree that disableAudio can be useful. However, it must be provided in a different way to make it scalable and universal. For example, I think the "preferred codecs and preferred codecs only" option would be more general and useful. This is also available with WebRTC Provider. Also, how to pull RTSP can be done with the API and OriginMap/OrignMapStore. Even in this case, this option should be available. I will think more about these options in the future and apply them.

For now I can merge the part of your code that automatically ignores unsupported codecs and creates the stream (instead of the stream creation failing).

The last committed keep-alive feature was in our backlog. (https://github.com/AirenSoft/OvenMediaEngine/issues/1187) Thanks for committing this.

However, since PRs must be separated by topic (including content that does not match the title of this PR), it will be merged if you separate and PR for bug fixes, ignoring unsupported codecs, and keep-alive. thank you

krsvital:
ok, but i have an idea. I want to create server which dont pulling streams on creating with rtsp (with API pull request), but created session and keep it in pause state. And i want to do API feature for OME that can play|pause it. And streams, which not active have not to be TEARDOWN, but PAUSED. It must be usefull for streaming online cameras on demand. And it would be fast and save traffic.

getroot:
@krsvital Ok, if you have any ideas for new features or design changes, please create a discussion on the Discussion board.

src/projects/providers/rtspc/rtspc_stream.cpp
```diff
@@ -914,6 +1059,10 @@ namespace pvd
 			}
 			else
 			{
+				if(time(nullptr) - _last_check_time >= 30){
```
jonas-resch, line 1062: I think the timeout should be taken from the SETUP response of the RTSP camera

```
[04-27 19:21:56.360] I [StreamCollector:630581] RtspcStream | rtspc_stream.cpp:448 | Response SETUP: Response: RTSP/1.0 200 OK
Connection: Keep-Alive
CSeq: 11
Date: Thu, 27 Apr 2023 10:21:45 GMT
Session: 2613658800411225924828208291162;timeout=120
Transport: RTP/AVP/TCP;unicast;interleaved=20-21;ssrc=78CD202E
```

jonas-resch:
Hi @krsvital & @getroot is there a way to, in the meantime, merge the fix for the RTSP session timeout? Maybe @krsvital could create a separate pull request for the fix?
Thank you :)

getroot:
@jonas-resch You (and anyone) can merge this PR into your local branch at any time. We would be grateful if you could also let us know the test results.

src/projects/providers/rtspc/rtspc_stream.cpp
getroot: I agree. It is recommended to update at about half of the SETUP's timeout.

getroot (changes requested):
In order for this PR to be merged, the disableAudio option must be deleted and developed in a different way (eg preffered codecs only option) through future discussion so that it can be applied to all providers. Please delete this commit.

src/projects/providers/rtspc/rtspc_stream.cpp
krsvital: I have a camera that shown timeout=0 and drop connection after 90 sec

jonas-resch: Then maybe use the timeout if provided and fall back to 30s if there is no timeout set or it is 0?

jonas-resch:
I did a test with @krsvital master branch. I let the stream run for around 15 minutes and it seems to work.

The only thing I get now is an infinite amount of ["[PTS] Reverse wrap around detected. track:20"](https://github.com/AirenSoft/OvenMediaEngine/blob/03ce7b6b3e0e60d3e3ba2bc8308f95a9ea2bf2d9/src/projects/base/provider/stream.cpp#LL290C5-L290C5)). But maybe that has something to do with @getroot commit a791f9b? So far it didn't cause any issues, besides constantly outputting the message to stdout.

getroot:
@jonas-resch That's probably because your RTSP camera initially sent the wrong timestamp. I also often find cameras like this. I'm working on this issue and will patch it soon.

getroot:
@krsvital 
I was wondering if you could kindly confirm if you are currently working on the request for this PR. The keep-alive feature has been on my backlog, so please let me know if you no longer wish to work on it. It is something I must attend to.

getroot:
Since this PR is no longer active, it won't be merged. If you would like to continue working on this PR, please reopen and discuss.

PR #1258, opened 2023-05-30
Fix potential memory leak

cuishuang:
When using time.After in a loop, it should be defined outside the loop, otherwise memory leaks will occur due to non-stop creation of new variables to apply for memory


getroot:
Thank you very much! 

I am not familiar with the Go language, so I cannot verify if this is correct, but I will trust you and merge it since it is a simple modification.

getroot:
This patch has a problem that causes the -n option to not work. Any ideas about this?

PR #1269, opened 2023-06-08
Add "lastThroughputIn" and "lastThroughputOut" to Stream Stats API

bchah:
Hi guys! I noticed the Stream Stats API only returns the _average_ incoming video bitrate. So if you are streaming 4Mb/s for a long time then switch to 1Mb/s, it will take some time for the users to see this.

Please accept this pull request which tested OK on my end. It reveals "lastThroughputIn" and "lastThroughputOut" to show the bitrate in or out for the last second.


getroot:
Thank you very much!

PR #1337, opened 2023-08-09
fix: use correct path for config file volume in docker-compose.yml

basisbit:
This can never have worked as specified in the commented out snippet. This merge request sets the correct internal path to the config folders as defined in the Dockerfile.


dimiden:
You are correct. The ```bin``` was missing. Thank you for the pull request!

PR #1362, opened 2023-09-01
Add x264 to enable 10-bit high422 WebRTC and HLS to Chrome

cameronelliott:
The Chrome browser has support for 10-bit high422 H.264 WebRTC playback.
[libwebrtc-commit][h264]

10-bit video support is valued in the film and video industry, and also the
gaming industry.

This PR enables OME to send 10-bit high422 H.264 over WebRTC to Chrome.

This enables piplines like this: OBStudio -> 10bit/422/HEVC/SRT-or-WHIP -> OME
-> 10bit/422/H264/WebRTC -> Chrome

This enables real-time low latency 10-bit broadcasting enabled by OME!

This may be the first pipeline enabling 10-bit low latency broadcasting to the
Chrome browser using OBS. OME can make this possible.

We don't believe the GPLv2 license of x264 presents a conflict with licensed
used in OME:

1. OME is licensed under the AGPLv3
1. The [x264 GPLv2 allows redistribution under the terms of the GPLv3][1]
1. The [AGPLv3 and GPLv3 are compatible][agplgpl]
1. OME does not redistribute x264 in source form
1. OME/Airensoft does redistribute x264 in binary form in Docker images

If, in spite of these arguments, the OME maintainers are worried about
redistributing Docker images containing x264 object code, then the x264 features
could be left out of the default Docker images. Only to be built and included
using special build flags.

Although not the focus of my efforts, this may also enable 10-bit/422 H.264
LLHLS and other streaming methods also.

PLEASE let me know if these PR's benefits are of interest, and I will proceed
with editing this PR to the maintainers standards in the interest of merging
when acceptable.

Thank you

[1]: https://github.com/mirror/x264/blob/master/COPYING#L296
[agplgpl]: https://www.gnu.org/licenses/gpl-faq.html#AGPLGPL
[h264]: https://webrtc-review.googlesource.com/c/src/+/256964

### DIFF
--- a/src/projects/transcoder/codec/encoder/encoder_avc_x264.h
+++ b/src/projects/transcoder/codec/encoder/encoder_avc_x264.h
@@ -25,7 +25,7 @@ public:
 
 	int GetSupportedFormat() const noexcept override
 	{
-		return AV_PIX_FMT_YUV420P;
+		return AV_PIX_FMT_YUV422P10LE;
 	}
 
 	cmn::BitstreamFormat GetBitstreamFormat() const noexcept override
--- a/src/projects/transcoder/codec/encoder/encoder_avc_x264.h
+++ b/src/projects/transcoder/codec/encoder/encoder_avc_x264.h
@@ -1,0 +1,42 @@
+//==============================================================================
+//
+//  Transcode
+//
+//  Created by Kwon Keuk Han
+//  Copyright (c) 2018 AirenSoft. All rights reserved.
+//
+//==============================================================================
+#pragma once
+
+#include "../../transcoder_encoder.h"
+
+class EncoderAVC : public TranscodeEncoder
+{
+public:
+	EncoderAVC(const info::Stream &stream_info)
+		: TranscodeEncoder(stream_info)
+	{
+	}
+
+	AVCodecID GetCodecID() const noexcept override
+	{
+		return AV_CODEC_ID_H264;
+	}
+
+	int GetSupportedFormat() const noexcept override
+	{
+		return AV_PIX_FMT_YUV420P;
+	}
+
+	cmn::BitstreamFormat GetBitstreamFormat() const noexcept override
+	{
+		return cmn::BitstreamFormat::H264_ANNEXB;
+	}
+
+	bool Configure(std::shared_ptr<MediaTrack> context) override;
+
+	void CodecThread() override;
+
+private:
+	bool SetCodecParams() override;
+};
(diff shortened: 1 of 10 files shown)


getroot:
Thanks for the PR.

First of all, we have all the code related to x264. (Currently only in private repo) If we decide to add x264, we just have to deploy the code. We will look further into licensing-related issues.

For things related to 10 bit, it would be better to handle it with an encoding profile option for versatility.

The fixed fmtp property value in SDP is a trick for compatibility, as you may have seen in my comments. If it is changed to a high value it will most likely not play in any other browser except chrome. (Have you tried other browsers?) I'll think about this again. In modern browsers, it may be better to change it to a normal value.

src/projects/transcoder/codec/encoder/encoder_avc_x264.h
```diff
@@ -0,0 +1,42 @@
+//==============================================================================
```
captain-hoity-toity, line 1: The code appears to be well-structured and follows good coding practices. However, I have a few suggestions that could potentially improve the code:

1. **Include necessary headers**: The code doesn't include the necessary headers for the types and functions used. Make sure to include all necessary headers for clarity and to prevent potential compilation errors.

```cpp
#include <memory>
#include "media_track.h"
#include "bitstream_format.h"
```

2. **Use explicit constructors**: The single-argument constructor `EncoderAVC(const info::Stream &stream_info)` can potentially be used as an implicit conversion. To avoid this, mark the constructor as explicit.

```cpp
explicit EncoderAVC(const info::Stream &stream_info)
```

3. **Document functions**: Although the code is quite clear, adding comments to explain what each function does would be beneficial for other developers reading your code.

```cpp
/**
 * Sets the codec parameters.
 * @return true if successful, false otherwise
 */
bool SetCodecParams() override;
```

4. **Error handling in `SetCodecParams`**: The `SetCodecParams` function should handle errors properly. If an error occurs while setting the parameters, it should return false or throw an exception.

5. **Consistent Naming Conventions**: Ensure to follow a consistent naming convention. The function `GetSupportedFormat` and `GetBitstreamFormat` are in PascalCase whereas `SetCodecParams` is in camelCase. It's usually best to stick to one naming convention for consistency.

6. **Thread safety in `CodecThread`**: If `CodecThread` is accessed by multiple threads, make sure to add necessary locks to prevent data races.

Overall, the code looks good and with a few adjustments, it can be made even better.

PR #1378, opened 2023-09-13
Correct typo in Alerts sample documentation

basisbit:
HasBFrame is not valid config, and according to the code, it should be HasBFrames instead.


choigilhoon:
thank you!

PR #1295, opened 2023-07-05
Cuda driver installation update on ubuntu

nums:
To prevent missing libnvcuvid lib, add a script to install cuda lib

https://github.com/AirenSoft/OvenMediaEngine/issues/929

PR #1402, opened 2023-10-03
Add dockerfiles for driver-agnostic CUDA builds

d-uzlov:
Currently OME requires you to [build a special docker image](https://github.com/AirenSoft/OvenMediaEngine/tree/efa9ea69ff59ccee4d8c6238342e5a002a78a45d/docs/transcoding/gpu-usage#how-to-build-and-run-docker-image) for nvidia transcoding, with driver version matching that on the host.
This is not the intended way of using the nvidia container runtime.

Nvidia doesn't do a great job of spreading the documentation on its forums and github issues but the intended way is to mount host driver files into the container.
You can find the documentation here: https://github.com/NVIDIA/nvidia-container-runtime

In this PR I optimized build process for docker images with CUDA:
- I used a prebuilt CUDA image which saves a lot of time
- Since we are using a separate base image anyway, I separated CUDA Dockerfile from the main one
- I added the `NVIDIA_DRIVER_CAPABILITIES` environment variable which allows you to skip driver installation inside the container
- I rearranged the commands to make a more efficient use of docker build cache for `.local` Dockerfile builds

Resulting image should work with any driver version.

Resulting image is based on `ubuntu:20.04` without any additional tools installed, which makes it 337 MiB uncompressed and 93 MiB compressed (which, I believe, is even smaller than the official images without CUDA).

You can find my image here:
- `docker.io/daniluzlov/k8s-snippets:ome-v0.15.16-cuda`
- https://hub.docker.com/layers/daniluzlov/k8s-snippets/ome-v0.15.16-cuda/images/sha256-da6ee9bb89abed3a7a810840ef07b04007838e2baa3c21080dc753fabf602233?context=repo

This PR will also fix this, and probably some other issues:
- https://github.com/AirenSoft/OvenMediaEngine/discussions/1398

# Notes

- With this image you could publish an official image with NVIDIA/CUDA support that doesn't require users to rebuild the image for each machine.
- Considering that this image is even smaller than the official one, maybe it could become the standard image. I believe there are no issues running cuda-enabled images without cuda hardware.
- While I believe that there should be no issues, because nvidia container runtime is the same everywhere, and I'm able to run the image linked above on my server, I don't have too many servers to test it on. I didn't even test it in docker, because I use containerd+runc+k8s.
- I believe, this dockerfile bumps CUDA version from 10.1 to 12.2. While I don't think this really matters, as I think there are no reasons users would not be able to install modern nvidia drivers with 12.2 support, if you want to, it should be really easy to downgrade the CUDA version: you just change the base image and `NVIDIA_REQUIRE_CUDA`.
- I didn't touch the documentation, or the main dockerfiles, or the install scripts. If you merge this PR, they should probably be updated.

### DIFF
--- a/Dockerfile.cuda
+++ b/Dockerfile.cuda
@@ -1,0 +1,53 @@
+FROM    nvidia/cuda:12.2.0-devel-ubuntu20.04 AS build
+
+WORKDIR /tmp
+
+ENV     DEBIAN_FRONTEND=noninteractive
+RUN     apt-get update && apt-get install -y --no-install-recommends tzdata sudo curl git
+
+# Where all libs will be installed
+ENV     PREFIX=/opt/ovenmediaengine
+
+## Download sources
+ARG     OME_REPO=github.com/AirenSoft/OvenMediaEngine
+ARG     OME_VERSION=master
+RUN \
+        mkdir -p /tmp/ome && \
+        cd /tmp/ome && \
+        curl -sLf https://${OME_REPO}/archive/${OME_VERSION}.tar.gz | tar -xz --strip-components=1
+
+## Install dependencies
+RUN     apt-get update && /tmp/ome/misc/prerequisites.sh --enable-nvc
+
+## Build OvenMediaEngine
+ARG 	STRIP=TRUE
+RUN \
+        cd /tmp/ome/src && \
+        make release -j$(nproc) && \
+        if [ "$STRIP" = "TRUE" ] ; then strip /tmp/ome/src/bin/RELEASE/OvenMediaEngine ; fi
+
+## Export release files
+RUN \
+        cd /tmp/ome/src && \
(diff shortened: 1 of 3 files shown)


getroot:
This is a very nice PR.
@Keukhan , please review this.

Keukhan:
@d-uzlov  Thanks for the great PR. I will test this and then merge.

nums:
This may allow this PR to be removed: https://github.com/AirenSoft/OvenMediaEngine/pull/1295

Keukhan:
> This may allow this PR to be removed: #1295



> This may allow this PR to be removed: #1295

Keukhan:
> This may allow this PR to be removed: #1295

#1295 PR will also be tested together. It's the end of the year, so I'm very busy.  Thanks :)

Keukhan:
Thank you for the PR. Some improvements seem to be needed after merging. I will update it. :)

PR #1418, opened 2023-10-24
added IgnoreRtcpSRTimestamp api option

krsvital:
Added new option [ IgnoreRtcpSRTimestamp ] for pulling streams to api server which disable waiting answer from server if it dont supporting sending RTCP

### DIFF
--- a/src/projects/api_server/controllers/v1/vhosts/apps/streams/streams_controller.cpp
+++ b/src/projects/api_server/controllers/v1/vhosts/apps/streams/streams_controller.cpp
@@ -94,9 +94,14 @@ namespace api
 						{
 							properties->SetUnusedStreamDeletionTimeout(jv_properties["unusedStreamDeletionTimeoutMs"].asInt());
 						}
+
+						if (jv_properties["defaultRtpCalculationMethod"].isNull() == false && jv_properties["defaultRtpCalculationMethod"].isInt())
+						{
+							properties->SetdefaultRtpCalculationMethod(jv_properties["defaultRtpCalculationMethod"].asInt());
+						}
 					}
 					
-					logti("Request to pull stream: %s/%s - persistent(%s) noInputFailoverTimeoutMs(%d) unusedStreamDeletionTimeoutMs(%d)", app->GetName().CStr(), stream_name.CStr(), properties->IsPersistent() ? "true" : "false", properties->GetNoInputFailoverTimeout(), properties->GetUnusedStreamDeletionTimeout());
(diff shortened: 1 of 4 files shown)


src/projects/api_server/controllers/v1/vhosts/apps/streams/streams_controller.cpp
```diff
@@ -94,9 +94,14 @@ namespace api
 						{
 							properties->SetUnusedStreamDeletionTimeout(jv_properties["unusedStreamDeletionTimeoutMs"].asInt());
 						}
+
+						if (jv_properties["defaultRtpCalculationMethod"].isNull() == false && jv_properties["defaultRtpCalculationMethod"].isInt())
+						{
+							properties->SetdefaultRtpCalculationMethod(jv_properties["defaultRtpCalculationMethod"].asInt());
```
getroot, line 100: If your API receives these values as 1 or 2, your users will be confused. Rather, it would be better to use something like IgnoreRtcpSRTimestamp in boolean form. I would like the default to be false.

src/projects/base/provider/stream.cpp
```diff
@@ -248,7 +250,21 @@ namespace pvd
 
 	bool Stream::AdjustRtpTimestamp(uint32_t track_id, int64_t timestamp, int64_t max_timestamp, int64_t &adjusted_timestamp)
 	{
+		auto stream = std::dynamic_pointer_cast<pvd::PullStream>(GetSharedPtr());
```
getroot, line 253: This codes must be in the UNDER_DECISION block below. Otherwise this code will be called every time.

src/projects/base/provider/pull_provider/stream_props.h
```diff
@@ -123,12 +131,22 @@ namespace pvd
 			_failback_timeout = milliseconds;
 		}
 
+		void SetdefaultRtpCalculationMethod (int32_t method){
```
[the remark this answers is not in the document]
getroot: Setdefault name should be modified to SetDefault

getroot:
And this function must work not only in Pull Stream using API, but also in Pull Stream using OriginMap and OriginMapStore.
Additionally, users must be able to use this function as an option in RTP-based push providers such as WebRTC Provider.

krsvital:
> And this function must work not only in Pull Stream using API, but also in Pull Stream using OriginMap and OriginMapStore. Additionally, users must be able to use this function as an option in RTP-based push providers such as WebRTC Provider. Users will be confused if this option only works with the Pull Stream API.

I think that for other variants this option is not needed, since it only matters in those streams that are added on the fly with the persistent = false key (as temporary), so the api will be enough

I changed defaultRtpCalculationMethod to IgnoreRtcpSRTimestamp option

getroot (approved):

getroot:
Thank you for submitting the PR. This option should also be made available in OriginMap and OriginMapStore, but we will work on that gradually. Before merging, could you please add an explanation for this option in the manual?

https://github.com/AirenSoft/OvenMediaEngine/tree/master/docs/rest-api/v1/virtualhost/application/stream#create-stream-pull

PR #1487, opened 2023-12-25
docs: extend SRT streaming instructions for client config

basisbit:
Extend documentation for how to configure OBS or Blackmagic Web Presenter or similar clients for streaming over SRT. This also adds information on how to specify latency parameter for less frustration when first time trying SRT.


getroot:
Thank you very much!

PR #1492, opened 2024-01-04
Fixed drift/loss of sync after "RTP/RTCP timestamp is not monotonic"

DorianRouby:
Hello,
 
**Observations**:
When publishing a stream in Webrtc, the audio and video streams are no longer synchronized after a while.
The video track may be delayed by several seconds compared to the audio track.

To be sure that the desynchronization didn't come from my PC, I checked using the webcodecs that the timestamps were correct at the time of sending.
I had no desync, while the OvenMedia logs showed a desync of around 2 seconds:

```
[01-02 11:07:28.994] D [InboundWorker:35] MediaRouter | mediarouter_stream.cpp:944 |
  - MediaRouter Stream | id: 102, type: Inbound, name: #main#meeting-dev/64be5e0ffd9dab001172db99-6593e7df36a7ec0012ac2d73-6593e7df36a7ec0012ac2d74, uptime: 1680466ms, queue: 0, sync: 2002ms
track:-1894865268, type: Audio, codec: OPUS(8,none), pts: 1679827ms, dly: 639ms, tb: 1/48000, pkt_cnt: 83896, pkt_siz: 18.6MB, bps: 88.54Kb/0b
track: 538366599, type: Video, codec: H264(1,none), pts: 1677825ms, dly: 2641ms, tb: 1/90000, pkt_cnt: 26452, pkt_siz: 219.7MB, bps: 1.05Mb/0b, fps: 15.76
track: 1416380800, type: Data, codec: Unknown(0,none), pts: 0ms, dly: 1680466ms, tb: 1/ 1000, pkt_cnt: 0, pkt_siz: 0B, bps: 0b/0b
```

After analyzing the code, it seems that the problem comes from the RTP and RTCP timestamps.
More precisely, this would come from calculating the "extended" values in lip_sync_clock:
  - _extended_rtp_timestamp
  - _extended_rtcp_timestamp

Each time an RTP packet is incorrectly ordered ("RTP/RTCP timestamp is not monotonic"), we lose synchronization between tracks and delay is added.

**Resolutions**:
I changed the delta value in the case mentioned above.
Setting the delta to 0 generates an offset on the next timestamps. This causes drift and loss of synchronization.

**Testing**:
I tested my code here in advance with many fake values: https://onlinegdb.com/K11cDIXiY
Then I tested by building a docker image that I hosted on my own server.

I remain available for any observations and/or comments. Thanks in advance. :)

### DIFF
--- a/src/projects/base/ovlibrary/lip_sync_clock.cpp
+++ b/src/projects/base/ovlibrary/lip_sync_clock.cpp
@@ -43,7 +43,7 @@ std::optional<uint64_t> LipSyncClock::CalcPTS(uint32_t id, uint32_t rtp_timestam
 		// UpdateSenderReportTime(id, msw, lsw, rtp_timestamp);
 	}
 
-	uint32_t delta = 0;
+	int64_t delta = 0;
 	if (clock->_last_rtp_timestamp == 0)
 	{
 		clock->_extended_rtp_timestamp = rtp_timestamp;
@@ -65,8 +65,10 @@ std::optional<uint64_t> LipSyncClock::CalcPTS(uint32_t id, uint32_t rtp_timestam
 			else
 			{
 				// reordering or duplicate or error
-				delta = 0;
-				logtw("RTP timestamp is not monotonic: %u -> %u", clock->_last_rtp_timestamp, rtp_timestamp);
+				// delta = 0; /!\ Setting the delta to 0 generates an offset on the next timestamps. This can cause drift and loss of synchronization
+				// delta cannot be greater clock->_extended_rtp_timestamp
+        delta *= -1;
+				logtw("RTP timestamp is not monotonic: %u -> %u delta: %u", clock->_last_rtp_timestamp, rtp_timestamp, delta);
 			}
 		}
 
@@ -120,7 +122,7 @@ bool LipSyncClock::UpdateSenderReportTime(uint32_t id, uint32_t ntp_msw, uint32_
 	}
 	else
 	{
-		uint32_t delta = 0;
+		int64_t delta = 0;
 		if (rtcp_timestamp > clock->_last_rtcp_timestamp)
 		{
 			delta = rtcp_timestamp - clock->_last_rtcp_timestamp;
@@ -137,8 +139,10 @@ bool LipSyncClock::UpdateSenderReportTime(uint32_t id, uint32_t ntp_msw, uint32_
 			else
 			{
 				// reordering or duplicate or error
-				delta = 0;
-				logtw("RTCP timestamp is not monotonic: %u -> %u", clock->_last_rtcp_timestamp, rtcp_timestamp);
+				// delta = 0; /!\ Setting the delta to 0 generates an offset on the next timestamps. This can cause drift and loss of synchronization
+				// delta cannot be greater clock->_extended_rtcp_timestamp
+        delta *= -1;
+				logtw("RTCP timestamp is not monotonic: %u -> %u delta: %u", clock->_last_rtcp_timestamp, rtcp_timestamp, delta);
 			}
 		}
 


getroot:
Thanks for your hard work. I'll review it soon.

DorianRouby:
Hello,

I see that the pull request has moved to "Closed" status. Let me add some information.

With a good internet connection, the problem mentioned is a problem that doesn't appear (or rarely). Drifting can happen after several hours of streaming. However, with an unstable internet connection, this may happen more quickly.

Indeed, sync loss occurs when RTP packets are lost. Currently working on a project using a satellite connection (Marlink), I have many lost packets and therefore a drift which appears quickly (or very quickly). Correcting this problem is essential in the context of my project but I also think that this problem will also impact all those who will push a webrtc stream via an unstable network (mobile network, satellite network, public network, etc.) .

Don't hesitate to get back to me if you have any questions.
I thank you in advance for your time and for all the work you provide.

getroot:
Hi, I haven't been paying attention to this lately. I'm really sorry. But I really appreciate your contribution.

Is this an issue that occurs even when RTCP SR reaches OME? And I'm still inexperienced with packet reordering since I mainly use WebRTC/tcp.

DorianRouby:
No problem :)


To be honest, I didn't really look at the rtcp part (I only worked on RTP packets) and I carried out the fix on these because the calculation error is the same.

Here we are just dealing with a calculation error.

If we have A, B, C, D, E, F (6 RTP packets) with the following timestamps:
 - A: 10
 - B: 30
 - C:50
 - D:70
 - E:90
 - F: 120

If packets arrive out of order arriving in the following order: A, B, **D**, **C**, E, F (instead of A, B, C, D, E, F)
This is how "extended values" ​​are currently calculated (simplifying overflow cases)

|   | timestamp | delta (timestemp - last timestamp) | extended (last extended + delta) | order is correct | result
|--|:-------------:|:-------------:|:-------------:|:-------------:|------:|
| A | 10 | - | 10 | ✅️ | first value
| B | 30 | (30 - 10) = **20** | 10 + 20 = **30** | ✅️ | ✅️
| D | 70 | (70 - 30) = **40** | 30 + 40 = **70** | ✅️ | ✅️
| C | 50 | (50 - 70) = **-20** | 70 + 0  = **70** | ❌️ | ❌️ correct value is: 50
| E | 90 | (90 - 50) = **40** | 70 + 40 = **110** | ✅️ | ❌️ => correct value is: 90
| F | 120 | (120 - 90) = **30** | 110 + 30 = **140** | ✅️ | ❌️ => correct value is: 120

In the previous table, we used delta = 0 when "RTP timestamp is not monotonic"
We observe that the value has drifted by 20 and we will have this loss of synchronization until the end of the stream. If this happens many times, then we can quickly reach a lag of several hundred milliseconds (or seconds).

In order to solve the problem, we must not touch the delta formula. Do not put a particular case with a delta = 0. You must apply the value even if the delta is negative (so that the next values ​​are not impacted). If we touch a link in the chain (delta), then all the values ​​that follow will be incorrect.

By keeping the delta value (even if it is negative):
|   | timestamp | delta (timestemp - last timestamp) | extended (last extended + delta) | order is correct | result
|--|:-------------:|:-------------:|:-------------:|:-------------:|------:|
| A | 10 | - | 10 | ✅️ | first value
| B | 30 | (30 - 10) = **20** | 10 + 20 = **30** | ✅️ | ✅️
| D | 70 | (70 - 30) = **40** | 30 + 40 = **70** | ✅️ | ✅️
| C | 50 | (50 - 70) = **-20** | 70 + (-20)  = **50** | ❌️ | ✅️
| E | 90 | (90 - 50) = **40** | 50 + 40 = **90** | ✅️ | ✅️
| F | 120 | (120 - 90) = **30** | 90 + 30 = **120** | ✅️ | ✅️

Here we keep our sync.
I don't know if these tables were able to clarify our problem.

getroot:
I agree with your solution.

However, I don't quite understand the situation where video frames arrive out of order. When OvenMediaEngine receives an RTP packet, it puts it in the RtpFrameJitterBuffer, and when the frame is complete, it takes it out. And it drops the old frame. It doesn't take the old frame out again. If the frame comes in out of order, it's a bug. Have you seen this before?

Regarding A/V sync, the RTP client must send RTCP SR packets. It helps OME to keep A/V sync. All browsers supporting WebRTC send RTCP SR.

DorianRouby:
I'm not an expert on these subjects. I quickly looked at how packets were managed before.
You are right, I don't understand how we can get into the situation described above.

However, this is what I have observed and reproduced many times: "RTP timestamp is not monotonic". For my project, I pushed this fix into production a few months ago and I no longer had this problem. A/V sync was lost because video packets seemed to be more affected by this problem (so only the video stream was lagging).

Therefore, the origin of the problem could come from the reordering of RTP packets at the RtpFrameJitterBuffer level?
As soon as I have a little time, I can try to look in more detail at how and why I get into this situation.

getroot:
Thanks. One possibility is B-Frame.

PR #1528, opened 2024-02-17
Update prerequisites.sh

basisbit:
Updates bunch of dependencies:
```
OPENSSL_VERSION=3.0.13
SRTP_VERSION=2.5.0
SRT_VERSION=1.5.3
OPUS_VERSION=1.4
VPX_VERSION=1.14.0
FDKAAC_VERSION=2.0.3
NASM_VERSION=2.16.01
FFMPEG_VERSION=6.1.1
PCRE2_VERSION=10.43
```

I already did smoke-testing with a sample stream: compiling works, starting as origin works, starting as edge works, rtmp provider works, srt provider works, OVT publisher works, ovt provider works, WebRTC provider works over udp and over tcp, LLHLS provider works, HLS provider works, x264 transcoding works, AAC -> OPUS transcoding works and rtmp push to YouTube still works. So far I didn't find any new issues after applying these changes and rebuilding the docker images.

(These changes don't get Enhanced RTMP to work yet, that will require changes to the OME code for rtmp-push_session initialization, rtmp-export_chunk and also the code which feeds the stream data to ffmpeg.)
cc #1480


basisbit:
The ffmpeg configure options `--disable-lzo`, `--enable-network`, `--enable-dct` and `--enable-rdft` do not exist any more. The ffmpeg mailing list had an explanation that lzo check now is done through header-checks instead through a config flag, and the other three `--enable` options were replaced by being enabled by default.

getroot:
ffmpeg major version upgrades may affect hardware encoders, so take time to check. Is there a need to upgrade ffmpeg major?

trapasso:
@basisbit ... I'm testing ffmpeg 6, nice work.

I build on ubuntu 20.04 and nasm version 2.16.01 requires autoconf to version 2.71 or greater:
**"configure.ac:3: error: Autoconf version 2.71 or higher is required"**.

With the older version of nasm 2.15.05, it works fine.
Maybe it's better to wait a bit to change the nasm version.

Thank's for your job.

basisbit:
autoconf 2.71 is already 3 years...

trapasso:
The version of autoconf is 2.69, in the Ubuntu 20.04 repositories.
From here two possibilities: remove support for Ubuntu 20.04 or install it from the prerequisites.
What did you think you were doing?

PR #1536, opened 2024-02-23
Don't overwrite existing systemd service files in make install 

Wrhector:
Also display warning if not creating a config file


dimiden:
Thank you for this PR!

PR #1539, opened 2024-02-27
Added NetInt Quadra Support

n8o:
This PR adds basic support for NetInt Quadra.

- Added Docker Image
- Updated docs
- Tested AVC ABR encoding with 8bit and 10bit content.
- Decode, Scale, Encode all in hardware

Needs:
- Tests. Where are the unit tests?
- Test HEVC encoding and decoding
- Add JPEG and AV1 Support

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,41 @@
+FROM    ubuntu:20.04 AS base
+
+## Install libraries by package
+ENV     DEBIAN_FRONTEND=noninteractive
+RUN     apt-get update && apt-get install -y tzdata sudo curl
+
+RUN apt-get update \
+  && apt-get install -y build-essential \
+      gcc \
+      g++ \
+      gdb \
+      clang \
+      make \
+      ninja-build \
+      cmake \
+      autoconf \
+      automake \
+      libtool \
+      valgrind \
+      locales-all \
+      dos2unix \
+      rsync \
+      tar \
+      python \
+      python-dev \
+  && apt-get clean
+
+WORKDIR /tmp
+
+ENV     PREFIX=/opt/ovenmediaengine
+ENV     TEMP_DIR=/tmp/ome
+
+ARG     OME_VERSION=local
+ARG 	STRIP=TRUE
+
+## Copy code to image
+COPY    ./ ${TEMP_DIR}/
+
+## Install dependencies
+RUN \
+        ${TEMP_DIR}/misc/prerequisites.sh
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,3 +1,7 @@
+# Instruction: Build local OvenMediaEngine Dev Environment. Can be used with CLion dev environment as a toolchain
+# Information on hot to setup: https://www.jetbrains.com/help/clion/clion-toolchains-in-docker.html#sample-dockerfile
+
+
 FROM    ubuntu:20.04 AS base
 
 ## Install libraries by package
--- a/Dockerfile.niquadra.local
+++ b/Dockerfile.niquadra.local
@@ -103,9 +103,9 @@ RUN     apt-get update && apt-get install -y tzdata sudo
 WORKDIR         /opt/ovenmediaengine/bin
 EXPOSE          80/tcp 8080/tcp 8090/tcp 1935/tcp 3333/tcp 3334/tcp 4000-4005/udp 10000-10010/udp 9000/tcp
 COPY            --from=build /opt/ovenmediaengine /opt/ovenmediaengine
-COPY            --from=build /usr/local/lib/ /usr/local/lib/
+COPY            --from=build /usr/local/ /usr/local/
 RUN             ldconfig
 
 # Default run as Origin mode
-CMD             ["/opt/ovenmediaengine/bin/OvenMediaEngine", "-c", "origin_conf"]
+CMD             init_rsrc; /opt/ovenmediaengine/bin/OvenMediaEngine -c origin_conf
 
--- a/Dockerfile.niquadra.local
+++ b/Dockerfile.niquadra.local
@@ -1,0 +1,111 @@
+# Instruction: Build OvenMediaEngine with Netint Quadra support
+# Adapted From: https://github.com/NETINT-Technologies/quadra_dockerfile/blob/main/ffmpeg/Dockerfile
+# Usage:
+# $ docker build -t airensoft/ovenmediaengine_nv:dev -f Dockerfile.niquadra.local .
+# $ docker run -it --privileged --device /dev/nvme0n1 --device /dev/nvme0 airensoft/ovenmediaengine_nv:dev
+
+FROM ubuntu:20.04 AS Base
+
+ARG NI_RELEASE_VERSION=4.7.5
+ARG NI_PACKAGE_NAME="Quadra_V${NI_RELEASE_VERSION}.zip"
+ARG NI_PACKAGE_FOLDER="Quadra_V${NI_RELEASE_VERSION}"
+ARG RC=RC3
+#FFMPEG_VERSION can be: n3.4.2, n4.1.3, n4.2.1, n4.3, n4.3.1, n4.4, n5.0
+ARG FFMPEG_VERSION=n5.0
+
+
+ARG DEBIAN_FRONTEND=noninteractive
+SHELL ["/bin/bash", "-c"]
+
+#packages install
+RUN apt-get update
+RUN apt-get install -y pkg-config git gcc make g++ sudo wget uuid-runtime udev unzip
+
+#copy ni release package to docker /NI_Release directory
+COPY $NI_PACKAGE_NAME /NI_Release/
+WORKDIR /NI_Release
+
+#nvme cli install
+WORKDIR /NI_Release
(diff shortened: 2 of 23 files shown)


getroot:
Thanks for your contribution. Unfortunately we don't have this device so we can't test it. We have no choice but to rely on contributors for this equipment.

We don't have a unit test template yet, and each developer is doing his own testing.

AV1 cannot be tested because no publisher supports it yet. We will support this for WebRTC in the future.

@nums Could you please review this?

nums:
Yes, I will do a review.

I was also not able to test directly: unfortunately my netint card is a "logan" but not a "quadra"

nums:
hello @n8o , I did a quick check and I noticed this line:
https://github.com/n8o/OvenMediaEngine/blob/master/src/projects/transcoder/codec/encoder/encoder_avc_niquadra.cpp#L7
On the avc side it is commented but you kept it on the hevc.
On the Logan, I had to apply it on both.

It's probably normal but in case it's forgotten after a test...

n8o:
What line are you referring to? L7 is a copyright header line. I haven't actually tested HEVC decode and encode. I'll do that today and should be done before merging.

nums:
> What line are you referring to? L7 is a copyright header line. I haven't actually tested HEVC decode and encode. I'll do that today and should be done before merging.

sorry, bad copy and paste: https://github.com/n8o/OvenMediaEngine/blob/master/src/projects/transcoder/codec/encoder/encoder_avc_niquadra.cpp#L78

n8o:
I should be able to enable that line. I don't think it matters. I believe i disabled it while debugging issues before i got it working. the low latency setting is probably nice for webrtc. not sure if it affects quality or not for ABR HLS. 

I believe OME already forces keyframes when they are needed. so the keyframe param may not be needed. 
https://github.com/n8o/OvenMediaEngine/blob/59e7368a7494c6b1d7b2c434b70b5715a97fae32/src/projects/transcoder/codec/encoder/encoder_avc_niquadra.cpp#L40

nums:
For the `lowdelay` parameter I didn't see any difference but I added it just in case, however the `gopPresetIdx` allowed b-frames to be deactivated. (for the record, it seems to me that the gopPreset is at 4)

n8o:
I tried enabling it and got a bunch of errors. Looking at the documentation for quadra it reports:

**lowDelay**
Specifies whether or not to enable the low latency mode in encoding. When enabled, libxcoder increases its rate of polling the encoder and only permits buffering of a single frame to minimize the delay.

NOTE – That when enabled, the gopPresetIdx must be 1, 3, 7, 9, 10, or 0 with a consecutive order gop pattern, lookaheadDepth must be 0, and multicoreJointMode must be 0.


Supported values for **gopPresetIdx** are:

-1: Adaptive Gop (default)
0 : Custom Gop
1 : I-I-I-I,..I (all intra, gop_size=1)
3 : I-B-B-B,...B (consecutive B, gop_size=1)
4 : I-B-P-B-P,... (gop_size=2)
5 : I-B-B-B-P,... (gop_size=4)
7 : I-B-B-B-B,... (consecutive B, gop_size=4)
8 : I-B-B-B-B-B-B-B-B,... (random access, gop_size=8) 9 : I-P-P-P,... P (consecutive P, gop_size=1)
10 : I-P-P-P-P,... (hierarchical P, gop_size=4)

So with that, I set it to 1 and 1:
    ::av_opt_set(_codec_context->priv_data, "xcoder-params", "gopPresetIdx=1:lowDelay=1", 0);

that seems to work.

nums:
Sorry, a little bit busy this week.

@getroot, on my side it's ok and ready to be merge :)

n8o:
Actually i just noticed that gopPresetIdx 1 is all i-frames. I need to change that. Either idx 9 or 10?

nums:
yes, 10 seems to be the same than 2 (https://github.com/AirenSoft/OvenMediaEngine/blob/master/src/projects/transcoder/codec/encoder/encoder_avc_nilogan.cpp#L48C2-L48C17) on logan so better than 1

Dockerfile.niquadra.local
```diff
@@ -0,0 +1,111 @@
...
+RUN wget -c https://github.com/linux-nvme/nvme-cli/archive/refs/tags/v1.16.tar.gz
+RUN tar -xzf v1.16.tar.gz
+WORKDIR /NI_Release/nvme-cli-1.16/
+RUN make
+RUN make install
+RUN nvme list
+
+#yasm install
+WORKDIR /NI_Release
+RUN wget -c http://www.tortall.net/projects/yasm/releases/yasm-1.3.0.tar.gz
```
basisbit, line 39: please change this to https, there is no hash / checksum checking here.

misc/prerequisites.sh
```diff
@@ -302,6 +305,24 @@ install_ffmpeg()
...
+    if [ "$NETINT_QUADRA_HWACCELS" = true ] ; then
+        echo "we are applying the patch founded in $NETINT_QUADRA_PATCH_PATH"
+        patch_name=$(basename $NETINT_QUADRA_PATCH_PATH)
+        cp $NETINT_QUADRA_PATCH_PATH ${DIR}
+        cd ${DIR} && patch -t -p 1 < $patch_name
+        if [ "$NETINT_QUADRA_XCODER_COMPILE_PATH" != "" ] ; then
+            cd $NETINT_QUADRA_XCODER_COMPILE_PATH && bash build.sh && ldconfig #the compilation of libxcoder can be done before
+        fi
+        ADDI_LIBS+=" --enable-ni_quadra --enable-avfilter  --enable-pthreads "
+        ADDI_ENCODER+=",h264_ni_quadra,h265_ni_quadra"
```
basisbit, line 318: What are the license implications of this? Is releasing a pre-built docker container with this compliant wit AGPL?

src/projects/transcoder/filter/filter_rescaler.cpp
```diff
@@ -231,6 +232,28 @@ bool FilterRescaler::InitializeFilterDescription()
 		desc += ov::String::FormatString("multiscale_xma=lxlnx_hwdev=%d:outputs=1:out_1_width=%d:out_1_height=%d:out_1_rate=full",
 			_output_track->GetCodecDeviceId(), _output_track->GetWidth(), _output_track->GetHeight());
 	}
+    else if (output_module_id == cmn::MediaCodecModuleId::NIQUADRA)
+    {
+        switch (input_module_id)
+        {
+            case cmn::MediaCodecModuleId::NIQUADRA: { 	// Zero Copy
+                //TODO: Exception handling required if Device ID is different.
```
basisbit, line 240: Please finish this before merging, thanks

src/projects/transcoder/transcoder_stream.cpp
```diff
@@ -802,10 +802,10 @@ int32_t TranscoderStream::CreateEncoders(MediaFrame *buffer)
 			// These values are used in the Resampler/Rescaler filter.
 			if (output_track->GetMediaType() == cmn::MediaType::Video)
 			{
-				auto encoder = _encoders[encoder_id];
+				//auto encoder = _encoders[encoder_id];
+                output_track->SetColorspace(buffer->GetFormat());  // used AVPixelFormat
 
-				output_track->SetColorspace(encoder->GetSupportedFormat());
-			}
+            }
```
basisbit, line 808: was this indentation change on purpose? If not, please revert

basisbit (changes requested):
Just a few comments.

basisbit:
Also, might be worth it to update this PR to ffmpeg 6 or newer

Dockerfile.dev
```diff
@@ -0,0 +1,45 @@
+# Instruction: Build local OvenMediaEngine Dev Environment. Can be used with CLion dev environment as a toolchain
```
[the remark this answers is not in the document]
n8o: That makes sense. I will merge them and maybe add a comment on how to only do a build image vs release image. 
https://docs.docker.com/build/building/multi-stage/

Dockerfile.niquadra.local
```diff
@@ -0,0 +1,111 @@
+# Instruction: Build OvenMediaEngine with Netint Quadra support
```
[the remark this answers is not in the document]
n8o: I'll investigate this

misc/prerequisites.sh
n8o: I'm not sure about licensing and how this works for AGPL. wouldn't this be the same as all the others? Nvidia driver is proprietary. so is the logan and quadra.  OME uses it through ffmpeg which is compiled with nvidia and netint support. i do believe the one header file that ffmpeg uses for nvidia is open source. but everything else is close source. I'm not affiliated with netint. so i'm not sure

I just assumed since you already had logan support that quadra would be fine. this section is basically just a copy of how logan's prerequisites work..

getroot:
> Also, might be worth it to update this PR to ffmpeg 6 or newer

ffmpeg 6 will take some time for us.

catapop84:
Hello. Linode added instances with Netint Quadra T1U  in case someone needs to test it. I tried to compile it but without any luck.
 The documentation is for Logan which I think is the older version.

PR #1549, opened 2024-03-12
Add two environment variables

bchah:


getroot:
Thank you for PR.

I think it is a good idea to add HLS_CONTENT_TYPE_HEADER to the Server.xml configuration. If you add it to the Server.xml configuration, it would be better to configure it separately as HLSPlaylistContentTypeHeader and HLSSegmentContentTypeHeader. But I don't understand Rokue's player. Because vnd.apple.mpegurl is standard.
https://www.iana.org/assignments/media-types/application/vnd.apple.mpegurl

I think it would be better to remove the MAX_STREAM_WORKER_THREAD_COUNT restriction. I'm not sure if it makes sense to register it as an environment variable and use it.

bchah:
Thank you @getroot I agree with you on both items. Removing MAX_STREAM_WORKER_THREAD_COUNT restriction is pretty simple.

For Roku, here is what I discovered for their HLS compatibility:

- They only support the legacy HLS format provided by OME
- Namely, they only support .TS not .fmp4
- They seemed to require an argument of 'roku_trick_play=1' at the end of the URL

So in the end it wasn't really the MIME Type that was holding it back. If you are interested in Roku testing, they offer a free 'Roku Stream Tester' app that can link to a local Roku device and test any public HLS / DASH feed. Personally I never tried one until this customer had this particular need. It's an impressive little Roku ecosystem they offer!

PR #1611, opened 2024-04-24
Patch, schedulerChannel reading .ts file format, crash.

trapasso:
ScheduledChannel, when it reads a file, ts format, previously saved with startRecord, CRASH.
This P.R. solves the problem, simply by changing the bitstream type.

`bool use_annexb { std::strncmp(context->iformat->name, "mpegts", 6) == 0 };`

```
case cmn::MediaCodecId::H264:
	bitstream_format = (use_annexb) ? cmn::BitstreamFormat::H264_ANNEXB : cmn::BitstreamFormat::H264_AVCC;
	packet_type = cmn::PacketType::NALU;
	break;
case cmn::MediaCodecId::H265:
	bitstream_format = (use_annexb) ? cmn::BitstreamFormat::H265_ANNEXB : cmn::BitstreamFormat::HVCC;
	packet_type = cmn::PacketType::NALU;
	break;
```

Mpegts use annexb.
Many thanks in advance.

### DIFF
--- a/src/projects/modules/ffmpeg/ffmpeg_conv.h
+++ b/src/projects/modules/ffmpeg/ffmpeg_conv.h
@@ -382,46 +382,56 @@ namespace ffmpeg
 			{
 				case cmn::MediaCodecId::H264:
 				{
-					// AVCC format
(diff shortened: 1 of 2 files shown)


getroot:
Thank you!

PR #1618, opened 2024-05-04
ScheduledChannel: patchs for audio, add opus for mpegts and mp4.

trapasso:
This P.R. solves some audio problems on the scheduled provider.
It also adds the possibility of saving opus audio tracks on mpegts and mp4 containers.

1) Problem. When a file contains only one audio track the stream does not start. Patch to file "src/projects/publishers/file/file_session.cpp" fixes this problem.

2) Problem. When using a mpegts container with aac encoded audio, it doesn't work. Mpegts uses an AAC_ADTS type bitstream. Patch to the "src/projects/providers/scheduled/scheduled_stream.cpp" file resolves this issue.

3) Improvement. Added the ability to save opus audio tracks in mpegts and mp4 containers. This allows you to best use the "bypassTranscoder" parameter of the scheduledChannel API. All other changes add this feature.

4) Added a line to the .gitignore file, for Qt Creator.

Thanks in advance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ src/projects/main/git_info.h
 *.cxxflags
 *.files
 *.includes
+.qtc_clangd
 
 *.o
 *.d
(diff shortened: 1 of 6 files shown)


getroot:
Thank you so much, you're the best!

PR #1626, opened 2024-05-25
ffmpeg 7.0 Update: Use new channel layout

hashworks:
Context: At Arch Linux I build OvenMediaEngine against the system libraries. Since we currently rebuild stuff for ffmpeg 7.0 this patch is required.

Please bear with me here, I don't really code in C++.

Note: There was a flag `FF_API_OLD_CHANNEL_LAYOUT`, which has been dropped in v7.


src/projects/modules/ffmpeg/ffmpeg_conv.h
```diff
@@ -772,10 +772,15 @@ namespace ffmpeg
 				break;
 
 				case cmn::MediaType::Audio: {
-					codecpar->channels 				= static_cast<int>(media_track->GetChannel().GetCounts());
-					codecpar->channel_layout 		= ToAVChannelLayout(media_track->GetChannel().GetLayout());
-					codecpar->sample_rate 			= media_track->GetSample().GetRateNum();
-					codecpar->frame_size 			= (media_track->GetAudioSamplesPerFrame()!=0)?media_track->GetAudioSamplesPerFrame():1024;
+					uint64_t channel_layout = ToAVChannelLayout(media_track->GetChannel().GetLayout());
+					codecpar->ch_layout = {
+						.order = AVChannelOrder::AV_CHANNEL_ORDER_UNSPEC,
```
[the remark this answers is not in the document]
hashworks: Never mind, according to https://github.com/FFmpeg/FFmpeg/blob/n7.0/libavutil/channel_layout.h#L335 we must not use `AV_CHANNEL_ORDER_UNSPEC` if we provide a mask.

misc/prerequisites.sh
```diff
@@ -13,7 +13,7 @@ OPUS_VERSION=1.3.1
 VPX_VERSION=1.11.0
 FDKAAC_VERSION=2.0.2
 NASM_VERSION=2.15.05
-FFMPEG_VERSION=5.0.1
+FFMPEG_VERSION=7.0
```
hashworks, line 16: Note: I have not tested the `prerequisites.sh` script, it would be great if one could do so.

Especially since AirenSoft provides a patch for enterprise users.

hashworks:
This *compiles* with ffmpeg 7.0 and is currently used in the Arch Linux extra-staging repository: https://gitlab.archlinux.org/archlinux/packaging/packages/ovenmediaengine/-/commit/3a029703f7f9ce35d31487848efd9524516bf220

It is however entirely untested and will remain a draft until then.

misc/prerequisites.sh
basisbit: I will not work, because some of the flags used in the `prerequisites.sh` script for building ffmpeg don't exist any more in current versions of ffmpeg.

hashworks: Alright, tested out `podman build -f Dockerfile.local`:
* `Unknown option "--disable-lzo"` Disabling this isn't allowed anymore https://github.com/FFmpeg/FFmpeg/commit/df272928ff8041f87b92f4cf8dbf12a28fe9cdf7
* `Unknown option "--enable-dct"` Couldn't find the commit
* `Unknown option "--enable-rdft"` Couldn't find the commit

Removed them in commit a027c9af.

Afterward ffmpeg seems to build. jemalloc then fails on my end:
```
+ curl -sSLf https://github.com/jemalloc/jemalloc/releases/download/5.3.0/jemalloc-5.3.0.tar.bz2
+ tar -jx --strip-components=1
tar: .appveyor.yml: Cannot change ownership to uid 66878, gid 100: Invalid argument
...
```

Adding `--no-same-owner` to the tar extraction works here (commit 6bf52f0d).

The build succeeds with those changes.

basisbit (approved):
looks good to me. Hope this gets merged into main, because getting ffmpeg 5 to work gets more and more difficult on more current OS versions

hashworks:
ffmpeg 7.1 changed the API again: https://github.com/FFmpeg/FFmpeg/commit/7f17e0e6dd0b28b29295c1eeb1a38546eebc3946

`AVFilterLink` no longer has access to `hw_frames_ctx`, which moved to the internal `FilterLink`. I'm not sure what I should do with `SetHWFramesCtxOfAVFilterLink` ([ffmpeg_conv.h#L958](https://github.com/AirenSoft/OvenMediaEngine/blob/9b01f66afa8f45deee945baab4dbbef542a6618c/src/projects/modules/ffmpeg/ffmpeg_conv.h#L958)):

```cpp
context->hw_frames_ctx = av_buffer_ref(hw_frames_ref);

av_buffer_unref(&hw_frames_ref);
```

I could drop those lines, but I'm not sure if that breaks functionality.

Any ideas?

hashworks:
> ([ffmpeg_conv.h#L958](https://github.com/AirenSoft/OvenMediaEngine/blob/9b01f66afa8f45deee945baab4dbbef542a6618c/src/projects/modules/ffmpeg/ffmpeg_conv.h#L958))

@Keukhan you added this in 1de467eb6136a7c195fb1424fb6ab6f87045ceaf, can you take a look?

Keukhan:
> > ([ffmpeg_conv.h#L958](https://github.com/AirenSoft/OvenMediaEngine/blob/9b01f66afa8f45deee945baab4dbbef542a6618c/src/projects/modules/ffmpeg/ffmpeg_conv.h#L958))
> 
> @Keukhan you added this in [1de467e](https://github.com/AirenSoft/OvenMediaEngine/commit/1de467eb6136a7c195fb1424fb6ab6f87045ceaf), can you take a look?

First of all, thank you for the PR. I will internally discuss changing FFmpeg to 7.x, and after that, I will modify the code related to hw_frames_ctx of AVFilterList.

hashworks:
Since this blocks the ffmpeg 7.1 release on Arch I've moved the package to our `ffmpeg4.4` package: https://gitlab.archlinux.org/archlinux/packaging/packages/ovenmediaengine/-/commit/d58322c814222a09ceb6d731905efe5bc3b632f5

dx314:
Is this still going to be merged in?

Remboooo:
The dependency on ffmpeg 5.0.1 is currently also preventing OvenMediaEngine from building on Ubuntu 24.04 LTS because it does not build with binutils >= 2.41. Would love this patch to ship and also to move `misc/prerequisites.sh` to a recent ffmpeg.

Right now I've locally applied a workaround by changing the ffmpeg version in `misc/prerequisites.sh` to 5.1.6, which also required removing the `--disable-lzo` flag.

PR #1694, opened 2024-09-07
Normalize HLSChunklistPathDepth attribute

mauricioabreu:
When setting a profile (transcoding) the option to set the depth of the playlist path uses HLSChunklistPathDepth, however, when using multiplexing approach it uses HlsChunklistPathDepth.

It might be a good idea using the same atribute name.

### DIFF
--- a/src/projects/providers/multiplex/multiplex_profile.cpp
+++ b/src/projects/providers/multiplex/multiplex_profile.cpp
@@ -418,7 +418,7 @@ namespace pvd
 					playlist->SetWebRtcAutoAbr(webrtc_auto_abr.text().as_bool());
 				}
 
-				auto hls_chunklist_path_depth = options_node.child("HLSChunklistPathDepth");
+				auto hls_chunklist_path_depth = options_node.child("HLSChunklistPathDepth") || options_node.child("HlsChunklistPathDepth");
 				if (hls_chunklist_path_depth)
 				{
 					playlist->SetHlsChunklistPathDepth(hls_chunklist_path_depth.text().as_int());
@@ -618,9 +618,10 @@ namespace pvd
 					options_node.append_child("WebRtcAutoAbr").text().set(playlist->IsWebRtcAutoAbr());
 				}
 
-				if (playlist->GetHlsChunklistPathDepth() != -1)
+				int hls_chunklist_path_depth = playlist->GetHlsChunklistPathDepth();
+				if (hls_chunklist_path_depth != -1)
 				{
-					options_node.append_child("HLSChunklistPathDepth").text().set(playlist->GetHlsChunklistPathDepth());
+					options_node.append_child("HLSChunklistPathDepth").text().set(hls_chunklist_path_depth);
 				}
 
 				if (playlist->IsTsPackagingEnabled())
--- a/src/projects/providers/multiplex/multiplex_profile.cpp
+++ b/src/projects/providers/multiplex/multiplex_profile.cpp
@@ -126,9 +126,9 @@ namespace pvd
 			_last_error = "Failed to find outputStream/name object";
 			return false;
 		}
-		
+
 		_output_stream_name = name_object.asString().c_str();
-		
+
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


getroot:
Thank you! This is my mistake.
However, since there are already people using Hls~~, I think it would be better to allow both HLS~~~ and Hls~~~ to be used in Mux. And if there are both, I think the priority should be HLS~~ higher.
Can you update this so that it is possible?

mauricioabreu:
Hello @getroot yes, sure!

mauricioabreu:
Updated @getroot 

This new approach maintains backward compatibility for reading existing configurations while encouraging the use of the new attribute name in all new or updated configurations, so HLSChunklistPathDepth will be used organically.

src/projects/providers/multiplex/multiplex_profile.cpp
```diff
@@ -418,7 +418,7 @@ namespace pvd
 					playlist->SetWebRtcAutoAbr(webrtc_auto_abr.text().as_bool());
 				}
 
-				auto hls_chunklist_path_depth = options_node.child("HlsChunklistPathDepth");
+				auto hls_chunklist_path_depth = options_node.child("HLSChunklistPathDepth") || options_node.child("HlsChunklistPathDepth");
```
getroot, line 421: Here, hls_chunklist_path_depth is a bool type. Is that what you intended?

src/projects/providers/multiplex/multiplex_profile.cpp
```diff
@@ -618,9 +618,10 @@ namespace pvd
 					options_node.append_child("WebRtcAutoAbr").text().set(playlist->IsWebRtcAutoAbr());
 				}
 
-				if (playlist->GetHlsChunklistPathDepth() != -1)
+				int hls_chunklist_path_depth = playlist->GetHlsChunklistPathDepth();
```
getroot, line 621: Please do not edit this part unless absolutely necessary.

src/projects/providers/multiplex/multiplex_profile.cpp
```diff
@@ -511,12 +511,12 @@ namespace pvd
 				_last_error = ov::String::FormatString("Failed to parse url: %s", url_str.CStr());
 				return false;
 			}
-			
+
```
getroot, line 514: I think it would be nice if there were no more edits that would add blank spaces like this.

src/projects/providers/multiplex/multiplex_profile.cpp
```diff
@@ -418,7 +418,7 @@ namespace pvd
 					playlist->SetWebRtcAutoAbr(webrtc_auto_abr.text().as_bool());
 				}
 
-				auto hls_chunklist_path_depth = options_node.child("HlsChunklistPathDepth");
+				auto hls_chunklist_path_depth = options_node.child("HLSChunklistPathDepth") || options_node.child("HlsChunklistPathDepth");
 				if (hls_chunklist_path_depth)
 				{
 					playlist->SetHlsChunklistPathDepth(hls_chunklist_path_depth.text().as_int());
```
getroot, line 424: hls_chunklist_path_depth.text() seems to result in a syntax error because hls_chunklist_path_depth is a bool.

getroot (changes requested):

src/projects/providers/multiplex/multiplex_profile.cpp
mauricioabreu: It is actually removing extra spaces. It is probably my editor doing it. I can keep those extra spaces, no worries.

src/projects/providers/multiplex/multiplex_profile.cpp
mauricioabreu: No. It is my fault. I wll fix it

getroot (approved):

getroot:
Great work, thank you.

mauricioabreu:
@getroot thank you for the awesome project.

PR #1698, opened 2024-09-20
Adjust EXT-X-TARGETDURATION to round target duration

fingul:
In protocol version 6, EXT-X-TARGETDURATION requires rounding the target duration to the nearest integer. This change ensures compliance by casting and rounding the target duration before formatting it into the playlist string.


getroot:
Thanks for the PR.

1. publishers/hls is for HLS V3 only. LLHLS is in the publihsers/llhls folder.
2. _config.target_duration is of type size_t and is a value that the user sets as an integer. This is not a float, so I don't know why std::round is needed.

fingul:
Sorry.  My fault. At LL-HLS segment duration 4.8, It`s EXT-X-TARGETDURATION set to 4. (Not 5). I will investigate more.

( In protocol version 6, EXT-X-TARGETDURATION requires rounding the target duration to the nearest integer. This change ensures compliance by casting and rounding the target duration before formatting it into the playlist string. )

PR #1733, opened 2024-10-31
fix(gcc_version build): gcc_version is always empty

Linmengoy:
shell:gcc -v 2>&1 | tail -1 | awk ' { print $3 } ' 
output: x.x.x
GCC_VERSION := $(word 3, $(GCC_VERSION))
GCC_VERSION is always '' ,then it always goes to the else


dimiden:
@Linmengoy 
On Ubuntu 20.04, the last line of `gcc -v` is `gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.2)`, so that line is essential.
What environment did you test on? How does the `gcc -v` output appear in that environment?

dimiden (changes requested):
This modification breaks the retrieval of the `GCC_VERSION` value on Ubuntu.

Issue #1739, opened 2024-11-14
Dev image does not send proper url for webhook callbacks


naanlizard:
**Describe the bug**
On the latest dev image (`airensoft/ovenmediaengine@sha256:773acc79230548f8d942b2321f94d10e0846135128a5b0dc35eab60ff0fa528e`), admission webhooks do not work because the URL sent does not include the streamkey

**To Reproduce**
[Server.xml](https://gist.github.com/naanlizard/8c1ce4fe2ae9f09bda21a17f69a09f63)

(Note, in Server.xml, the Decodes section is either commented out (for 0.17.1) or uncommented (for dev)

[Dockerfile](https://gist.github.com/naanlizard/73927d1e198ca8f48e585064e4eec4bb)

Maybe relevant, if you put the streamkey in the URL (e.g. in OBS, your rtmp url is set to `rtmp://ingest.hostname.tld:1935/live/streamkey`) both versions work. 

**Expected behavior**
URL sent should include the streamkey, like on 0.17.1.

**Logs**
Relevant log lines from our backend API, Parameters being the values from the request OME sends

0.17.1
`
Parameters: {"client"=>{"address"=>"x", "port"=>58323, "real_ip"=>"x"}, "request"=>{"direction"=>"incoming", "protocol"=>"RTMP", "status"=>"opening", "time"=>"2024-11-14T15:23:54.039+00:00", "url"=>"rtmp://ingest.hostname.tld:1935/live/streamkey"}, "stream"=>{}}
`

dev
`Parameters: {"client"=>{"address"=>"x", "port"=>58619, "real_ip"=>"x"}, "request"=>{"direction"=>"incoming", "protocol"=>"RTMP", "status"=>"opening", "time"=>"2024-11-14T15:28:11.658+00:00", "url"=>"rtmp://ingest.hostname.tld:1935/live"}, "stream"=>{}}`

Additional info:

~~Interestingly, it seems like the closing webhook *does* send the streamkey in the URL~~

~~`Request params: {"direction"=>"incoming", "new_url"=>"rtmp://ingest.hostname.tld:1935/live/streamkey", "protocol"=>"RTMP", "status"=>"closing", "time"=>"2024-11-14T15:27:15.762+00:00", "url"=>"rtmp://apollo-rtmp://ingest.hostname.tld:1935/live/streamkey"}`~~

Double checked the above, it seems like it does NOT send the streamkey in the closing call, or at least not always. Strange

[...]

PR #1751, opened 2024-12-16
(fix) Fixed an issue that admission webhook url is not proper

negtak:
Fix: set rtmp tokens, only if url scheme is rtmp

relative issue: #1739

### DIFF
--- a/src/projects/base/ovlibrary/url.cpp
+++ b/src/projects/base/ovlibrary/url.cpp
@@ -120,14 +120,13 @@ namespace ov
 
 	bool Url::SetPath(const ov::String &path)
 	{
-		_path = path;
 
 		_file = "";
 		_stream = "";
 		_app = "";
 
 		// split <path> to /<app>/<stream>/<file> (4 tokens)
-		auto tokens = _path.Split("/");
+		auto tokens = path.Split("/");
 
 		switch (tokens.size())
 		{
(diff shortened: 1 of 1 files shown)


getroot:
Thanks for the PR.

What problem does this solve?

ovlibrary should be implemented in a way that is as independent of specific protocols or modules as possible. So it seems like we need to use a different approach.

negtak:
@getroot 

OvenMediaEngine use the same url library for both rtmp and webhook url.
So if we set the admission webhook like `http://localhost/path/to/admission/webhook`, it request to `http://localhost/path/to/admission`.

```
[2024-12-16 05:47:36.295] I [SPRTMP-t1935:14] AccessController | access_controller.cpp:100  | AdmissionWebhooks queried http://host.docker.internal/path/to/admission/webhook whether client 172.27.0.1:50542 could access rtmp://localhost:1935/app/5d087ce095600934a86ed695c39fd77ea419000860. (Result : Reject Elapsed : 2 ms)
```

```
$ python3 -m http.server 80
Serving HTTP on :: port 80 (http://[::]:80/) ...
::ffff:127.0.0.1 - - [16/Dec/2024 14:59:28] code 501, message Unsupported method ('POST')
::ffff:127.0.0.1 - - [16/Dec/2024 14:47:45] "POST /path/to/admission HTTP/1.1" 501 -
```

After patched

```
[2024-12-16 05:59:28.611] I [SPRTMP-t1935:14] AccessController | access_controller.cpp:198  | AdmissionWebhooks queried http://host.docker.internal/path/to/admission/webhook whether client 172.27.0.1:60872 could access rtmp://localhost:1935/app. (Result : Reject Elapsed : 8 ms)
```

```
$ python3 -m http.server 80
Serving HTTP on :: port 80 (http://[::]:80/) ...
::ffff:127.0.0.1 - - [16/Dec/2024 14:59:28] code 501, message Unsupported method ('POST')
::ffff:127.0.0.1 - - [16/Dec/2024 14:59:28] "POST /path/to/admission/webhook HTTP/1.1" 501 -
```

> ovlibrary should be implemented in a way that is as independent of specific protocols or modules as possible. So it seems like we need to use a different approach.

I agree with this opinion.

getroot:
Thank you. 

After checking, I found out that this is a regression bug. I think it would be better to just remove the part that resets _path in the UpdateUrl function.

@dimiden Could you check this?

negtak:
The bug was fixed in v0.17.3. Thank you!

PR #1779, opened 2025-02-16
TLS support added to the docker compose file

Johnr24:
TLS documetation needs updating and this is a great start,

### DIFF
--- a/docs/configuration/tls-encryption.md
+++ b/docs/configuration/tls-encryption.md
@@ -2,6 +2,9 @@
 
 Most browsers can't load resources via HTTP and WS (WebSocket) from HTTPS web pages secured with TLS. Therefore, if the player is on an HTTPS page, the player must request streaming through "https" and "wss" URLs secured with TLS. In this case, you must apply the TLS certificate to the OvenMediaEngine.
 
+### Docker
+You can enable the #commented out setting in volumes to link certificates from the host machine, from the docker compose file or flag that through the creation command, 
+
 You can set the port for TLS in `TLSPort`. Currently, LLHLS and WebRTC Signaling support TLS.
 
 ```markup
@@ -78,7 +81,22 @@ To enable HTTP for HLS and WebRTC signaling servers, you must enable the TLS ele
 
 Assuming the certificate settings are correctly configured, WebRTC streaming can then be played via the wss://url protocol, while LLHLS streaming can be accessed via [https://url](https://url/).
 
+### Let's Encrypt, 
 
+It is likely that you used certbot to create your certificates, if so you likely generated files in a .PEM format if this is the case the correct syntax for these in your Server.xml is 
 
+```markup
+			<!-- Settings for multi ip/domain and TLS -->
+			<Host>
+				<Names>
+					<Name>live.johnrogerscolour.co.uk</Name>
+				</Names>
+				<TLS>
+					<CertPath>/opt/ovenmediaengine/bin/cert/live/example.com/cert.pem</CertPath>
+					<KeyPath>/opt/ovenmediaengine/bin/cert/live/example.com/privkey.pem</KeyPath>
+					<ChainCertPath>/opt/ovenmediaengine/bin/cert/live/example.com/chain.pem</ChainCertPath>
+				</TLS>
+			</Host>
+```
 
 
--- a/docs/configuration/tls-encryption.md
+++ b/docs/configuration/tls-encryption.md
@@ -81,14 +81,13 @@ To enable HTTP for HLS and WebRTC signaling servers, you must enable the TLS ele
 Assuming the certificate settings are correctly configured, WebRTC streaming can then be played via the wss://url protocol, while LLHLS streaming can be accessed via [https://url](https://url/).
 
 ### Let's Encrypt, 
-
-It is likely that you used certbot to create your certificates, if so you likely generated files in a .PEM format if this is the case the correct syntax for these in your Server.xml is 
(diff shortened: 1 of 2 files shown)


bchah (changes requested):
@Johnr24 can you please update the server config examples with the implied ./certs path?

Johnr24:
> @Johnr24 can you please update the server config examples with the implied ./certs path?

are you talking about the Server.xml file?

bchah:
> are you talking about the Server.xml file?

Server.xml, Edge.xml, Origin.xml yes - so the default config will pick up the mapped directory.

getroot:
There is also something about certificate paths in the default settings in the manual.

https://github.com/AirenSoft/OvenMediaEngine/blob/master/docs/getting-started/getting-started-with-docker.md

I think we need to match the locations of these.

Johnr24:
I've tweaked the docs those arent in this branch of the PR, to talk about this more, These where a couple of the things I struggled with when setting up OME via docker compose,  - https://github.com/Johnr24/OvenMediaEngine/tree/docs

the location will depend on where you put them, certbot for example generates aliases in a live folder and then the archive folder which has the actual certificates, one of the issues with this approach is if you only link the live folder and not the parent folder, the container doesnt see the certificates, you can see this when you call on them through 'cat' via docker exec. 

I've also created a different docker compose file which has a caddy container, and a oliefa scheduler container, the schedualer container runs start and stop commands on the caddy container every 85 days. caddy has one job which is to generate certificates, which are then called on by OME, 

the reason for the schedular container is to not bind ports 80 or 443 on that machine.

getroot:
Great. I think your document will be helpful to many beginners. If you PR it, @bchah will review and merge it.

Johnr24:
> @Johnr24 Great. I think your document will be helpful to many beginners. If you PR it, @bchah will review and merge it.


That's all done,

docs/configuration/tls-encryption.md
```diff
@@ -78,7 +81,22 @@ To enable HTTP for HLS and WebRTC signaling servers, you must enable the TLS ele
...
 
+### Let's Encrypt, 
 
+It is likely that you used certbot to create your certificates, if so you likely generated files in a .PEM format if this is the case the correct syntax for these in your Server.xml is 
 
+```markup
+			<!-- Settings for multi ip/domain and TLS -->
+			<Host>
+				<Names>
+					<Name>live.johnrogerscolour.co.uk</Name>
```
bchah, line 92: Please change to vanilla domain like "live.example.com"

docs/configuration/tls-encryption.md
```diff
@@ -78,7 +81,22 @@ To enable HTTP for HLS and WebRTC signaling servers, you must enable the TLS ele
...
+				</Names>
+				<TLS>
+					<CertPath>/opt/ovenmediaengine/bin/cert/live/example.com/cert.pem</CertPath>
```
bchah, line 95: The domain here should match the name specified above (live.example.com)

docs/configuration/tls-encryption.md
```diff
@@ -78,7 +81,22 @@ To enable HTTP for HLS and WebRTC signaling servers, you must enable the TLS ele
 
 Assuming the certificate settings are correctly configured, WebRTC streaming can then be played via the wss://url protocol, while LLHLS streaming can be accessed via [https://url](https://url/).
 
+### Let's Encrypt, 
 
+It is likely that you used certbot to create your certificates, if so you likely generated files in a .PEM format if this is the case the correct syntax for these in your Server.xml is
```
bchah, line 86: We do not know for sure how users will generate their certificates. Please rephrase to something like:

"If you used certbot to create your certificates, the PEM files it creates can be linked in your Server.xml like this:"

docs/configuration/tls-encryption.md
```diff
@@ -2,6 +2,9 @@
 
 Most browsers can't load resources via HTTP and WS (WebSocket) from HTTPS web pages secured with TLS. Therefore, if the player is on an HTTPS page, the player must request streaming through "https" and "wss" URLs secured with TLS. In this case, you must apply the TLS certificate to the OvenMediaEngine.
 
+### Docker
+You can enable the #commented out setting in volumes to link certificates from the host machine, from the docker compose file or flag that through the creation command,
```
bchah, line 6: Please correct grammar: "To link certificates from your Docker host, uncomment the example in the Docker compose file or manually connect a volume in the Docker run command, e.g. `-v ~/local/cert/path:/opt/ovenmediaengine/bin/certs`"

bchah (changes requested):
- Please see my inline comments regarding grammar and domain names.
- Please edit the TLS examples in Server.xml, Origin.xml, Edge.xml to match your Docker compose changes.

Johnr24:
1) I've done all of these, I've kept it to example.com rather than live.example.com as having live/live next to each other is confusing and frankly less-accessible/ND friendly. 

2) I'm not sure what you mean, the origin and edge xmls are upto date and the TLS additions match to them? as for certifcates 
as I wrote here https://github.com/AirenSoft/OvenMediaEngine/pull/1779#issuecomment-2667119203

It really depends on how these where generated and as you said we don't know.

bchah (approved):

bchah:
Thanks @Johnr24 PR approved - the examples in Server.xml, Origin.xml and Edge.xml are inconsistent but these can be changed later.

Johnr24:
When does this get merged?

PR #1804, opened 2025-03-15
Add OpenAPI spec for GET stream information

leporel:
Because the example in the documentation does not show all possible fields


getroot:
We sincerely appreciate your contribution.

PR #1929, opened 2025-09-22
fix: dereferenced output_path variable

c-cabral:
In my last PR #1884, I introduced a bug by dereferencing `output_path` when passing it to the setter. The effect was no dumps were being saved, since no path was provided.


getroot:
Thank you

Issue #1935, opened 2025-10-23
RTSP Pull cannot authenticate


Karl-dubidot:
**Describe the bug**
We are using the latest docker build. For the RTSP cameras we use the streaming API and it worked well.
We have now a customer with RTSP cameras we cannot connect with the streaming API.
The message from the cameras are 401 (not authorized).
The same cameras have no problems with VLC.
The cameras are VuStream 150 and are able to perform Basic and Digest auth.

**To Reproduce**
Our customer has provided one of his cameras with a public ip address.
We can send you the public rtsp-url via email.
Then send a POST request to start the stream to the OME stream api.

**Expected behavior**
OME can connect and authorize.

**Logs**
The relevant entries in the origin log are:
[2025-10-23 09:58:05.486] W [SPAPISvr-t8081:38] RtspMessage | rtsp_message.cpp:157  | Could not parse header field (WWW-Authenticate: Basic realm="YUAN"
WWW-Authenticate: Digest realm="YUAN", nonce="696c48bacc1aba320705e7ee58e54680", stale="FALSE"
Content-Length: 0)

[2025-10-23 09:58:05.579] E [SPAPISvr-t8081:38] RtspcStream | rtspc_stream.cpp:277  | Rtsp server(rtsp://xxxxxxx:yyyyyy@vuzzzzzzz150.me:554/S1) rejected the describe request : 401(Unauthorized)

**Server (please complete the following information):**
 latest docker image (tried both :dev and :latest)

**Player (please complete the following information):**
not relevant

**Additional context**

getroot:
Closing this issue since the related PR was closed for some reason. If this gets picked up again, please reopen this issue or open a new one.

PR #1936, opened 2025-10-23
FIX: RTSP Pull cannot authenticate #1935. 

Luteva-ssh:
https://github.com/AirenSoft/OvenMediaEngine/issues/1935 seems to show a non-RFC-compliant header: a missing comma. This should ideally be fixed by YUAN High‑Tech Development Co., Ltd., but in the meantime the following workaround may (or may not) resolve the issue.

### DIFF
--- a/src/projects/transcoder/codec/decoder/decoder_avc_nv.cpp
+++ b/src/projects/transcoder/codec/decoder/decoder_avc_nv.cpp
@@ -37,6 +37,9 @@ bool DecoderAVCxNV::InitCodec()
 	if (hw_device_ctx == nullptr)
 	{
 		logte("Could not get hw device context for %s", cmn::GetCodecIdString(GetCodecID()));
+
+		::avcodec_free_context(&_codec_context);
+
 		return false;
 	}
 
@@ -44,6 +47,9 @@ bool DecoderAVCxNV::InitCodec()
 	if (ffmpeg::compat::SetHwDeviceCtxOfAVCodecContext(_codec_context, hw_device_ctx) == false)
 	{
 		logte("Could not set hw device context for %s", cmn::GetCodecIdString(GetCodecID()));
+
+		::avcodec_free_context(&_codec_context);
+
 		return false;
 	}
 
@@ -198,6 +204,9 @@ void DecoderAVCxNV::CodecThread()
 				else if (ret < 0)
 				{
 					logte("Error error occurred while sending a packet for decoding. reason(%s)", ffmpeg::compat::AVErrorToString(ret).CStr());
+
+					Complete(TranscodeResult::DataError, nullptr);
+					
 					break;
 				}
 			}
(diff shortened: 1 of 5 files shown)


Luteva-ssh:
The “non-RFC-compliant” part was unfortunately misinformation — sorry about that! The “new line” should actually work.
Nevertheless, the workaround should be able to parse the given header.

Ideally, it would be configurable which authentication method to use when multiple are available. However, I’m not sure where that can be configured. Maybe someone from the team can help me out here?

Karl-dubidot:
Tried your patch but unfortunately it doesnt work.
The error message has changed, now we get:
[2025-10-24 15:56:14.690] E [SPAPISvr-t8081:37] RtspcStream | rtspc_stream.cpp:259  | No response(CSeq : 0) was received from the rtsp server(rtsp://xxxxx:zzzzz@wwwww.ccccc.me:554/S1)

The other (working) rtsp cameras are not affected.
Thank you for your help

Karl-dubidot:
Your patch works!
Unfortunately the public rtsp was temporarily not available.
So the last test was not working.
Such a silly error!

Luteva-ssh:
I'm glad to hear that the patch works!
However, I think it would be better to have a cleaner solution.
Ideally, the desired authentication method should be configurable.
Unfortunately, I'm not familiar enough with the project to implement this properly — so far, I've only adjusted the parser.
Could someone please help me and point out where such a configuration would best be added?

getroot (changes requested):
@Luteva-ssh 

Thank you so much for your PR.

It looks like the log shows line breaks using only \n instead of proper CRLF (\r\n) sequences. This is a violation of the specification.

```
[2025-10-23 09:58:05.486] W [SPAPISvr-t8081:38] RtspMessage | rtsp_message.cpp:157 | Could not parse header field (WWW-Authenticate: Basic realm="YUAN"
WWW-Authenticate: Digest realm="YUAN", nonce="696c48bacc1aba320705e7ee58e54680", stale="FALSE"
Content-Length: 0)
```

As you can see, multiple header fields appear on a single line because only \n is being used as a separator.
The first thing we need to decide is whether to allow \n (LF) as a valid line delimiter in addition to CRLF.
To support that, we’ll need to modify this part of the code:

https://github.com/AirenSoft/OvenMediaEngine/blob/ea0f64df618c0955550301e1d862e270066db71f/src/projects/modules/rtsp/rtsp_message.cpp#L107

Once that’s fixed, the authentication issue should be resolved with the rtsp camera.
Even though the current RTSP module doesn’t support multiple headers with the same name (that’s an OME limitation, not a protocol violation), the client should still work correctly using the first Basic authentication field.

If we want to properly support multiple headers of the same name, we can change the container here to use a multimap:

https://github.com/AirenSoft/OvenMediaEngine/blob/ea0f64df618c0955550301e1d862e270066db71f/src/projects/modules/rtsp/rtsp_message.h#L122

And to prioritize **Digest** over **Basic**, we should modify the logic at the following location:

https://github.com/AirenSoft/OvenMediaEngine/blob/4a18ba6784a4c3669291cb5338ca65a299818bd2/src/projects/providers/rtspc/rtspc_stream.cpp#L273

PR #1949, opened 2025-11-14
FIX: Support '\n' as line breaks + Digest authentication should be preferred over Basic authenticaton

Luteva-ssh:
accept \n as line breaks. This is not the standard. But the old versiojn of rtsp (1996) says, that:
"receivers should be prepared to also interpret CR and LF"
[4](https://www.rfc-editor.org/rfc/rfc2326#section-4)
____________
RTSP Message
   RTSP is a text-based protocol and uses the ISO 10646 character set in
   UTF-8 encoding ([RFC 2279](https://www.rfc-editor.org/rfc/rfc2279) [21]). Lines are terminated by CRLF, but
   receivers should be prepared to also interpret CR and LF by
   themselves as line terminators.
____________
It is not "good practice", I know. But as this doesn't hurt... we should also accept (quite) old versions of the protocoll. :)

### DIFF
--- a/src/projects/modules/rtsp/rtsp_message.cpp
+++ b/src/projects/modules/rtsp/rtsp_message.cpp
@@ -1,5 +1,6 @@
 #include "rtsp_message.h"
 #include "header_fields/rtsp_header_field_parser.h"
+#include "header_fields/rtsp_header_authenticate.h"
 
 #define OV_LOG_TAG "RtspMessage"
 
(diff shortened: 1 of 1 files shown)


Luteva-ssh:
The second commit should prioritize digest over basic auth.

src/projects/modules/rtsp/rtsp_message.cpp
```diff
@@ -223,10 +228,41 @@ RtspMessage::RtspMessage(uint32_t status_code, uint32_t cseq, ov::String reason_
 // For example, currently only the first of WWW-Authenticate is processed.
 bool RtspMessage::AddHeaderField(const std::shared_ptr<RtspHeaderField> &field)
```
getroot: Since `RtspMessage` is a general-purpose class, adding special-case handling inside it would hurt readability.  
It would be better to change  
`std::map<ov::String, std::shared_ptr<RtspHeaderField>> _header_fields`  
to `std::multimap`, and handle the priority at  
[`rtspc_stream.cpp` line 273 (the link above).](https://github.com/AirenSoft/OvenMediaEngine/blob/4a18ba6784a4c3669291cb5338ca65a299818bd2/src/projects/providers/rtspc/rtspc_stream.cpp#L273)

getroot (changes requested):
Thank you for opening the PR. It looks good, but it needs a few adjustments before it can be merged.

src/projects/modules/rtsp/rtsp_message.cpp
getroot: If you’re not familiar with multimap, please revert the commit that changes the auth method priority. The newline handling alone will still allow the initial authentication to work (SDP will use the first method it finds), and we’ll take care of the multimap part later.

getroot:
@Luteva-ssh Hi, Happy new year!
Just checking in on this PR. Any update?

PR #1961, opened 2025-11-26
feat: Added video/audio index hint to multiplex playlists

mbakumenkov:
## Description
This PR adds support for `videoIndexHint` and `audioIndexHint` parameters in multiplex playlist renditions, allowing explicit control over track selection when multiple tracks of the same type are available in a variant.

## Motivation
When working with multiplex channels that have multiple video or audio tracks in a single variant, there was no way to specify which track should be used for a particular rendition. This enhancement allows users to explicitly specify track indices through configuration.
### Configuration Support
- Added parsing for `VideoIndexHint` and `AudioIndexHint` fields in XML configuration (`multiplex_profile.cpp`)- Added parsing for `videoIndexHint` and `audioIndexHint` fields in JSON configuration (`multiplex_profile.cpp`)

### Stream Processing
- Updated `LLHlsStream::CreateMasterPlaylist` to use the specified track indices when selecting video and audio tracks for renditions (`llhls_stream.cpp`)
- Changed from `GetFirstTrackByVariant()` to `GetTrackByVariant(variant_name, index_hint)` for both video and audio tracks

### DIFF
--- a/src/projects/publishers/llhls/llhls_stream.cpp
+++ b/src/projects/publishers/llhls/llhls_stream.cpp
@@ -530,7 +530,7 @@ std::shared_ptr<LLHlsMasterPlaylist> LLHlsStream::CreateMasterPlaylist(const std
 		{
 			video_index_hint = 0;
 		}
-		auto video_track = GetTrackByVariant(rendition->GetVideoVariantName(), video_index_hint);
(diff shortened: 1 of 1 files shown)


getroot (changes requested):
Thanks for the PR!

getroot:
Thank you!

PR #1992, opened 2026-02-08
Fix missing port in SignedPolicy documentation example

0xjemm:
The final URL example was missing the :3333 port, contradicting the warning that ports must always be included when generating signatures. Also added a domain name example to clarify that the port requirement applies to domain names as well, not just IP addresses.


getroot:
Thank you!

PR #1999, opened 2026-02-19
Added rtsps support

devedse:
I've added RTSPS support. I've tested this with Unifi camera's and it works.

Please have a good look at the PR to see if this is as desired. Obviously the github actions stuff needs to be omitted (I just used that to test my own changes).

### DIFF
--- a/docs/live-source/webrtc.md
+++ b/docs/live-source/webrtc.md
@@ -46,16 +46,19 @@ WebRTC input can be turned on/off for each application. As follows Setting enabl
 <Providers>
     ...
     <WebRTC>
-        ...
         <Timeout>30000</Timeout>
+        <FIRInterval>3000</FIRInterval>
         <CrossDomains>
             <Url>*</Url>
         </CrossDomains>
-        ...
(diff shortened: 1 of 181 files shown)


getroot:
Thank you for the PR. Since there are many changes, it may take some time to review.

Before starting a full review, it would be helpful if you could clean up the changes unrelated to `RTSPS` support as much as possible. For example, files like `githubactionsbuilds.yml` do not seem to be related to `RTSPS`. Also, the changes inside the `Stream::AdjustRtpTimestamp` function do not appear to be related to `RTSPS`.

It looks like several topics are mixed into this PR. If those changes are necessary, could you please separate them into different PRs by topic?

devedse:
@getroot  I've split the PR into multiple.

The one I'm most uncertain about is the one with the timestamps. I used AI to figure out what went wrong there but it seems that for my Unifi camera's some of the streams don't include these timestamps. So OvenMediaEngine keeps waiting and then crashes.

Please have a look at that part yourself and decide on what to do with it.

I'll close this PR

PR #2005, opened 2026-02-27
feat: Per-track RTCP SR fallback to delta-based PTS calculation

devedse:
## Summary

Adds per-track fallback from RTCP SR-based to delta-based PTS calculation when a specific track never receives Sender Reports.

## What this does

- Starts a per-track timer when `CalcPTS` first fails (no SR received yet)
- After 5 seconds without an SR for a given track, falls back to delta-based PTS calculation for that track only
- Other tracks that do receive SRs continue using lip-sync based calculation
- Logs a warning when falling back
- Clears fallback state on stream reset

## Why it's needed

Some IP cameras (e.g. certain Unifi models) never send RTCP Sender Reports for one or more tracks. Without this fix, those tracks stall indefinitely waiting for an SR that never arrives, making the entire stream unusable.

### DIFF
--- a/src/projects/base/provider/stream.cpp
+++ b/src/projects/base/provider/stream.cpp
@@ -393,6 +393,8 @@ namespace pvd
 		_start_timestamp_us = -1LL;
 
 		_source_timestamp_map.clear();
+		_sr_fallback_tracks.clear();
+		_per_track_sr_wait.clear();
 	}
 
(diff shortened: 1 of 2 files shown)


getroot:
If the timestamp calculation method differs by track, there is a very high chance that AV sync will go out of alignment. So I solved this in the following way: https://github.com/OvenMediaLabs/OvenMediaEngine/pull/2031

PR #2006, opened 2026-02-27
fix: Add libgomp1 dependency to Dockerfile.local

devedse:
## Summary

Adds `libgomp1` to the package install list in `Dockerfile.local`.

## Why it's needed

The local Docker build fails at runtime when OpenMP (libgomp) is not installed. This is a transitive dependency that isn't pulled in automatically by the other packages in the base image.


Keukhan:
Thanks for the PR. I'm currently revising the Dockerfile overall

PR #2007, opened 2026-02-27
fix: Correct VP8 case label indentation in rtspc_stream.cpp

devedse:
## Summary

Fixes incorrect indentation of the `VP8` case label in the codec switch statement in `rtspc_stream.cpp`.

## What this does

The `case PayloadAttr::SupportCodec::VP8:` label had an extra tab, placing it at the same indentation level as the code inside the case block rather than aligned with the other case labels (`H264`, `H265`, `MPEG4_GENERIC`, etc.).

One-line whitespace fix, no functional change.


getroot:
Thank you for the contribution! Our policy is to only accept PRs that include functional changes. Cosmetic-only changes such as whitespace and formatting fixes are out of scope. Closing this PR per that policy.

PR #2008, opened 2026-03-01
fix: Data race in track

dimiden:


src/projects/base/info/audio_track.cpp
```diff
@@ -21,45 +22,44 @@ AudioTrack::AudioTrack()
 
 void AudioTrack::SetSampleRate(int32_t sample_rate)
 {
-	std::unique_lock<std::shared_mutex> lock(_amutex);
+	std::scoped_lock lock(_audio_mutex);
```
getroot: Is there any reason why it was changed to scoped_lock even when locking only one mutex?

getroot:
LGTM

PR #2021, opened 2026-03-16
chore(ci): skip docker build on non-code changes and add PR title validation

getroot:
## Summary

Two GitHub Actions workflow improvements:

### 1. Skip Docker build on non-code changes (`docker-image-dev-multi.yml`)

Previously, any push to `master` triggered the multi-arch Docker build — including commits that only changed docs, markdown files, or GitHub config.

Now the dev build only triggers when files under `src/**` or `misc/**` are changed.

> **Note:** When the build system is fully migrated to CMake, `cmake/**` and `CMakeLists.txt` should also be added to the `paths` filter.

### 2. PR title validation workflow (`ome-pr-check.yml`)

Added a workflow that enforces [Conventional Commits](https://www.conventionalcommits.org/) format for PR titles. On validation failure, a bot comment is posted explaining the required format and allowed types.

### DIFF
--- a/.github/workflows/ome-pr-check.yml
+++ b/.github/workflows/ome-pr-check.yml
@@ -1,0 +1,86 @@
+name: "OME PR Title Checker"
+
+on:
+  pull_request:
+    # Trigger on PR creation, title edits, or new commits
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  validate-title:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-image-dev-multi.yml
dimiden: It might be good to add `cmake/**` as well, since there could be cases where a rebuild should occur when CMake files are changed.

PR #2032, opened 2026-03-19
fix: improve RTP depacketizer robustness for H264 and H265

getroot:
- Improved H264/H265 depacketizers to properly cache decoding parameter sets (SPS/PPS/VPS) when they are delivered via FU-A/FU fragmentation, which is uncommon but valid per RFC 6184/7798
- Fixed minor bugs in H264 and H265 depacketizers
  - Removed the redundant `start_payload` flag in H264. First fragment detection now relies solely on the FU-A S-bit (RFC 6184)
  - Fixed an incorrect DONL size check in H265 that silently dropped valid FU packets when `sprop-max-don-diff=0`


Keukhan:
I’ve removed my previous review. After reviewing again, it seems that the fragmented decoding configuration will parse well. I have no objections.
LGTM

PR #2035, opened 2026-03-20
refactor: improve pull stream error handling in Orchestrator

dimiden:
- Change `RequestPullStream()` return type from `bool` to `std::shared_ptr<ocst::Error>` to provide detailed failure reasons.
- Add URL scheme validation to ensure all URLs in a pull request share the same scheme.
- Update API controllers and publishers to handle the new Error object and provide better error responses/logging.
- Fix a bug where `RequestPullStreamWithUrls()` would return success even on failure.


getroot:
Can you refactor it using the new `std::expected`? 
I think it would be a good idea to take this opportunity to introduce it starting from here.

dimiden:
Applying `std::expected` efficiently requires further consideration, so I believe it's best to handle it in a separate PR. I will submit a follow-up PR for this once the current task is completed.

getroot:
ok

PR #2056, opened 2026-03-30
refactor: remove unused FILE provider

Keukhan:
### Summary
The File Provider feature was developed to provide fallback stream when a stream was interrupted, but it is no longer used due to the Schedule Channel feature. Therefore, it is being removed.

### Changes
- Keep the `Providers.FILE `configuration in Server.xml and added warning logs
- Removed File Provider related source code
- Updated Makefile and CMakeLists files


src/projects/base/common_types.h
```diff
@@ -55,7 +55,6 @@ enum class StreamSourceType : int8_t
 	Mpegts,
 	Srt,
 	Transcoder,
-	File,
```
getroot, line 58: I recommend keeping the definitions as it is for future use.

Keukhan: Restore this value for future use

getroot (approved):
LGTM

PR #2058, opened 2026-03-31
refactor: add Instantiate() to probe encoder properties before full creation

getroot:
Currently, encoder properties (like IsInputOnly()) can only be read after a successful Create(), which makes it impossible to inspect them when creation fails or before committing to full initialization.

This PR introduces Instantiate(), a lightweight static method that constructs an encoder instance without calling Configure(). This allows callers to probe encoder properties regardless of whether full creation succeeds or fails.

As a first use case, IsInputOnly() is added as a virtual property for encoders that produce no output frames (e.g. WhisperEncoder).

Note: This is a prerequisite for the upcoming Whisper STT improvement PR. The follow-up PR depends on this being merged first.

### DIFF
--- a/src/projects/transcoder/transcoder_encoder.cpp
+++ b/src/projects/transcoder/transcoder_encoder.cpp
@@ -155,41 +155,57 @@ std::shared_ptr<std::vector<std::shared_ptr<info::CodecCandidate>>> TranscodeEnc
 	return candidate_modules;
 }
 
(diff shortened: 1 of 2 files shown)


dimiden (approved):
LGTM

getroot:
@Keukhan looks good to you?

Keukhan (approved):
The exception for the default module (for non-H.264 codecs) was removed.
but, since only one codec is created, I think it should still work fine.
I can add a condition later if module separation is needed.

Looks good to me. 
thanks.

PR #2067, opened 2026-04-13
fix: protect _transcode_apps with shared_mutex in Transcoder

getroot:
## Problem

`Transcoder::_transcode_apps` (`std::map`) was accessed from multiple threads without synchronization:

- **Write**: `OnCreateApplication`, `OnDeleteApplication` — called from the MediaRouter thread
- **Read**: `GetApplicationById` — can be called from API server threads or other contexts

This is a data race on `std::map`, which is undefined behavior and can cause crashes or corruption.

## Changes

- Add `mutable std::shared_mutex _transcode_apps_mutex` to `Transcoder`
- `unique_lock` for all write paths (`OnCreateApplication`, `OnDeleteApplication`)
- `shared_lock` for read paths (`GetApplicationById`)
- `OnDeleteApplication` now extracts the `shared_ptr` under the lock, then calls `UnregisterObserverApp`/`UnregisterConnectorApp` outside the lock to avoid lock inversion with MediaRouter
- `OnCreateApplication` rolls back the map insertion if MediaRouter registration fails

### DIFF
--- a/src/projects/transcoder/transcoder.cpp
+++ b/src/projects/transcoder/transcoder.cpp
@@ -76,13 +76,18 @@ bool Transcoder::OnCreateApplication(const info::Application &app_info)
 		return false;
 	}
 
-	_transcode_apps[application_id] = application;
+	{
+		std::unique_lock<std::shared_mutex> lock(_transcode_apps_mutex);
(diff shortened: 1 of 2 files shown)


dimiden (approved):
LGTM

PR #2068, opened 2026-04-16
fix: resolve data races in provider stream

dimiden:
### Summary
Fixes data races in `pvd::Stream` and related classes.

### Changes

**`pvd::PushStream` - silence timer race**
- Replace `ov::StopWatch _packet_silence_timer` with `std::atomic<int64_t> _last_received_time_ms`
- Remove non-atomic access in `UpdateLastReceivedTime()`, `GetElapsedMsSinceLastReceived()`, and `PublishChannel()`
- Use `steady_clock` instead of `system_clock` for elapsed time measurement to avoid NTP jump impact

**`pvd::Stream` - state race**
- Replace `State _state` with `std::atomic<State>` to eliminate race between `StreamMotor` and `WhiteElephantStreamCollector` threads

**`pvd::Stream` - last packet received time race**
- Replace `std::chrono::time_point _last_pkt_received_time` with `std::atomic<int64_t> _last_pkt_received_time_us`

**`pvd::Stream` - timestamp state race**
- Introduce `_source_stream_timestamp_mutex` to protect timestamp state
- Guard `AdjustTimestampByBase()`, `AdjustTimestampByDelta()`, `ResetSourceStreamTimestamp()`, `UpdateReconnectTimeToBasetime()`, and `GetBaseTimestamp()` under the new mutex

**`MpegTsProvider::Stop()` - unlocked map move**
- Acquire `_stream_port_map_lock` before moving `_stream_port_map` to prevent concurrent access during stop

### DIFF
--- a/src/projects/base/provider/stream.cpp
+++ b/src/projects/base/provider/stream.cpp
@@ -79,6 +79,8 @@ namespace pvd
 	// Consider the reconnection time and add it to the base timestamp
 	void Stream::UpdateReconnectTimeToBasetime()
 	{
+		std::scoped_lock lock(_source_stream_timestamp_mutex);
+
 		auto last_pkt_received_time_us = _last_pkt_received_time_us.load();
 		if (last_pkt_received_time_us >= 0)
 		{
@@ -375,6 +377,8 @@ namespace pvd
 
 	void Stream::ResetSourceStreamTimestamp()
 	{
+		std::scoped_lock lock(_source_stream_timestamp_mutex);
+
 		// Set the last timestamp of the highest value of all tracks
 		// In this algorithm, the timestamp of A or V jumps for synchronization.
 		// But after testing with a variety of players, this is better.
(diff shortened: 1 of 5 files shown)


src/projects/base/provider/stream.cpp
```diff
@@ -632,9 +643,12 @@ namespace pvd
 		}
 
 		int64_t base_timestamp = 0;
-		if (_base_timestamp_us != -1)
 		{
-			base_timestamp = _base_timestamp_us;
+			std::scoped_lock lock(_source_stream_timestamp_mutex);
```
Keukhan, line 647: It seems inefficient to use std::scoped_lock when only reading.

Keukhan (approved):
lgtm

src/projects/base/provider/stream.cpp
dimiden: The reason I didn't use `std::shared_lock` is that I wanted to ensure the execution order is strictly preserved. `std::mutex` provides the full synchronization needed to keep everything in the right sequence.

PR #2069, opened 2026-04-16
perf: optimize MPEG-2 TS packetizer performance

getroot:
Pre-compute the required packet count and write all TS packets for a PES directly into a single flat buffer, avoiding per-packet heap allocation and repeated memcpy rounds.

Measured CPU usage (single-core, SRT output): 4–6% → 2–3%

### DIFF
--- a/src/projects/modules/containers/mpegts/mpegts_packet.cpp
+++ b/src/projects/modules/containers/mpegts/mpegts_packet.cpp
@@ -68,7 +68,7 @@ namespace mpegts
 		return packet;
 	}
 
-	std::vector<std::shared_ptr<Packet>> Packet::Build(const std::shared_ptr<Pes> &pes, bool has_pcr, uint8_t continuity_counter)
+	std::vector<std::shared_ptr<Packet>> Packet::Build(const std::shared_ptr<Pes> &pes, bool has_pcr, bool is_keyframe, uint8_t continuity_counter)
 	{
 		std::vector<std::shared_ptr<Packet>> packets;
 
@@ -162,7 +162,7 @@ namespace mpegts
 				size_t stuffing_bytes = payload_buffer_size - payload_size;
 				packet->_adaptation_field._length = 1 + (has_pcr ? 6 : 0) + stuffing_bytes; // flags(8bits) + PCR(6) + stuffing_bytes
 				packet->_adaptation_field._pcr_flag = has_pcr;
-				packet->_adaptation_field._random_access_indicator = true;
+				packet->_adaptation_field._random_access_indicator = first_packet && is_keyframe;
 
 				packet->_adaptation_field_size = 1 + packet->_adaptation_field._length; // length(8bits) + flags(8bits) + PCR(6) + stuffing_bytes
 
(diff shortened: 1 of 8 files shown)


src/projects/modules/containers/mpegts/mpegts_packet.cpp
```diff
@@ -162,7 +162,7 @@ namespace mpegts
 				size_t stuffing_bytes = payload_buffer_size - payload_size;
 				packet->_adaptation_field._length = 1 + (has_pcr ? 6 : 0) + stuffing_bytes; // flags(8bits) + PCR(6) + stuffing_bytes
 				packet->_adaptation_field._pcr_flag = has_pcr;
-				packet->_adaptation_field._random_access_indicator = true;
+				packet->_adaptation_field._random_access_indicator = first_packet && is_keyframe;
```
dimiden, line 165: @getroot 
Is this related to performance improvement, or is it a different update?

getroot: Congratulations on finding the hidden picture.

While developing a new function, I noticed a non-standard implementation, so I rewrote it. Neither the existing code nor the new code affects playback, and since the change is very minor, I simply included it in this PR.

src/projects/publishers/srt/srt_playlist.cpp
```diff
@@ -86,20 +86,29 @@ namespace pub
...
+		while (offset < total)
 		{
-			auto size		 = _data_to_send->GetLength();
-			const auto &data = packet->GetData();
+			const size_t current  = _data_to_send->GetLength();
+			const size_t avail    = total - offset;
 
-			// Broadcast if the data size exceeds the SRT's payload length
-			if ((size + data->GetLength()) > SRT_LIVE_DEF_PLSIZE)
+			if (current + avail > SRT_LIVE_DEF_PLSIZE)
```
dimiden: @getroot 
When `current + avail == SRT_LIVE_DEF_PLSIZE`, the buffer is exactly full but won't be flushed until the next `SendData()` call, adding up to one frame of unnecessary latency. Using `>=` would flush immediately when the buffer hits exactly 1316 bytes.
What are your thoughts on this?

getroot: You are right. I've applied this. Thanks

dimiden (approved):
LGTM. Ready to merge as soon as the Copilot issues are addressed. 👍

src/projects/modules/containers/mpegts/mpegts_packetizer.h
```diff
@@ -24,8 +24,8 @@ namespace mpegts
         virtual ~PacketizerSink() = default;
         // PAT, PMT, ...
         virtual void OnPsi(const std::vector<std::shared_ptr<const MediaTrack>> &tracks, const std::vector<std::shared_ptr<mpegts::Packet>> &psi_packets) = 0;
-        // PES packets for a frame
-        virtual void OnFrame(const std::shared_ptr<const MediaPacket> &media_packet, const std::vector<std::shared_ptr<mpegts::Packet>> &pes_packets) = 0;
+        // All TS packets for one frame, pre-serialised into a single flat buffer
+        virtual void OnFrame(const std::shared_ptr<const MediaPacket> &media_packet, const std::shared_ptr<ov::Data> &ts_data) = 0;
```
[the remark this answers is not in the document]
getroot: Applied it

PR #2071, opened 2026-04-17
feat: add multi-audio track metadata support to Multiplex provider

nums:
feat: add multi-audio track metadata support to Multiplex provider

## Summary

Extends the Multiplex provider's `TrackMap` to support per-track audio metadata (`PublicName`, `Language`, `Characteristics`) and an `AudioIndex` field to differentiate multiple audio tracks sharing the same `SourceTrackName`. This brings multi-language audio support to the Multiplex provider, on par with what the Scheduled provider already offers.

> Implemented with the help of Claude (Anthropic AI assistant).

## What this does

- Adds `PublicName`, `Language`, `Characteristics` and `AudioIndex` fields  to `NewTrackInfo` in `MultiplexProfile`
- When multiple audio tracks in a source stream share the same variant name, `AudioIndex` (0-based) selects which occurrence receives which metadata
- Without `AudioIndex`, a single `<Track>` entry acts as a catch-all and preserves the original track metadata from the source stream
- Both XML (`.mux` files) and JSON (REST API) formats are supported
- Metadata is applied at track creation time and flows through to LL-HLS
  `EXT-X-MEDIA` attributes (`NAME`, `LANGUAGE`, `CHARACTERISTICS`)
- Updates documentation with a multi-audio example

## Why it's needed

The Multiplex provider is commonly used to combine streams for ABR delivery. Sources such as SRT streams or broadcast feeds often carry multiple embedded audio tracks. Without this feature, only one audio track per source variant could be mapped (additional tracks sharing the same `SourceTrackName` were silently dropped) making it impossible to expose multiple language tracks simultaneously in the LL-HLS master playlist.

## Files changed

- `providers/multiplex/multiplex_profile.h` - added fields to `NewTrackInfo`, changed `_new_track_infos_map` to support multiple entries per `SourceTrackName` (vector), updated `GetNewTrackInfo` signature to accept an occurrence index
- `providers/multiplex/multiplex_profile.cpp` - XML/JSON parsing and serialization for the new fields
- `providers/multiplex/multiplex_stream.cpp` - occurrence tracking per variant name, apply metadata overrides when cloning tracks
- `docs/live-source/multiplex-channel.md` - added multi-audio example with `AudioIndex`, `PublicName` and `Language`


getroot:
@nums 
Thank you for submitting the PR.

Recently, our internal developers have also been committing changes through PRs, and receiving a Copilot review is now required. Could you please review Copilot’s comments first?

src/projects/providers/multiplex/multiplex_profile.h
```diff
@@ -180,36 +191,42 @@ namespace pvd
...
-            {
-                auto it = _new_track_infos_map.find(source_track_name);
-                if (it == _new_track_infos_map.end())
+                // Single entry without explicit index → applies to all occurrences
+                if (infos.size() == 1 && infos[0].audio_index < 0)
                 {
-                    return false;
+                    new_track_info = infos[0];
+                    return true;
                 }
```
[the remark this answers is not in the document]
nums: With the new commit 9c8285f1957ab08c24a7056c0a8bf2b171689640. The XML parser was already missing the < -1 validation that the JSON parser had, so a value like -2 could silently pass through. Fixed in both parsers: XML now rejects any value < -1 with an explicit error message, and GetNewTrackInfo now uses == -1 instead of < 0 to make the catch-all contract exact and unambiguous.

src/projects/providers/multiplex/multiplex_profile.h
```diff
@@ -180,36 +191,42 @@ namespace pvd
...
+                {
+                    if (info.audio_index == occurrence)
+                    {
+                        new_track_info = info;
+                        return true;
+                    }
+                }
 
-                return true;
+                return false;
```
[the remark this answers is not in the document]
nums: With the new commit 9c8285f1957ab08c24a7056c0a8bf2b171689640. The old infos.size() == 1 guard was too strict - if two `<Track>` entries share the same SourceTrackName without `<AudioIndex>`, both get audio_index == -1, the catch-all branch never fires, and all occurrences are silently dropped.                                                                                                                            
                                                         
Fixed with two changes:                                                                                                                                                                  
1. GetNewTrackInfo now uses two passes: first an explicit index match (audio_index == occurrence), then a fallback scan for any entry with audio_index == -1. This means a catch-all  always works as a fallback regardless of how many entries exist.                                                                                                                         
2. At parse time (both XML and JSON), a logtw warning is emitted when a second catch-all entry is detected for the same SourceTrackName, so misconfiguration is visible in the logs rather than silently tolerated.

src/projects/providers/multiplex/multiplex_profile.cpp
```diff
@@ -600,9 +658,61 @@ namespace pvd
...
+				{
+					characteristics = characteristics_node.text().as_string();
+				}
+
+				int32_t audio_index = -1;
+				auto audio_index_node = track_node.child("AudioIndex");
+				if (audio_index_node)
+				{
+					audio_index = audio_index_node.text().as_int();
+					if (audio_index < -1)
```
getroot, line 700: `as_int()` returns 0 when the value can't be parsed as an integer (e.g. empty tag or non-numeric text), so invalid input is silently interpreted as `AudioIndex 0` instead of being rejected. The `audio_index < -1` check won't catch this. Consider validating the raw text, similar to the isInt() check on the JSON path.

src/projects/providers/multiplex/multiplex_profile.cpp
```diff
@@ -217,9 +217,67 @@ namespace pvd
...
+						return false;
+					}
+				}
+
 				ov::String source_track_name = source_track_name_object.asString().c_str();
 				ov::String new_track_name = new_track_name_object.asString().c_str();
-				source_stream->AddTrackMap(source_track_name, NewTrackInfo(source_track_name, new_track_name, bitrate_conf, framerate_conf));
+
+				// Warn if adding a second catch-all for the same source track name
+				if (audio_index == -1)
```
getroot: This warning only fires for duplicate catch-all entries (`audio_index == -1`). The same situation with two entries sharing an explicit `AudioIndex` (e.g. both `AudioIndex=1`) is also silently dropped at lookup time but produces no warning. It would be helpful to emit a similar warning for that case too, so the user has a hint that part of their config is being ignored.

getroot (changes requested):

getroot:
@nums I reviewed your PR three weeks ago. Have you had a chance to see it?

nums:
> @nums I reviewed your PR three weeks ago. Have you had a chance to see it?

Hi @getroot , sorry, we’ve been quite busy. I’ll try to take your feedback into account by the end of this week.

src/projects/providers/multiplex/multiplex_profile.cpp
nums: https://github.com/nums/OvenMediaEngine/commit/d85bfbba3d27085ac939ee38b953429c0a8a1255
Silent fallback on invalid AudioIndex text (XML parser).
Fixed. Instead of calling as_int() directly (which returns 0 for empty or non-numeric text), the XML parser now reads the raw string and validates it with strtol, checking that the entire input was consumed. An invalid value now produces an explicit error message rather than silently mapping to index 0.

src/projects/providers/multiplex/multiplex_profile.cpp
nums: https://github.com/nums/OvenMediaEngine/commit/d85bfbba3d27085ac939ee38b953429c0a8a1255
No warning for duplicate explicit AudioIndex:
Fixed in both the XML and JSON parsers. The duplicate-detection block previously only fired when audio_index == -1 (catch-all). It now also emits a logtw warning when two entries share the same explicit AudioIndex value for the same SourceTrackName, so the user is informed that the later entry will be silently ignored at lookup time.

getroot (approved):
Thank you

PR #2075, opened 2026-04-20
build: fix cmake prerequisite installation failures on Ubuntu 22 with cmake 4.x

getroot:
## Problem

Several issues caused `cmake -B` to fail on a fresh Ubuntu 22 installation when cmake 4.x (from the Kitware PPA) is used:

1. **OS detection regex failure** — The `NAME` field in `/etc/os-release` is not on the first line (it is preceded by `PRETTY_NAME`). The anchored regex `^NAME=` failed to match, leaving `OSNAME` empty. As a result, all OS-specific package installation steps were silently skipped, including the base packages needed to build dependencies.

2. **Missing `automake` in Ubuntu base packages** — NASM's `autogen.sh` requires `automake`, which was not included in the Ubuntu apt package list.

3. **cmake 4.x compatibility** — cmake 4.x removed backward compatibility for projects using `cmake_minimum_required` versions older than 3.5. SRT 1.5.2, spdlog, whisper, and the internal stubs target are affected. Additionally, SRT's `./configure` wrapper (a Tcl script) does not support passthrough cmake arguments, so it is replaced with a direct cmake invocation.

4. **Missing nasm for targeted reinstalls** — When a single dependency (e.g. `libvpx`) is reinstalled via `REINSTALL_TARGET`, nasm was not installed first, causing the libvpx configure step to fail with "Neither yasm nor nasm have been found".

## Changes

- Fix `OSNAME` detection regex: `^NAME=` → `(^|\n)NAME=` and use `CMAKE_MATCH_2`
- Add `automake` to Ubuntu base packages
- Replace SRT `./configure` wrapper with direct `cmake` invocation and add `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`
- Add `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` to spdlog, whisper, and stubs cmake calls
- Ensure `nasm` is installed before `libvpx`, `fdk_aac`, and `libx264` in targeted reinstall mode

## Test plan

- [ ] Fresh Ubuntu 22.04 Docker container with cmake 4.x (Kitware PPA)
- [ ] `cmake -B build/Debug -G Ninja -DOME_DEP_PREFIX=/opt/ovenmediaengine_oss_debug` completes without errors

### DIFF
--- a/cmake/InstallPrerequisites.cmake
+++ b/cmake/InstallPrerequisites.cmake
@@ -200,8 +200,8 @@ set(WHISPER_SOURCE_URL "https://github.com/ggml-org/whisper.cpp/archive/${WHISPE
 if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
     if(EXISTS /etc/os-release)
         file(READ /etc/os-release _os_release)
-        string(REGEX MATCH "^NAME=\"?([^\"\n]+)\"?" _m "${_os_release}")
-        set(OSNAME "${CMAKE_MATCH_1}")
+        string(REGEX MATCH "(^|\n)NAME=\"?([^\"\n]+)\"?" _m "${_os_release}")
+        set(OSNAME "${CMAKE_MATCH_2}")
         string(REGEX MATCH "VERSION=\"?([0-9]+)" _m "${_os_release}")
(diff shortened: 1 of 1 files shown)


dimiden (approved):
LGTM

PR #2077, opened 2026-04-20
fix: Restore ProviderType::File for compatibility

Keukhan:
**SUMMARY**
Restored the deleted ProviderType::File for backward compatibility.

* Added `File` to the `ProviderType` enum in `common_types.h`


dimiden (approved):
LGTM

PR #2078, opened 2026-04-20
fix: keep OutputTrack quality even when input stream is updated

Keukhan:
### Summary

Improved OutputTrack quality retention behavior when the source stream is updated, along with minor refactoring.

#### 1. Default Parameter Handling
- OutputTrack now uses parameters defined in the Config as default values during create output track. so, removed previous logic that selected default values based on whether Config parsing was performed.

#### 2. Output Track Initialization
- If no user-defined settings are provided, parameters are automatically derived from decoded frame information:
  - Resolution, Timebase, Framerate, Bitrate, Samplerate, Channel

#### 3. Behavior on Input Stream Update
- **Encoding mode**
  - OutputTrack keeps its initial quality even if the input stream quality change.

- **Bypass mode**
  - OutputTrack updates to match the changed input stream quality

### DIFF
--- a/src/projects/transcoder/transcoder_stream.cpp
+++ b/src/projects/transcoder/transcoder_stream.cpp
@@ -1583,7 +1583,7 @@ void TranscoderStream::UpdateOutputTrack(std::shared_ptr<MediaFrame> buffer)
 
 		if (output_track->IsBypass())
 		{
-			UpdateOutputTrackPassthrough(output_track, buffer);
+			UpdateOutputTrackPassthrough(output_track, input_track, buffer);
 		}
 		else
 		{
--- a/src/projects/transcoder/transcoder_stream.cpp
+++ b/src/projects/transcoder/transcoder_stream.cpp
@@ -369,6 +369,9 @@ bool TranscoderStream::UpdateInternal(const std::shared_ptr<info::Stream> &strea
 
 		// Restrict transcoding while all decoders/filters/encoders are being generated
 		std::unique_lock<std::shared_mutex> pipeline_lock(_pipeline_mutex);
+
+		UpdatePassthroughOutputTracks(stream);
+
 		RemoveDecoders();
 		RemoveFilters();
 		RemoveSpecificEncoders();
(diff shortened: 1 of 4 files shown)


src/projects/transcoder/transcoder_stream.cpp
```diff
@@ -1583,7 +1583,7 @@ void TranscoderStream::UpdateOutputTrack(std::shared_ptr<MediaFrame> buffer)
 
 		if (output_track->IsBypass())
 		{
-			UpdateOutputTrackPassthrough(output_track, buffer);
+			UpdateOutputTrackPassthrough(output_track, input_track, buffer);
```
dimiden: `UpdateOutputTrack()` iterates `_composite.GetInputOutputListByDecoderId(...)`, but bypass outputs are stored in `_input_to_outputs` and are not part of the decoder/encoder path. That means the new `UpdateOutputTrackPassthrough(...)` branch is never reached for actual bypass tracks, so this change does not seem to make bypass outputs follow input-track updates as intended.

src/projects/transcoder/transcoder_stream_internal.cpp
```diff
@@ -437,121 +311,41 @@ std::shared_ptr<MediaTrack> TranscoderStreamInternal::CreateOutputTrack(const st
...
-	output_track->SetTimeBase(GetDefaultTimebaseByCodecId(output_track->GetCodecId()));
+
+	cmn::Resolution resolution;
+	resolution.width  = profile.GetWidth();
+	resolution.height = profile.GetHeight();
+	output_track->SetResolutionByConfig(resolution);
+	output_track->SetResolution(resolution);
+	output_track->SetFrameRateByConfig(profile.GetFramerate());
+
+	ApplySkipFrames(profile.GetSkipFrames(), input_track, output_track);
```
dimiden: `ApplySkipFrames()` is now called unconditionally for image profiles. Since `ImageProfile::_skip_frames` defaults to `0`, an omitted `<SkipFrames>` is treated the same as an explicit `0`, so the output framerate may be overwritten even when the user did not configure `SkipFrames`. Previously this logic only ran when the field was explicitly parsed.
Is this intended?

Keukhan: This is intended behavior. When SkipFrames is set to 0, the value is dynamically calculated and adjusted automatically. It is disabled only when set to -1. 

However, in the ApplySkipFrames function, when the skip_frames value is less than 0, the SetSkipFramesByConfig function of the Track is not being called, so this needs to be fixed.

Thanks.

src/projects/transcoder/transcoder_stream.cpp
Keukhan: You're right, that function won’t work as expected in the current code. Good catch. This part should be removed in the future.

Keukhan:
@dimiden Please review again. :)

dimiden:
LGTM

dimiden (approved):

PR #2082, opened 2026-04-22
fix: fixed incorrect libdir in hiredis.pc caused by the LIBRARY_PATH env

Keukhan:
### Summary

Fix an issue where the libdir path in hiredis.pc is incorrectly set due to the LIBRARY_PATH environment variable when building in CUDA-based Ubuntu Docker images.

### Root Cause

CUDA-based Docker images define the following environment variable:
```
LIBRARY_PATH=/usr/local/cuda/lib64/stubs
```
The hiredis Makefile uses LIBRARY_PATH when generating hiredis.pc:

```
PREFIX?=/usr/local
LIBRARY_PATH?=lib

$(PKGCONFNAME):
	@echo prefix=$(PREFIX) > $@
	@echo libdir=$(PREFIX)/$(LIBRARY_PATH) >> $@
```
If LIBRARY_PATH is set, it overrides the default value (lib), resulting in an incorrect libdir path in hiredis.pc.
ex) `ex) libdir=/opt/ovenmediaengine//usr/local/cuda/lib64/stubs`
### Solution

Explicitly override LIBRARY_PATH to lib during the build and installation steps in install_hiredis within InstallPrerequisites.cmake:
```
make ${_J} PREFIX=${PREFIX} LIBRARY_PATH=lib &&  sudo make install PREFIX=${PREFIX} LIBRARY_PATH=lib
```
This ensures that the correct libdir is set in hiredis.pc, regardless of the environment.

### DIFF
--- a/cmake/InstallPrerequisites.cmake
+++ b/cmake/InstallPrerequisites.cmake
@@ -367,7 +367,7 @@ make ${_J} && sudo make install && rm -rf ${TEMP_PATH}/x264
 set(_install_nvcc_hdr "
 mkdir -p ${TEMP_PATH}/nvcc-hdr && cd ${TEMP_PATH}/nvcc-hdr &&
 curl -sSLf ${NVCC_HDR_SOURCE_URL} | tar -xz --strip-components=1 &&
-make PREFIX="${PREFIX}" LIBDIR=lib install
+sudo make PREFIX="${PREFIX}" LIBDIR=lib install
 ")
 
 # ---- FFmpeg ----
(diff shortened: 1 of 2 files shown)


getroot:
Please check why previous `prerequisites.sh` works well.

Keukhan:
> Please check why previous `prerequisites.sh` works well.

The hiredis.pc file and library are correctly installed under the PREFIX path. 
Although the libdir in hiredis.pc was incorrectly set to /usr/local/lib, the build still completed successfully. 
It seems that the Makefile system referenced the libhiredis.so file located in /opt/ovenmediaengine/lib.

dimiden:
I'm on the same page as Copilot here. It would be great if we could standardize the `sudo` usage to keep it consistent. Everything else looks good.

dimiden (approved):
LGTM

Issue #115, opened 2020-05-16
Unable to install OME version 1.10.2


swedishdeveloper:
Unable to install the latest version 1.10.2 of OME, because it fails in middle of the command "**make release**".

I get this error code: 
**projects/providers/rtspc/rtspc_stream.cpp: In member function ‘virtual int pvd::RtspcStream::GetFileDescriptorForDetectingEvent()’:
projects/providers/rtspc/rtspc_stream.cpp:334:10: error: ‘av_get_iformat_file_descriptor’ was not declared in this scope
   return av_get_iformat_file_descriptor(_format_context);
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated due to -Wfatal-errors.
make: *** [core/build_objects.mk:21: intermediates/RELEASE/objs/providers/rtspc/rtspc_stream.o] Error 1**

**To Reproduce**
1. Follow the installation guide at https://airensoft.gitbook.io/ovenmediaengine/getting-started
2. When you come to down to type "make release" you will get the error when the script is near complete.

**Expected behavior**
It should not fail in the middle of the installation.

**Server (please complete the following information):**
 - OS: [Centos 7]
 - OvenMediaEngine Version: [v0.10.2]
 - Branch: [Master]

Do you not provide any better and simplier way to update OME from e.g 1.10.1 to 1.10.2 or to you need to delete all "OME" files before you can install the new version, do you not have any "UPDATE" command?

Another TIP:
You are using old version of e.g. SRT in the prerequisites file. The libraries have release much newer versions and I think It can make OME more stable?

swedishdeveloper:
I fixed the problem of deleting all "OME related files" and then I ran the installation script again. Then it worked fine.

getroot:
Yes, Some external libraries have been changed and added, so you have to run prerequisites.sh again.

refer this, https://github.com/AirenSoft/OvenMediaEngine/releases

PR #2083, opened 2026-04-24
feat: add Transport field to SocketAddress

getroot:
## Summary

Add a `Transport` field (UDP/TCP/SRT) to `SocketAddress` so that transport
information is naturally included in log output everywhere an address is printed.

Before:

    <ClientSocket: 0x7ffee0000c10, #115, Connected, TCP, Nonblocking, 192.168.0.200:13914>

After:

    <ClientSocket: 0x7ffee0000c10, #115, Connected, TCP, Nonblocking, 192.168.0.200:13914/tcp>

## Motivation

OvenMediaEngine is planning to support TCP ICE Candidates. When both UDP and TCP
candidates coexist, logs that omit transport type cause confusion because it becomes
impossible to tell which transport a given address belongs to.

By embedding transport at the `SocketAddress` level, all log output gains this
context without any per-call changes.

### DIFF
--- a/src/projects/base/ovsocket/client_socket.cpp
+++ b/src/projects/base/ovsocket/client_socket.cpp
@@ -105,6 +105,16 @@ namespace ov
 		}
 
 		_local_address = std::make_shared<SocketAddress>("", local_addr);
+		_local_address->SetTransport(GetType() == SocketType::Srt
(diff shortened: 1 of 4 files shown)


Keukhan (approved):
It's simple, so there is not much to review. 
Looks good.

PR #2084, opened 2026-04-24
fix: fix nvcc_hdr library compilation error with CMake

Keukhan:
Removed double quotes around the `PREFIX` variable in the `make install` command for installing NVCC headers.


getroot (approved):
LGTM

PR #2089, opened 2026-04-27
refactor: add expired flag to ICE state change callbacks

dimiden:
## Summary

Propagate ICE session expiration metadata through ICE state change callbacks.

This change extends `IcePortObserver::OnStateChanged(...)` with an `is_expired`
flag and passes `IceSession::IsExpired()` from `IcePort`. WebRTC observer
implementations are updated to match the new callback signature.

## Changes

- add `bool is_expired` to `IcePortObserver::OnStateChanged(...)`
- pass `session->IsExpired()` from `IcePort::NotifyIceSessionStateChanged(...)`
- update WebRTC provider callback signature
- update WebRTC publisher callback signature

## Why

`IceConnectionState` represents transport/session state, but it does not carry
the reason why a session reached a disconnected/closed state.

Session lifetime expiration is additional metadata, not a new transport state.
Passing it separately keeps the existing state model intact while allowing
callers to distinguish expiration from ordinary disconnect cases.

### DIFF
--- a/src/projects/modules/ice/ice_port.cpp
+++ b/src/projects/modules/ice/ice_port.cpp
@@ -1357,7 +1357,7 @@ void IcePort::NotifyIceSessionStateChanged(std::shared_ptr<IceSession> &session)
 {
 	if (session->GetObserver() != nullptr)
 	{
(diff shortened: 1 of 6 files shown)


Keukhan (approved):
It looks like a simple interface change. looks good to me.

PR #2095, opened 2026-04-30
fix: prevent warning logs for queue exceed caused by buffering delay setting

Keukhan:
### Summary
When `DelayBuffer `is configured, the MediaRouter outbound stream queue intentionally retains packets for the specified duration. The existing count-based threshold had no awareness of this, causing false-positive Warning logs even though the packet accumulation was expected behavior.

Fix: Extended ManagedQueue to support a` time-based threshold` mode (`SetThresholdByTime`). The effective count is computed dynamically from the input rate (imps × time_ms), and the buffering delay portion is included so that delay-held packets never trigger threshold warnings.

When `DelayBuffer` is set, `SetBufferRetentionDuration(delay_ms + 1000)` is called on the outbound stream queue 
(extra 1 second absorbs transient spikes.)

### Changes
- info::ManagedQueue — Added ThresholdMode enum, SetThresholdByTime(), and _threshold_value (user-configured raw value). _threshold always holds the computed effective count limit.
- ov::ManagedQueue — Added SetThresholdByTime(), UpdateThreshold() (recalculates _threshold, IsThresholdExceeded(), 

- MediaRouteStream — Added SetBufferRetentionDuration() which sets a time-based threshold on the packets queue.
- MediaRouteApplication — Calls SetBufferRetentionDuration() on outbound streams when DelayBufferTime is configured.

- queue_metrics.h — _threshold moved to UpdateMetrics() to always reflect the latest computed value.

### Chore
- Removed deprecated skip message related code

### DIFF
--- a/src/projects/modules/managed_queue/managed_queue.h
+++ b/src/projects/modules/managed_queue/managed_queue.h
@@ -360,14 +360,13 @@ namespace ov
 		{
 			auto shared_lock = std::shared_lock(_name_mutex);
 
-			const char* urn_string = (_urn != nullptr) ? _urn->ToString().CStr() : "NoUrn";
-
+			ov::String urn_str = (_urn != nullptr) ? _urn->ToString() : ov::String("NoUrn");
 			return ov::String::FormatString(
 				"ManagedQueue [Id: %u, Size: %zu, Threshold: %zu (%s %zu%s + delay %dms), Peak: %zu, Imps: %zu, Omps: %zu, Wait: %s, Urn: %s]",
 				GetId(), _size, _threshold,
 				GetThresholdModeString(), _threshold_value, (_threshold_mode == ThresholdMode::TimeBased) ? " ms" : "", _buffering_delay,
 				_peak, _input_message_per_second, _output_message_per_second, _exceed_threshold_and_wait_enabled ? "On" : "Off",
-				urn_string);
+				urn_str.CStr());
 		}
 
 	private:
--- a/src/projects/modules/managed_queue/managed_queue.h
+++ b/src/projects/modules/managed_queue/managed_queue.h
@@ -99,7 +99,16 @@ namespace ov
 		{
 			info::ManagedQueue::SetThreshold(threshold);
 
-			MonitorInstance->OnQueueUpdated(*this, true);
+			MonitorInstance->OnQueueUpdated(*this);
+		}
+
+		// Set threshold in time-base mode.
+		// The effective item count is estimated as: input_message_per_second * time_ms / 1000.
+		void SetThresholdByTime(int time_ms)
+		{
+			info::ManagedQueue::SetThresholdByTime(time_ms);
+
+			MonitorInstance->OnQueueUpdated(*this);
 		}				
 
 		// Urgent item will be inserted at the front of the queue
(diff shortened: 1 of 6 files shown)


src/projects/modules/managed_queue/managed_queue.h
```diff
@@ -608,14 +511,16 @@ namespace ov
...
 
 						_last_logged_peak = _peak;
 					}
 				}
 				else
 				{
 					_threshold_exceeded_time_in_us = 0;
+
+					logt(LOG_TAG, "Stable. %s", GetInfoString().CStr());
 				}
```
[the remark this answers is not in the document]
Keukhan: Good suggestion

src/projects/modules/managed_queue/managed_queue.h
```diff
@@ -360,6 +355,18 @@ namespace ov
...
+		ov::String GetInfoString()
+		{
+			auto shared_lock = std::shared_lock(_name_mutex);
+
+			return ov::String::FormatString(
+				"ManagedQueue [Id: %u, Size: %zu, Threshold: %zu (%s %ld%s + delay %dms), Peak: %zu, Imps: %zu, Omps: %zu, Wait: %s, Urn: %s]",
+				GetId(), _size, _threshold,
+				GetThresholdModeString(), _threshold_value, (_threshold_mode == ThresholdMode::TimeBased) ? " ms" : "", _buffering_delay,
+				_peak, _input_message_per_second, _output_message_per_second, _exceed_threshold_and_wait_enabled ? "On" : "Off",
+				_urn->ToString().CStr());
```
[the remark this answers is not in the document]
Keukhan: Good suggestion

src/projects/modules/managed_queue/managed_queue.h
```diff
@@ -360,6 +356,20 @@ namespace ov
 			_buffering_delay = delay_ms;
 		}
 
+		ov::String GetInfoString()
+		{
+			auto shared_lock = std::shared_lock(_name_mutex);
+
+			const char* urn_string = (_urn != nullptr) ? _urn->ToString().CStr() : "NoUrn";
```
dimiden: This line could lead to a dangling pointer, potentially causing a crash. We should address this.

dimiden (changes requested):

src/projects/modules/managed_queue/managed_queue.h
Keukhan: I modified the code. please check again.

src/projects/modules/managed_queue/managed_queue.h
```diff
@@ -360,14 +360,13 @@ namespace ov
 		{
 			auto shared_lock = std::shared_lock(_name_mutex);
 
-			const char* urn_string = (_urn != nullptr) ? _urn->ToString().CStr() : "NoUrn";
-
+			ov::String urn_str = (_urn != nullptr) ? _urn->ToString() : ov::String("NoUrn");
```
dimiden, line 363: @Keukhan 
Is there a specific reason for creating a temporary object here instead of printing it directly?

Keukhan: There's simple reason. i just split it out because the parameters for FormatString() were getting to long :)

dimiden (approved):
LGTM

PR #2096, opened 2026-04-30
chore: Install NVIDIA driver using DKMS on RHEL 8/9

Keukhan:
## Improved driver selection logic

- Updated the awk pattern in find_nearest_higher_nvidia_driver_rocky() to select only NVIDIA driver modules that end with` -dkms` (DKMS-based drivers only).


choigilhoon (approved):
lgtm!

PR #2097, opened 2026-04-30
feat: rework WebRTC JitterBuffer with adaptive PTS-anchored pacing

getroot:
## Problem

When frames arrive at OME in bursts due to encoder delays, network
conditions, or publisher-side pacing, they are forwarded immediately
at that burst rate, causing uneven frame delivery that results in
degraded playback quality on WebRTC players.

## Solution

Replace the previous JitterBuffer with a per-track frame pacer that
dispatches each frame at its PTS-derived expected time. Bursty source
delivery is reshaped into evenly spaced output before packetization,
and the smoothing delay is adjusted automatically based on the actual
source jitter, with no manual tuning required.

The user-facing configuration is unchanged:

```xml
<WebRTC>
    <JitterBuffer>true</JitterBuffer>
</WebRTC>
```


src/projects/publishers/webrtc/rtc_stream.cpp
```diff
@@ -194,9 +210,37 @@ bool RtcStream::Start()
 			AddRtpHistory(track);
 		}
 
-		if (_jitter_buffer_enabled == true)
+		if (_pacer_enabled && _pacer_scheduler)
 		{
-			_jitter_buffer_delay.CreateJitterBuffer(track->GetId(), track->GetTimeBase().GetDen());
+			auto pacer = std::make_shared<FramePacer>(
```
Keukhan, line 215: `pacer`:  add exception handling for memory allocation failure

src/projects/publishers/webrtc/rtc_stream.cpp
```diff
@@ -194,9 +210,37 @@ bool RtcStream::Start()
...
+							auto rtc_stream = std::dynamic_pointer_cast<RtcStream>(stream);
+							if (!rtc_stream)
+								return;
+							if (is_video)
+								rtc_stream->PacketizeVideoFrame(pkt);
+							else
+								rtc_stream->PacketizeAudioFrame(pkt);
+						});
+
+			if (_adaptive_delay_controller)
```
Keukhan: If _adaptive_delay_controller is not allocated or initialized, it may not be assigned to the Pacer, which could cause the Pacer to not function properly. Exception handling is required.

How about using: 213: `if (_pacer_enabled && _pacer_scheduler && _adaptive_delay_controller)`

If the intended behavior is use `kPacerMinMs(fallback_ms)` when _adaptive_delay_controller is not set, this comment can be ignored. 👍

Keukhan (commented):
If Recompute runs every kRecomputeInterval (500ms), there may be a brief period of slow video playback when the FPS changes dynamically—for example, from 5 FPS to 60 FPS—lasting up to 500ms. This isn’t a major issue, but to improve it, how about triggering Recompute when the number of RecordSample() calls suddenly increases, so that _current_delay_ms can be updated more quickly?

src/projects/modules/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ add_subdirectory(containers)
 add_subdirectory(origin_map_client)
 add_subdirectory(cpix_client)
 add_subdirectory(transcode_webhook)
-add_subdirectory(jitter_buffer)
+add_subdirectory(pacer)
 add_subdirectory(rtsp)
```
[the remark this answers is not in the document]
getroot: Updated pacer/AMS.mk and main/AMS.mk to use the pacer target.

src/projects/publishers/webrtc/rtc_stream.cpp
```diff
@@ -128,13 +134,23 @@ bool RtcStream::Start()
...
+	_pacer_enabled = webrtc_config.IsJitterBufferEnabled();
+
+	if (_pacer_enabled)
+	{
+		_pacer_scheduler = std::make_shared<ov::DelayQueue>("FramePacer");
+		_pacer_scheduler->Start();
+
+		_adaptive_delay_controller = std::make_shared<AdaptiveDelayController>(
+			kPacerMinMs, kPacerMaxMs);
+	}
```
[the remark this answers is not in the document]
getroot: Acknowledged. Keeping per-stream for simplicity in this PR; will
revisit with a shared scheduler pool in a follow-up if it shows up
as a bottleneck at scale.

src/projects/modules/pacer/frame_pacer.cpp
```diff
@@ -0,0 +1,145 @@
...
+	auto packet_copy	 = packet;
+	_scheduler->Push(
+		[dispatcher_copy, packet_copy](void *) -> ov::DelayQueueAction {
+			if (dispatcher_copy)
+			{
+				dispatcher_copy(packet_copy);
+			}
+			return ov::DelayQueueAction::Stop;
+		},
+		after_ms);
```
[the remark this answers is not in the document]
getroot: Good catch. The fix touches ov::DelayQueue itself, so we'll address
it in a follow-up that gives the queue a steady_clock-based path
for pacing-sensitive scheduling.

src/projects/publishers/webrtc/rtc_stream.cpp
getroot: std::make_shared throws std::bad_alloc on failure rather than returning null, so a null check wouldn't trigger. We rely on the standard propagation, consistent with the rest of OME and at the point this allocation fails, OME isn't in a state where local recovery would help.

src/projects/publishers/webrtc/rtc_stream.cpp
getroot: Applied. Added the controller to the outer condition and dropped the now-redundant inner null check.

getroot:
@Keukhan 
The 500 ms is the Recompute interval for re-evaluating current_delay, not a dispatch wait. Frames are scheduled immediately in Push() at (expected_arrival + current_delay). A FPS change also doesn't change the lateness distribution by itself (same source pipeline, only the output rate changes), so current_delay doesn't need to react quickly here.

Keukhan (approved):
lgtm

PR #2098, opened 2026-04-30
test(rtmp): fix rtmp_test build after RTMP timestamp helper refactor

dimiden:
## Summary
- fix `ome_test_providers_rtmp` build failure caused by `rtmp_test.cpp` still calling removed `RtmpChunkParser` member methods
- update RTMP parser helper assertions in `rtmp_test.cpp` to use shared timestamp helper functions directly
- keep existing parser behavior coverage for both legacy RTMP and RTMP v2 test cases


getroot (approved):
lgtm

PR #2099, opened 2026-04-30
build(cmake): fix stubs install and improve NVIDIA linking condition

Keukhan:
### Summary

Fixes GPU stub libraries not being installed and improves NVIDIA hardware detection during cmake configuration.

### Changes:

- NVIDIA driver link condition — Replaced PKG_FFNVCODEC_FOUND flag check with explicit file existence `checks for libcuda.so, libcudart.so, libnvidia-ml.so, and nvcc`

- nv-codec-headers (ffnvcodec) install condition - used `ome_find_pkg `instead of `pkg_check_modules`

- GPU stubs install timing — Moved stubs installation from build phase to CMake configure phase stubs are installed automatically via InstallPrerequisites.cmake when` ${OME_DEP_PREFIX}/lib/stubs `directory is absent. The `OME_BUILD_STUBS `option is removed.


dimiden (approved):
LGTM

PR #2100, opened 2026-05-04
refactor(providers/api): prepare stream delete flow for API intent and watcher handoff

dimiden:
## Summary

Prepare shared stream deletion paths for API-requested termination and watcher-side handoff.

## Changes

- add `api_requested` to `Orchestrator::TerminateStream()`
- pass `true` from the stream delete API path
- update scheduled/multiplex watcher delete helpers to carry deleted stream snapshots
- align watcher delete flow so follow-up callbacks can be handled outside the watcher lock scope

## Why

Some stream deletion paths need to preserve the original delete trigger and
handoff deleted stream information through the watcher flow in a consistent way.

This change prepares the shared control flow first so later changes can stay
smaller and avoid repeatedly touching the same watcher and delete-path code.

## Notes

- no intended functional change for existing OSS behavior
- this is primarily a refactoring of shared delete-path plumbing

### DIFF
--- a/src/projects/api_server/controllers/v1/vhosts/apps/streams/streams_controller.cpp
+++ b/src/projects/api_server/controllers/v1/vhosts/apps/streams/streams_controller.cpp
@@ -250,7 +250,7 @@ namespace api
 			auto app_name	  = app->GetVHostAppName();
 			auto stream_name  = stream->GetName();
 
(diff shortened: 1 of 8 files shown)


getroot (approved):
Looks good to me

PR #2104, opened 2026-05-04
build(cmake): fix ffnvcodec auto-install regression

getroot:
## Problem
#2099 changed `PKG_FFNVCODEC`'s `REINSTALL_TARGET` from `nvcc_hdr`
to `ffnvcodec`, but the install map key in `InstallPrerequisites.cmake`
stayed as `_install_nvcc_hdr`. The lookup misses and silently no-ops,
so configure fails with `ffnvcodec - not found` whenever FFmpeg is
already installed but `ffnvcodec` is not. Clean installs dodge this
path because `TARGET=ffmpeg` provisions `ffnvcodec` first.

Also, just installing `ffnvcodec` headers does not add NVIDIA
encoders to an existing FFmpeg binary, so the reinstall has to
rebuild FFmpeg.

## Solution
- Rename `_install_nvcc_hdr` to `_install_ffnvcodec` to match the
  pkg-config module name.
- Change `PKG_FFNVCODEC`'s `REINSTALL_TARGET` to `ffmpeg` so the
  reinstall pulls in `ffnvcodec` via `_ffmpeg_deps` and rebuilds
  FFmpeg with NVIDIA support.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -330,7 +330,7 @@ ome_find_pkg(PKG_LIBAVUTIL      libavutil       OME_VER_LIBAVUTIL       REINSTAL
 
 # NVIDIA CUDA/NVML 
 if(OME_HWACCEL_NVIDIA)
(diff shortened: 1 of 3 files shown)


dimiden:
LGTM

dimiden (approved):

PR #2105, opened 2026-05-06
feat(webrtc): Implement AP (Aggregation Packets) support for H.264/H.265 RTP packetizer

Keukhan:
### Summary

Adds support for Aggregation Packets (AP) to the H.264 and H.265 RTP packetizer.
Before, NAL units larger than the MTU were split into FU-A packets, and all others were sent as Single NAL Units.
With this change, multiple NAL units can be combined into one RTP packet. 
This reduces the number of packets and header overhead.

- H.264: STAP-A (RFC 6184, NAL type 24)
- H.265: AP (RFC 7798, NAL type 48)

### Changes
- `GeneratePackets():` Added STAP-A/AP logic : only used when at least two NAL units can be aggregated
- `IsAggregationPossible():` New function to check if multiple fragments can fit into one packet
- `PacketizeStapA():` Fixed a bug in fragment_headers_length initialization (includes H.265 16-bit NAL header handling)
- `NextPacket()`: Moved Single NAL unit logic into a separate function NextSingleUnitPacket()

### DIFF
--- a/src/projects/modules/rtp_rtcp/rtp_packetizer_h264.cpp
+++ b/src/projects/modules/rtp_rtcp/rtp_packetizer_h264.cpp
@@ -122,7 +122,7 @@ bool RtpPacketizerH264::IsAggregationPossible(size_t fragment_index) const
 
 	// First fragment: NAL header (1) + length field (2) + data.
 	const size_t first_needed = kNalHeaderSize + kLengthFieldSize + _input_fragments[fragment_index].length;
-	if (first_needed >= space_left)
+	if (first_needed > space_left)
 	{
 		return false;
 	}
@@ -313,21 +313,27 @@ void RtpPacketizerH264::NextSingleUnitPacket(RtpPacket* rtp_packet)
 //  +---------------+
 //  |F|NRI| Type(24)|
 //  +---------------+
-//    F   = forbidden_zero_bit (from highest priority NALU)
-//    NRI = max nal_ref_idc among aggregated NALUs
+//    F   = forbidden_zero_bit 
+//    NRI = max of all NRI values in the aggregated NAL units
 //    Type = 24 (kStapA)
 void RtpPacketizerH264::NextAggregatePacket(RtpPacket* rtp_packet, bool last) 
 {
 	uint8_t* buffer = rtp_packet->AllocatePayload(last ? _max_payload_len - _last_packet_reduction_len : _max_payload_len);
 	PacketUnit* packet = &_packets.front();
 
-	// STAP-A NALU header.
-	buffer[0] = (packet->header & (kFBit | kNriMask)) | NaluType::kStapA;
+	uint8_t stap_f   = 0;
+	uint8_t stap_nri = 0;
 	size_t index = kNalHeaderSize;
 	bool is_last_fragment = packet->last_fragment;
 	
 	while (packet->aggregated) 
 	{
+		// F = OR of all aggregated F bits; 
+		// NRI = max of all aggregated NRI values.
+		stap_f   |= (packet->header & kFBit);
+		uint8_t cur_nri = packet->header & kNriMask;
+		if (cur_nri > stap_nri) stap_nri = cur_nri;
+
 		const Fragment& fragment = packet->source_fragment;
 		// Add NAL unit length field.
 		ByteWriter<uint16_t>::WriteBigEndian(&buffer[index], fragment.length);
@@ -337,7 +343,6 @@ void RtpPacketizerH264::NextAggregatePacket(RtpPacket* rtp_packet, bool last)
 		index += fragment.length;
 
 		_packets.pop();
-
 		_input_fragments.pop_front();
 
 		if (is_last_fragment)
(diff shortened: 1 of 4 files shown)


src/projects/modules/rtp_rtcp/rtp_packetizer_h264.cpp
```diff
@@ -241,7 +287,35 @@ bool RtpPacketizerH264::NextPacket(RtpPacket* rtp_packet)
...
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |  NALU 2 Size (16bit)          |  NALU 2 Data ...              |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+//
+// STAP-A NAL Header (1 byte):
+//  +---------------+
+//  |F|NRI| Type(24)|
+//  +---------------+
+//    F   = forbidden_zero_bit (from highest priority NALU)
+//    NRI = max nal_ref_idc among aggregated NALUs
```
dimiden: The STAP-A NAL header is built from the first aggregated NALU's header only:

```
buffer[0] = (packet->header & (kFBit | kNriMask)) | NaluType::kStapA;
```

Per [RFC 6184 5.7.1](https://datatracker.ietf.org/doc/html/rfc6184#section-5.7.1), the NRI field of the STAP-A header SHOULD be the maximum `nal_ref_idc` among all aggregated NAL units, and the F bit MUST be set if any of them has its F bit set. The comment above this function even documents this correctly (`NRI = max nal_ref_idc among aggregated NALUs`), but the implementation doesn't follow through.

Suggested fix: scan all aggregated `PacketUnit`s before writing `buffer[0]` and accumulate the max NRI and OR the F bits, then write the final header byte.

src/projects/modules/rtp_rtcp/rtp_packetizer_h265.cpp
```diff
@@ -62,7 +62,11 @@ bool RtpPacketizerH265::GeneratePackets()
 					PacketizeFuA(i);
 					++i;
 				} 
-				else 
+				else if (IsAggregationPossible(i))
```
dimiden, line 72: Before this PR, the `SingleNalUnit` fall-through was harmless because there was no aggregation branch. Now that `IsAggregationPossible()` is added, `SingleNalUnit` mode in H.265 can produce AP packets, which contradicts its semantics - one NALU per RTP packet.

The existing comment acknowledges this (`// For HEVC, doesn't support single nal unit mode at present`), but the fall-through now makes it actively wrong rather than just unimplemented.

Consider adding an explicit `break` (or a `return false`) for the `SingleNalUnit` case to prevent AP aggregation from being triggered in that mode.

src/projects/modules/rtp_rtcp/rtp_packetizer_h265.cpp
```diff
@@ -73,6 +77,36 @@ bool RtpPacketizerH265::GeneratePackets()
...
+	{
+		return false;
+	}
+
+	size_t space_left = _max_payload_len;
+	
+	// First fragment: NAL header (2) + length field (2) + data.
+	const size_t first_needed = H265_NAL_HEADER_SIZE + H265_LENGTH_FIELD_SIZE + _input_fragments[fragment_index].length;
+
+	if (first_needed > space_left)
```
dimiden, line 99: There's an inconsistency between the two `IsAggregationPossible()` implementations:

- H.264 (`rtp_packetizer_h264.cpp:125`): `if (first_needed >= space_left)`
- H.265 (`rtp_packetizer_h265.cpp:92`): `if (first_needed > space_left)`

When `first_needed == space_left`, H.264 returns `false` early while H.265 continues (and ultimately also returns `false` after the subtraction leaves zero space for the second fragment). The end result is identical, but having the same logic written differently across two paired implementations is a maintenance hazard.

Please unify to `>` (the H.265 version), which is the more natural read: "if the first fragment doesn't fit, bail out."

dimiden (changes requested):

src/projects/modules/rtp_rtcp/rtp_packetizer_h264.cpp
Keukhan: improvement to ensure that the H.264/H.265 payload header field values are set according to the RFC specification.
 👍

src/projects/modules/rtp_rtcp/rtp_packetizer_h265.cpp
Keukhan: Not sure why the code was originally written to fall back to NonInterleaved mode in H.265 SingleNalUnit mode, but it has been changed so that, like H.264, it packetizes using PacketizeSingleNalu in SingleNalUnit mode.

src/projects/modules/rtp_rtcp/rtp_packetizer_h265.cpp
Keukhan: Changes completed.

dimiden (approved):
Good job! 👍

PR #2111, opened 2026-05-07
docs: document <PlayoutDelay> option for WebRTC publisher

getroot:
## Problem
The `<PlayoutDelay>` option under `<Publishers><WebRTC>` is supported by the engine but missing from the WebRTC Publisher manual, so users have no reference for the XML schema, value range, or the initial audio/video offset that appears when `<Min>` is non-trivial.

## Solution
Document `<PlayoutDelay>` in `docs/streaming/webrtc-publishing.md`: add it to the example XML block, add a row to the options table, and add a `#### PlayoutDelay` subsection that explains the `<Min>`/`<Max>` range, the latency tradeoff, and the brief initial A/V offset users may observe at session start.


dimiden (approved):
Looks Great To Me

PR #2116, opened 2026-05-08
docs: fix HLS dump example placed inside LLHLS block

getroot:
## Summary

The `## Dump` section of [docs/streaming/hls.md](docs/streaming/hls.md) wrapped the HLS dump example in `<LLHLS>` tags. The HLS publisher only reads `<HLS><Dumps>`, so a user copying this example would configure an LL-HLS dump and see no HLS dump output.

Replaced the wrapping element (and the surrounding sentence) with `<HLS>`.

Addresses: #2092


dimiden (approved):
Looks good to me!

PR #2119, opened 2026-05-11
feat(publishers/ovt): add `TrackSet` for selective track delivery

dimiden:
## Summary

Add `TrackSet`, a new `OutputProfile` element that names a subset of an output stream's tracks. The OVT publisher accepts an optional trailing URL segment to select which `TrackSet` to expose:

```
    ovt://host/app/stream             -> full stream (unchanged)
    ovt://host/app/stream/<name>      -> only the selected tracks
```

An Edge can now pull a slice of an Origin stream (one rendition, one audio language, etc.) instead of every track produced by the profile.

## Configuration

```xml
<OutputProfile>
	<Encodes>
		<Video><Name>video_720p</Name>...</Video>
		<Audio><Name>aac_audio</Name>...</Audio>
	</Encodes>

	<TrackSet>
		<Name>edge_kor</Name>
		<Video>
			<Name>video_720p</Name>
		</Video>
		<Audio>
			<Name>aac_audio</Name>
			<IndexHint>1</IndexHint>
		</Audio>
	</TrackSet>
</OutputProfile>
```

- `<Name>` (required) - URL segment.
- `<Video>` / `<Audio>` - at least one entry required.
- `<IndexHint>` - pick a specific `GroupIndex` within a group (e.g. one language from a multilingual audio encode).
- `<Strict>` (default `false`) - when `true`, a missing encode reference fails output stream creation; when `false`, the entry is warned and skipped.

The same schema is accepted via TranscodeWebhook JSON (`tracksets`, `videos`, `audios`, `name`, `indexHint`, `strict`).

## Bundled fix

`VirtualHost::FindOriginByRequestedLocation` now picks the longest matching non-strict `<Location>` prefix instead of the first list match, so overlapping prefixes resolve unambiguously.
This was uncovered while testing TrackSet with multiple Edge `<Origin>` entries; the fix is small and orthogonal so it ships with this PR.

## Compatibility

- `ovt://host/app/stream` is unchanged.
- `TrackSet` is opt-in; existing configs and webhook responses are unaffected.
- The Origin-selection change strictly improves disambiguation; setups whose `<Location>` entries do not overlap see no behavior change.


src/projects/publishers/ovt/ovt_stream.cpp
```diff
@@ -250,13 +252,131 @@ bool OvtStream::GetDescription(Json::Value &description)
...
+{
+	if (GetState() != Stream::State::STARTED)
+	{
+		return false;
+	}
+
+	std::set<uint32_t> allowed_track_ids;
+	size_t video_count = 0;
+	size_t audio_count = 0;
+	if (ResolveTrackSetTrackIds(track_set_name, allowed_track_ids, video_count, audio_count) == false)
```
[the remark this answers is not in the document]
dimiden: These are temporary variables used to call `ResolveTrackSetTrackIds()`. Since they are passed by reference, they won't trigger unused-variable warnings.

dimiden:
@getroot 
Ready for review, PTAL.

src/projects/publishers/ovt/ovt_session.cpp
```diff
@@ -83,13 +83,70 @@ void OvtSession::SendOutgoingData(const std::any &packet)
...
+	// MESSAGE_RESPONSE and are sent via OvtPublisher::SendResponse, not this path).
+	if (_track_set_filter_enabled &&
+		session_packet->PayloadType() == OVT_PAYLOAD_TYPE_MEDIA_PACKET)
+	{
+		if (_current_group_decision == GroupDecision::Pending)
+		{
+			if (session_packet->PayloadLength() >= 4 && session_packet->Payload() != nullptr)
+			{
+				uint32_t track_id = ByteReader<uint32_t>::ReadBigEndian(session_packet->Payload());
+				if (_allowed_track_ids.find(track_id) != _allowed_track_ids.end())
```
getroot: Another thread sets this. It looks like a mutex is needed.

src/projects/publishers/ovt/ovt_session.cpp
```diff
@@ -83,13 +83,70 @@ void OvtSession::SendOutgoingData(const std::any &packet)
 		return;
 	}
 
+	// TrackSet filter (applied only to media packets; signaling packets are payload type
+	// MESSAGE_RESPONSE and are sent via OvtPublisher::SendResponse, not this path).
+	if (_track_set_filter_enabled &&
```
getroot: Another thread sets this. It looks like a mutex or automic is needed.

getroot (changes requested):

src/projects/publishers/ovt/ovt_session.cpp
dimiden: It's a good point! Fixed.

src/projects/publishers/ovt/ovt_session.cpp
dimiden: Fixed.

getroot (approved):

PR #2122, opened 2026-05-13
docs: rewrite legacy GitBook URLs to ovenmedialabs.com /docs/ome paths

getroot:
## Summary

Rewrite 28 in-tree links across 5 files in `docs-site/` that pointed at `docs.ovenmediaengine.com` or `airensoft.gitbook.io/ovenmediaengine` so they point at the consolidated Docusaurus pages on ovenmedialabs.com under `/docs/ome/...`.

## Why

ovenmedialabs.com is becoming the single home for the OvenMediaEngine manual. `docs.ovenmediaengine.com` and the airensoft GitBook will be redirected to it at the DNS layer during cutover, but in-tree links should already be relative to the new home so the manual reads cleanly when viewed locally, on GitHub, or as a static build — without depending on the external redirect layer.

### DIFF
--- a/docs-site/access-control/admission-webhooks.md
+++ b/docs-site/access-control/admission-webhooks.md
@@ -117,7 +117,7 @@ As shown below, the trigger condition of request is different for each protocol.
 | -------- | ------------------------------------------------------------------------------------------------------------------------ |
 | WebRTC   | When a client requests Offer SDP                                                                                         |
 | RTMP     | When a client sends a publish message                                                                                    |
-| SRT      | When a client send a [streamid](https://docs.ovenmediaengine.com/live-source/srt-beta#encoders-and-streamid) |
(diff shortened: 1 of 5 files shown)


getroot:
Thanks for the catch. Pushed a follow-up commit (ebf74289) that switches all the rewrites to repo-relative `.md` links (e.g. `streaming/webrtc-publishing.md#webrtc-over-tcp`), so the manual reads correctly in three places:

1. GitHub.com when viewing a docs file in the repo
2. Local Markdown viewers (VS Code, etc.)
3. The deployed Docusaurus site (Docusaurus normalizes relative `.md` links at build time)

One link to `/docs/ome-enterprise/...` is intentionally left as an absolute path: that target lives in a different repo, so a repo-relative link wouldn't resolve.

dimiden (approved):
LGTM

PR #2126, opened 2026-05-14
docs: add local preview script and docs-site README

getroot:
## Summary

Add `docs-site/preview.sh` so editors can spin up a local copy of ovenmedialabs.com against the current `docs-site/` and see their changes rendered in the real theme before merging. Add `docs-site/README.md` explaining how to run it and what to expect.

## Why

Until now, the only way to verify a docs edit was to merge first and check the live site afterwards. The preview script clones ovenmedialabs.com into a per-product cache, symlinks the editor's `docs-site/` into it, and starts a dev server in docs-only mode — marketing chrome and other products' docs are hidden so the editor stays focused on OvenMediaEngine. HMR picks up edits as files are saved.

### DIFF
--- a/docs-site/README.md
+++ b/docs-site/README.md
@@ -1,0 +1,42 @@
+# `docs-site/` — OvenMediaEngine docs source
+
+This folder holds the **MDX source** for the OvenMediaEngine user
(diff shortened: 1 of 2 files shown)


dimiden (approved):
LGTM

PR #2127, opened 2026-05-14
ci: notify ovenmedialabs.com when docs-site/ changes

getroot:
## Summary

When master receives a push that touches `docs-site/`, fire a `docs-changed` repository_dispatch at OvenMediaLabs/airensoft.com so the consuming site re-syncs and ships the change automatically.

## Why

Docs edits in this repo's `docs-site/` need to land on ovenmedialabs.com without a manual sync step. The downstream site listens for this dispatch and runs its sync workflow on receipt, so OvenMediaEngine docs ship as soon as they're merged here.


Keukhan (approved):
No lookprobe 👍

PR #2128, opened 2026-05-14
ci: allow 'ci' prefix in PR title check

getroot:
## Summary

Add `ci` to the conventional-commit type list enforced by the PR title check.

## Why

The title check rejects PRs starting with `ci:` because the type isn't on the allowed list, even though `ci:` is a standard Conventional Commits prefix and the more accurate label for CI-only changes than `build:` or `chore:`.


dimiden (approved):
It's so simple PR! Good!

PR #2131, opened 2026-05-14
ci: point check-docs-build at main now that the site has cut over

getroot:
The feat/docusaurus-migration branch was used while the consuming ovenmedialabs.com site lived only there. It's merged to main now.


dimiden (approved):
LGTM

PR #2136, opened 2026-05-17
fix: docs preview.sh — copy+watch instead of symlink, host/remote options

getroot:
## Summary
- `docs/preview.sh`: copy docs into the site cache (was a symlink) and mirror later edits with a small chokidar watcher — Docusaurus content-docs breaks when its `path:` is a symlink, so we copy yet keep HMR.
- Default the cloned ovenmedialabs.com branch to `main` (was a stale migration branch); add `OML_PREVIEW_HOST` + `--no-open` so VSCode Remote-SSH / headless previews work; pick the docs folder by product.

## Why
- A symlinked content path makes the build fail with `Cannot read properties of undefined (reading 'id')`; the stale branch previewed an outdated site; remote users hit a pointless browser-open with no host control.


dimiden (approved):
LGTM

Issue #2102, opened 2026-05-04
Need help with all pushes stalling when one push is having issues.


Keukhan:
### Discussed in https://github.com/OvenMediaLabs/OvenMediaEngine/discussions/2101

<div type='discussions-op-text'>

<sup>Originally posted by **vampirefrog** May  4, 2026</sup>
Hello!

Sometimes when I use OME for pushing (restreaming to multiple platforms), I get an error like this

[2026-05-03 20:34:19.650] W [SW-Push:388] ManagedQueue | managed_queue.h:611  | [490] mngq:v=#default#live:s=1x0rgxc3JyTUX9ZwQ3KzBvXWLWWeES7p:p=pub:n=streamworker_push has exceeded the threshold and increased peak. size: 940, threshold: 500, peak: 946, imps: 105, omps: 69

but all of the pushes now lag and the peak value keeps increasing. I am assuming it's one of the pushes having problems but it's causing problems for all of them, and the error doesn't make it clear which push has the problem. What can I do in this situation? I've been struggling with this issue for years now. How can I debug this situation? Is there info that's missing from that warning? Perhaps it could mention which push has this issue? Or is this issue in another place in the data pipeline? I really have trouble with this, especially when it's many pushes to 10-12 destinations.

I'm on 0.20.5 docker image.

Please help me debug this somehow.</div>

Keukhan:
@vampirefrog 
Thanks for reporting. as you mentioned, there is an issue where if the speed of one push session slows down, it also affects other push sessions.  so, I am planing to address this issue now. I will get back to you when the improvement is completed. 
Thanks.

vampirefrog:
Thank you I really appreciate it. Let me know if you need anything such as config or logs. This has been a long standing issue that has prevented me from adding restreaming to my product.

Keukhan:
@vampirefrog The patch has been completed and will be included in next release. if you need to test it, please try it on the master branch. if you confirm, let me know. I'll close the this issue. 👍

vampirefrog:
There's certainly a difference, especially if the bandwidth is being choked by a fast download, now the pushes disconnect and reconnect instead of stalling all of them, and it also tells you which push is disconnecting. So it's looking much better. I'm trying an overnight stream now and I'll let you know what the lag after it has been live for long hours.

PR #2142, opened 2026-05-19
fix: resolve issue where one push session delay affects other push sessions

Keukhan:
### Summary
Multiple push sessions shared a common stream worker thread. When one session experienced network send delays (slow/congested destination), it blocked the shared worker and caused all other push sessions to slow down as well. This is a regression introduced by the shared-worker architecture and reported in issue #2102.

To fix this, each push session now owns a dedicated sender thread and an internal packet queue. Incoming packets are enqueued immediately, and each session drains its own queue independently — so a slow session no longer blocks any other session.

Related: #2102

### Changes
- Dedicated sender thread per push session (push_session.cpp, push_session.h)
-- Added SenderThread, StartSenderThread, StopSenderThread methods

- Stream worker count reduced from 2 → 1 (push_stream.cpp)
-- Since packet dispatch now only writes to a queue (fast, non-blocking), a single stream worker is sufficient

- Error state transition fix (push_application.cpp)
--  Error state now calls session->Stop() before restarting, ensuring proper cleanup on reconnect

- Refactoring
--  Extracted SetErrorState() helper to eliminate duplicate error handling code across push_session.cpp

### DIFF
--- a/src/projects/modules/managed_queue/managed_queue.h
+++ b/src/projects/modules/managed_queue/managed_queue.h
@@ -361,10 +361,17 @@ namespace ov
 			auto shared_lock = std::shared_lock(_name_mutex);
 
 			ov::String urn_str = (_urn != nullptr) ? _urn->ToString() : ov::String("NoUrn");
+
+			ov::String threshold_info = ov::String::FormatString("%zu (%s %zu%s", _threshold, GetThresholdModeString(), _threshold_value, (_threshold_mode == ThresholdMode::TimeBased) ? "ms" : "");
+			if(_buffering_delay > 0)
+			{
+				threshold_info.AppendFormat(" + delay %dms", _buffering_delay);
+			}
+			threshold_info.Append(")");
+
 			return ov::String::FormatString(
-				"ManagedQueue [Id: %u, Size: %zu, Threshold: %zu (%s %zu%s + delay %dms), Peak: %zu, Imps: %zu, Omps: %zu, Wait: %s, Urn: %s]",
-				GetId(), _size, _threshold,
-				GetThresholdModeString(), _threshold_value, (_threshold_mode == ThresholdMode::TimeBased) ? " ms" : "", _buffering_delay,
+				"ManagedQueue [Id: %u, Size: %zu, Threshold: %s, Peak: %zu, Imps: %zu, Omps: %zu, Wait: %s, Urn: %s]",
+				GetId(), _size, threshold_info.CStr(),
 				_peak, _input_message_per_second, _output_message_per_second, _exceed_threshold_and_wait_enabled ? "On" : "Off",
 				urn_str.CStr());
 		}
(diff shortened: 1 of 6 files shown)


src/projects/publishers/push/push_session.cpp
```diff
@@ -231,37 +242,144 @@ namespace pub
...
 
-		uint64_t sent_bytes = 0;
+		_sender_packet_queue.Clear();
+	}
 
-		bool ret = writer->SendPacket(session_packet, &sent_bytes);
-		if (ret == false)
+	bool PushSession::IsSenderQueueExceededFor(std::chrono::milliseconds duration)
+	{
+		if (_sender_packet_queue.IsThresholdExceeded() == false)
```
dimiden: This looks like it could be racy if `IsThresholdExceeded()` can run concurrently with queue updates.

src/projects/publishers/push/push_session.cpp
```diff
@@ -191,6 +195,8 @@ namespace pub
 				push->UpdatePushStartTime();
 			}
 
+			StopSenderThread();
```
dimiden: Compared to the previous implementation, `Stop()` now waits for `PushSession::SenderThread()` to exit before stopping the writer.
If that thread is still blocked in `ffmpeg::Writer::SendPacket()`, this could delay `PushApplication::SessionControlThread()` during stop or restart. Was this intended?

src/projects/publishers/push/push_session.cpp
```diff
@@ -231,37 +242,144 @@ namespace pub
...
+				stream->GetApplicationInfo().GetVHostAppName(),
+				stream->GetName(),
+				"pub",
+				ov::String::FormatString("push_session_%u", GetId()).LowerCaseString());
+			_sender_packet_queue.SetUrn(urn);
 		}
 
-		auto push = GetPush();
-		if (push == nullptr)
+		_sender_packet_queue.SetThresholdByTime(5000);
```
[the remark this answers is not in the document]
Keukhan: No, there’s no need to wait. If the threshold is exceeded continuously, the session will be terminated.

src/projects/publishers/push/push_session.cpp
Keukhan: Good. I'll move `IsThresholdExceededFor` from PushSession into ManagedQueue. the `IsThresholdExceeded `function will be reverted to a private function.

src/projects/publishers/push/push_session.cpp
Keukhan: That could happen. I’ll improve it so that Writer::Stop() is called first, and then StopSenderThread() is called afterward.

src/projects/modules/managed_queue/managed_queue.h
```diff
@@ -19,7 +19,7 @@
 #include "base/info/managed_queue.h"
 #include "base/ovlibrary/ovlibrary.h"
 
-#define MANAGED_QUEUE_METRICS_UPDATE_INTERVAL_IN_MSEC 1000
+#define MANAGED_QUEUE_METRICS_UPDATE_INTERVAL_IN_MSEC 100
```
getroot: 100 is too much

getroot (approved):

src/projects/modules/managed_queue/managed_queue.h
Keukhan: restore to 1000 👍 I also concerned about performance

PR #2151, opened 2026-05-21
build: fix jemalloc not enabled by default in Release builds

getroot:
## Problem
`OME_ENABLE_JEMALLOC` was documented to default ON in Release builds, but the gate in `cmake/Dependencies.cmake` checked `NOT DEFINED OME_ENABLE_JEMALLOC`. Because `option()` always materializes the cache variable, that branch was never reachable and Release builds silently shipped without jemalloc.

## Solution
Move the default into the `option()` declaration based on `CMAKE_BUILD_TYPE` (ON for Release, OFF for Debug), matching the `OME_WHISPER_STATIC` pattern in the same block. The dependency gate simplifies to `if(OME_ENABLE_JEMALLOC)`. Explicit `-DOME_ENABLE_JEMALLOC=ON/OFF` still takes precedence.


Keukhan (approved):
LGTM

PR #2158, opened 2026-05-22
perf: reduce Whisper STT GPU memory and utilization

getroot:
## Problem
A Whisper STT encoder keeps its per-instance GPU state for the whole stream lifetime, even while STT is disabled. Each inference also runs the encoder over a full 30 second context regardless of how much audio the window actually holds. Both waste GPU resources.

## Solution
Release the per-instance state while STT is disabled and reallocate it when enabled. Size the encoder audio context to the actual window so the encoder no longer processes the 30 second zero-padding. The first reduces GPU memory, the second reduces GPU utilization.


Keukhan (approved):
lgtm

PR #2163, opened 2026-05-25
fix(webrtc): improve Pacer startup smoothness and per-track buffering

getroot:
## Problem

WebRTC publisher's Pacer absorbs source-side jitter well, but viewers see stutter right after startup. Three causes compound. First, the shared controller's percentile is dominated by audio (50pps near zero lateness), under-buffering video. Second, the first frame is taken as the anchor, and if it sits at an extreme of the publisher's startup burst it bakes a permanent baseline lateness offset. This can happen with any source, whether due to conservative BWE startup, first-keyframe paced burst, or encoder warm-up. Third, after a jitter spike `current_delay` contracts quickly, which the viewer perceives as variable playback rate.

## Solution

Anchor calibration combined with per-track percentile and asymmetric smoothing.

- **Per-track percentile, take max across tracks.** Mixing tracks into one sorted window biases the percentile toward audio. The controller now computes p99 per track and uses the max as \`current_delay\`. The single shared delay keeps A/V sync, while any track's tail can size the buffer.
- **Anchor calibration.** After each (re)anchor, the first 30 valid lateness samples are collected and \`anchor_arrival\` is shifted by their median. Baseline lateness then centers on zero, so the percentile rides on real source jitter rather than on a burst-induced offset.
- **Calibration outlier guard.** Frames whose lateness exceeds Pacer Max are excluded from the calibration set. Including such spikes would pull the anchor into the future, needlessly delaying dispatch of subsequent normal frames. The spikes themselves are handled by the controller (p99) and the player's jitter buffer.
- **Calibration timeout.** After 90 observed frames without enough valid samples, calibration force-completes with whatever exists (no-op shift if all were outliers). Bounds worst-case startup latency at low framerates or under extreme source jitter.
- **Idle reset 2s to 500ms.** After an idle, the first frame may be part of the publisher's resume burst, reducing anchor accuracy. Tightening the threshold catches shorter idles so the anchor is reset and calibration re-runs.
- **Asymmetric EMA on \`current_delay\`.** Rises apply immediately so jitter spikes aren't lost. Drops are smoothed (alpha=0.1) so a shrinking buffer doesn't show up to viewers as variable playback rate.
- **Percentile 0.95 to 0.99, margin 10ms to 20ms.** Remote source jitter typically has a long thin tail too frequent to leave to the receiver.
- **Drift reset logs split into info (startup phase) and warning (post-calibration)**, so operational logs only highlight genuine anomalies.

### DIFF
--- a/src/projects/modules/pacer/adaptive_delay_controller.cpp
+++ b/src/projects/modules/pacer/adaptive_delay_controller.cpp
@@ -17,19 +17,21 @@ static constexpr auto kWindow = std::chrono::seconds(30);
 // Recompute is throttled to this interval (avoids re-sorting per-frame).
 static constexpr auto kRecomputeInterval = std::chrono::milliseconds(500);
 
-// Target percentile of recent lateness. The Pacer absorbs everything up to
-// this percentile; the remaining tail (here ~5%) passes straight through to
-// the receiver, where the player's own jitter buffer covers it. Trading
-// "perfect" pacing for lower sender-side latency.
-static constexpr double kTargetPercentile = 0.95;
+// Target percentile of recent lateness. The Pacer absorbs up to this; the
+// remaining tail passes through to the receiver's jitter buffer.
+static constexpr double kTargetPercentile = 0.99;
 
-// Below this many samples, percentile estimate is unreliable; skip recompute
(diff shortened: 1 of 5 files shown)


Keukhan (approved):
Looks good to me. trust you.

PR #2165, opened 2026-05-26
fix(publishers/ovt): data race on shared `_description`

dimiden:
## Summary
- Concurrent `DESCRIBE` handlers race on the shared `_description` member: one thread reassigns it while another deep-copies it, freeing RB-tree nodes mid-traversal.
- Remove the member and have `GenerateDescription` write directly into a caller-owned `Json::Value`, so each request operates on its own object with no shared mutable state.

## Root cause
`OvtStream::GetDescription()` rebuilt the tree into `_description` (`_description = json_root` is copy-and-swap; the old map is destroyed in the temporary's destructor) and then copied it out into the caller's parameter.
With two parallel requests, the reader's `Rb_tree::_M_copy` could traverse nodes that the writer's swap had just freed, segfaulting deep in the JsonCpp deep-copy chain - see the recursive `_M_copy` / `dupPayload` stack frames in the crash log.

## Fix
- Drop the `_description` member.
- `GenerateDescription` now takes `Json::Value &out_description` and move-assigns the built tree into it.
- Both `GetDescription` overloads build directly into the caller's local - no shared state, no race, no extra synchronization required.

Incidentally removes one redundant deep copy per `DESCRIBE`. `DESCRIBE` is once-per-edge-connection and not on the media path, so omitting the cache is not expected to have a significant impact.

### DIFF
--- a/src/projects/publishers/ovt/ovt_stream.cpp
+++ b/src/projects/publishers/ovt/ovt_stream.cpp
@@ -95,7 +95,7 @@ bool OvtStream::Stop()
 	return Stream::Stop();
 }
 
-bool OvtStream::GenerateDescription()
+bool OvtStream::GenerateDescription(Json::Value &out_description)
 {
 	Json::Value 	json_root;
 	Json::Value		json_stream;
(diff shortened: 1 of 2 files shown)


getroot (approved):
LGTM

naanlizard:
I was just coming here to report this crash! Glad to see it is already fixed

PR #2166, opened 2026-05-26
fix(webrtc): cache certificates inserted before signalling/WHIP HTTPS servers exist

dimiden:
## Summary
- `WhipServer::InsertCertificate()` / `RtcSignallingServer::InsertCertificate()` no-op'd against an empty `_https_server_list`.
- After #2149, the WebRTC provider's `OnCreateHost()` runs before `Bind()` opens the listeners, so every cert insert was silently dropped and WHIP TLS handshakes failed with `there is no certificate`.
- Both servers now cache certificates by name and replay them under the same lock right after `Start()` populates `_https_server_list`. `Remove` keeps cache and live servers in sync.
- Lock order: outer `_http_server_list_mutex` -> inner `HttpsServer::_https_certificate_map_mutex` (`HttpsServer` never reaches back, so one-way).


Keukhan (approved):
lgtm

PR #2167, opened 2026-05-26
fix: suppress repeated incomplete-frame error logs in jitter buffer

getroot:
## Problem
When RTP packets are lost, an incomplete frame stays at the front of the jitter buffer and `CheckCompleted()` runs on every subsequent packet arrival via `BurnOutExpiredFrames()`. Each call emits the same `Invalid frame` error, producing dozens of identical lines per lost frame.

## Solution
Demote the message to debug level and gate it behind a per-frame flag so it logs at most once per affected frame. The existing trace log in `BurnOutExpiredFrames()` still records actual frame discards.


dimiden (approved):
LGTM

PR #2170, opened 2026-05-27
fix(push): fill in error reason for failed packet writes (#2079)

### LINKED ISSUE #2079: `ffmpeg/writer.cpp`: bitstream conversion failures silently, producing empty `Reason()` in downstream errors
**Describe the bug**
In `src/projects/modules/ffmpeg/writer.cpp`, some `return false` paths log a warning but never call `SetErrorMessage()`. Downstream callers that format `writer->GetErrorMessage()` into their own error log then emit an empty reason string, which makes the real failure effectively invisible.

**To Reproduce**
When a publisher like FileSession hits a malformed AAC ADTS frame, the recording dies with:                                                                           

```
E [SW-File:...] FILE | file_session.cpp:450 | Failed to write packet. Reason(), id(...) ...                                        
```
Note the empty Reason(). The actual cause is logged elsewhere, at warning level, with no correlation ID:                                                    
```
W [SW-File:...] FFmpegWriter | writer.cpp:549 | Failed to convert adts to raw                                                                            
```
For any user triaging a failed recording, the connection between these two log lines is not obvious, and `GetErrorMessage()` returning `""` blocks any programmatic handling further up the stack.

**Location**
https://github.com/OvenMediaLabs/OvenMediaEngine/blob/8a72d5d5b92c7b09925f84222692b717149ff4b3/src/projects/modules/ffmpeg/writer.cpp#L549
there are other examples using the same error template in the file too.

**Expected behavior**
I'd suggest promoting `logw` to a `loge` or an explicit `SetErrorMessage(...)`. Happy to open a PR if needed.

Upstream muxer corruptions like a 1-byte truncation inside an ADTS frame can happen and with the empty Reason(), a recording failure looks like an unknown internal error. With the real string propagated, users can immediately grep for "adts to raw" and follow it to the source.

**Logs**
```
[2026-04-14 02:11:01.367] W [AW-LLHLS0:110] FMP4 Packager | fmp4_packager.cpp:576  | FMP4Packager::ConvertBitstreamFormat() - Failed to convert adts to raw
[2026-04-14 02:11:01.367] W [AW-LLHLS0:110] FMP4 Packager | fmp4_packager.cpp:169  | Failed to convert bitstream format for track(1)
[2026-04-14 02:11:01.367] W [SW-File:13269] FFmpegWriter | writer.cpp:549  | Failed to convert adts to raw
```

**Server (please complete the following information):**
 - OS: Ubuntu 24.04.4 LTS (Noble Numbat) on kernel 6.17.0-1008-gcp (x86_64, GCP-optimized kernel)
 - OvenMediaEngine Version: v0.20.5
 - Branch: master

Keukhan:
## Summary

Fixes #2079 — when a packet write failed (e.g. malformed AAC ADTS frame), the
log showed `Reason()` empty because `SetErrorMessage()` wasn't called in some
failure paths. This PR makes every failure populate the error message with
useful context (track id, format, payload size).

Also fixes two data races in the writer that were exposed once the error
message is touched from the FFmpeg interrupt callback.

## Changes

- Every error path now sets a clear error message and logs it at the call
  site, so the reason is visible in both `Reason(...)` and the FFmpegWriter
  log.
- `_error_message` is now mutex-protected (was a race between publisher
  thread and FFmpeg interrupt callback).
- `_last_packet_sent_time` is now `std::atomic` (same race).
- Log levels organized: setup errors → `logte`, runtime drops → `logtw`

### DIFF
--- a/src/projects/modules/ffmpeg/writer.cpp
+++ b/src/projects/modules/ffmpeg/writer.cpp
@@ -10,17 +10,19 @@
 #include <modules/rtmp/amf0/amf_document.h>
 
 #define OV_LOG_TAG "FFmpegWriter"
+#define OV_LOG_PREFIX_FORMAT "%s"
+#define OV_LOG_PREFIX_VALUE _log_prefix.CStr()
 
-#define _logti(target, fmt, ...) logti("%s" fmt, (target)->_log_prefix.CStr(), ##__VA_ARGS__)
-#define _logtd(target, fmt, ...) logtd("%s" fmt, (target)->_log_prefix.CStr(), ##__VA_ARGS__)
-#define _logtw(target, fmt, ...) logtw("%s" fmt, (target)->_log_prefix.CStr(), ##__VA_ARGS__)
-#define _logtt(target, fmt, ...) logtt("%s" fmt, (target)->_log_prefix.CStr(), ##__VA_ARGS__)
-#define _logte(target, fmt, ...)                                               \
-	do                                                                         \
-	{                                                                          \
-		auto _err_msg = ov::String::FormatString(fmt, ##__VA_ARGS__);          \
-		logte("%s%s", (target)->_log_prefix.CStr(), _err_msg.CStr());          \
-		(target)->SetErrorMessage(_err_msg);                                   \
+// Redefine 'logae" to include the log prefix and also store the formatted message on the Writer
+#ifdef logae
+#	undef logae
+#endif
+#define logae(target, fmt, ...)                                                           \
+	do                                                                                    \
+	{                                                                                     \
+		auto _err_msg = ov::String::FormatString(fmt, ##__VA_ARGS__);                     \
+		logte(OV_LOG_PREFIX_FORMAT "%s", (target)->OV_LOG_PREFIX_VALUE, _err_msg.CStr()); \
+		(target)->SetErrorMessage(_err_msg);                                              \
 	} while (0)
 
 namespace ffmpeg
@@ -76,7 +78,7 @@ namespace ffmpeg
 		auto elapsed = std::chrono::steady_clock::now() - writer->GetLastPacketSentTime();
 		if(writer->GetState() == WriterStateClosed)
 		{
-			_logte(writer, "Writer was closed before I/O operation completed.");
+			logae(writer, "Writer was closed before I/O operation completed.");
 			return 1;
 		}
 		else if(writer->GetState() == WriterStateConnecting)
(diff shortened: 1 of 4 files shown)


src/projects/modules/ffmpeg/writer.cpp
```diff
@@ -55,22 +57,26 @@ namespace ffmpeg
 		Writer *writer = (Writer *)opaque;
 		if(writer == nullptr)
 		{
-			logte("Writer is null. stop the writer.");
+			logte("Writer is null.");
```
getroot, line 74: Please add the path where the issue occurs to the log, in the format like `vhost#app/stream.` Without it, this log will make it difficult to analyze problems when they come up in the future.

src/projects/modules/ffmpeg/writer.cpp
```diff
@@ -405,14 +435,14 @@ namespace ffmpeg
...
+			logtd("Writer is not initialized. state:%s", WriterStateToString(GetState()));
 			return true;
 		}
 
 		if (!packet)
 		{
-			SetErrorMessage("Packet is null");
-			return false;
+			logtw("Invalid packet. packet is null. Dropping.");
+			return true;
```
getroot, line 435: This PR is meant to make the error messages more specific, but the behavior has changed. Is that intended?

src/projects/modules/ffmpeg/writer.cpp
```diff
@@ -356,7 +384,9 @@ namespace ffmpeg
 		if (avformat_write_header(av_format.get(), &format_options) < 0)
 		{
 			SetState(WriterStateError);
-			SetErrorMessage("Could not create header");
+			auto err_msg = ov::String("Could not create header");
```
getroot: This pattern is repeated throughout, so for maintainability, how about consolidating it into a single function like the one below?

```
void Writer::SetErrorAndLog(ov::LogLevel level, const ov::String &msg) {
    ov_log(level, OV_LOG_TAG, "%s", msg.CStr());
    SetErrorMessage(msg);
}
```

getroot (changes requested):
The vhost#app/stream path is missing from the logs throughout, which makes them hard to analyze. Please add it everywhere.

src/projects/modules/ffmpeg/writer.cpp
Keukhan: The packet is already checked for null earlier, so this part of the code is rarely used. Still, even if a null gets through, it shouldn't stop the Writer. so the behavior was changed here.

src/projects/modules/ffmpeg/writer.cpp
Keukhan: add prefix info to File Session, Push Session and FFmpegWrite 👍

src/projects/modules/ffmpeg/writer.cpp
Keukhan: Putting it in one function looks simpler, but the logte line number didn't show up right in the debug logs, so I did it this way.   however, it seems like using a define would solve this. I've made the fix.

getroot (approved):
LGTM

PR #2176, opened 2026-06-04
fix: distribute WebRTC UDP ICE across multiple ports

getroot:
## Problem
The documentation said a single UDP ICE port with a higher IceWorkerCount scales across threads. In reality a UDP port is bound to one socket and serviced by a single thread, so that setup never used more than one CPU core.

## Solution
Document and default to a UDP port range (4 ports) for thread distribution, and update the Docker port mappings, EXPOSE, compose, and config examples to match. Remove the IceWorkerCount knob from examples since it has no effect for UDP today.


dimiden (approved):
LGTM

PR #2177, opened 2026-06-05
feat: handle concurrent RTP/RTCP receive on ICE candidate switch

getroot:
## Problem
OME previously supported only a UDP candidate or a TCP relay (TURN) and did not allow switching between them. It recently started supporting UDP, TCP, and relay candidates together with dynamic switching. A switch can move a session's media to a different socket-pool worker, so the RTP/RTCP receive path can now run on two threads concurrently and races on the shared receive state.

## Solution
Guard each piece of shared receive state (ssrc map, receive statistics, jitter buffers, transport-cc generator, depacketizer parameter sets) with its own minimal-scope mutex, and keep per-call scratch local so only genuinely shared data sits under a lock.

### DIFF
--- a/src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
+++ b/src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
@@ -669,11 +669,18 @@ std::shared_ptr<RtcpPacket> RtpRtcp::GenerateTransportCcFeedbackIfNeeded(const s
 {
 	std::shared_ptr<RtcpTransportCcFeedbackGenerator> generator;
 	{
+		// Common path after init: shared lock for a concurrent read.
+		std::shared_lock<std::shared_mutex> lock(_transport_cc_generator_lock);
+		generator = _transport_cc_generator;
+	}
+	if (generator == nullptr)
+	{
+		// First packet: take the exclusive lock and create, re-checking in case
+		// another thread won the race. Receiver SSRC is unknown, so reuse the
+		// first track's RR ssrc (wide sequence means media ssrc may not be unique).
 		std::lock_guard<std::shared_mutex> lock(_transport_cc_generator_lock);
 		if (_transport_cc_generator == nullptr)
 		{
-			// Receiver SSRC is unknown, so reuse the first track's RR ssrc. Wide
-			// sequence means media ssrc may not be unique, so the first one is used.
 			_transport_cc_generator = std::make_shared<RtcpTransportCcFeedbackGenerator>(_transport_cc_feedback_extension_id.load(), receiver_ssrc);
 		}
 		generator = _transport_cc_generator;
--- a/src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
+++ b/src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
@@ -382,10 +382,10 @@ std::shared_ptr<MediaTrack> RtpRtcp::GetTrack(uint32_t track_id) const
 	return it->second;
 }
 
-std::shared_ptr<RtpFrameJitterBuffer> RtpRtcp::GetJitterBuffer(uint8_t payload_type)
+std::shared_ptr<RtpFrameJitterBuffer> RtpRtcp::GetJitterBuffer(uint32_t track_id)
 {
 	std::shared_lock<std::shared_mutex> lock(_track_info_lock);
-	auto it = _rtp_frame_jitter_buffers.find(payload_type);
+	auto it = _rtp_frame_jitter_buffers.find(track_id);
 	if (it == _rtp_frame_jitter_buffers.end())
 	{
 		return nullptr;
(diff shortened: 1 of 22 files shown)


src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
```diff
@@ -570,14 +587,111 @@ std::optional<uint32_t> RtpRtcp::FindTrackId(uint8_t rtsp_inter_channel) const
...
+		auto stat = std::make_shared<RtpReceiveStatistics>(ssrc, clock_rate);
+		_receive_statistics[track_id] = stat;
+		return stat;
+	}
+	return it->second;
+}
+
+std::shared_ptr<RtpReceiveStatistics> RtpRtcp::FindReceiveStatistics(uint32_t track_id) const
+{
+	std::shared_lock<std::shared_mutex> lock(_receive_statistics_lock);
```
Keukhan, line 659: Not related to this PR changes, but noting it here. `_receive_statistics `is a container guarded by `_receive_statistics_lock`, but **line 281** `_receive_statistics.find(track_id_opt.value()) ` access it without the lock.

src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
```diff
@@ -570,14 +587,111 @@ std::optional<uint32_t> RtpRtcp::FindTrackId(uint8_t rtsp_inter_channel) const
...
+void RtpRtcp::SetLastSentRtpPacket(const std::shared_ptr<RtpPacket> &packet)
+{
+	std::lock_guard<std::shared_mutex> lock(_last_sent_packet_lock);
+	_last_sent_rtp_packet = packet;
+}
+
+void RtpRtcp::SetLastSentRtcpPacket(const std::shared_ptr<RtcpPacket> &packet)
+{
+	std::lock_guard<std::shared_mutex> lock(_last_sent_packet_lock);
+	_last_sent_rtcp_packet = packet;
```
Keukhan, line 707: Not related to this PR changes, but two lock-guarded variables are accessed without the lock. **Line 347** `_last_sent_rtcp_packet = rtcp_packet` is a direct assignment that without  `_last_sent_packet_lock`, so it should use the `SetLastSentRtcpPacket(rtcp_packet)`.

src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
```diff
@@ -570,14 +587,111 @@ std::optional<uint32_t> RtpRtcp::FindTrackId(uint8_t rtsp_inter_channel) const
...
+std::shared_ptr<RtpReceiveStatistics> RtpRtcp::GetOrCreateReceiveStatistics(uint32_t track_id, uint32_t ssrc, uint32_t clock_rate)
+{
+	std::lock_guard<std::shared_mutex> lock(_receive_statistics_lock);
+	auto it = _receive_statistics.find(track_id);
+	// Some encoders or servers do not provide SSRC in SDP, so it is extracted from
+	// the received packet. If the ssrc changes, the previous statistics are replaced.
+	if (it == _receive_statistics.end() || it->second->GetMediaSSRC() != ssrc)
+	{
+		auto stat = std::make_shared<RtpReceiveStatistics>(ssrc, clock_rate);
+		_receive_statistics[track_id] = stat;
```
Keukhan, line 651: Not directly related to this PR changes, but this variable are accessed without the lock. **Line 326** `_receive_statistics.find(track_id)` accesses it without `_receive_statistics_lock`

src/projects/modules/rtp_rtcp/rtp_rtcp.h
```diff
@@ -151,20 +162,24 @@ class RtpRtcp : public ov::Node
...
-	// track_id : Receiver Statistics
+	std::atomic<bool> _transport_cc_feedback_enabled = false;
+	std::atomic<uint8_t> _transport_cc_feedback_extension_id = 0;
+
+	// _receive_statistics_lock guards _receive_statistics (track_id : receiver statistics)
 	std::unordered_map<uint32_t, std::shared_ptr<RtpReceiveStatistics>> _receive_statistics;
+	mutable std::shared_mutex _receive_statistics_lock;
 
 	// Per-track receive-side NACK generator.
 	std::unordered_map<uint32_t, std::shared_ptr<RtpNackGenerator>> _nack_generators;
```
Keukhan, line 185: No related to this PR changes. **Line 848  of rtp_rtcp.cpp** `_nack_generators.find(track_id) `accesses the map without any lock.  It should use the `_state_lock`

Keukhan (changes requested):

src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
getroot: Good catch. Fixed.

src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
getroot: Good catch. Fixed.

src/projects/modules/rtp_rtcp/rtp_rtcp.h
getroot: Good catch. Fixed.

src/projects/modules/rtp_rtcp/rtp_rtcp.cpp
getroot: Good catch. Fixed.

Keukhan (approved):
LGTM

PR #2178, opened 2026-06-05
fix: key RTP receiver jitter-buffer and track maps by track_id

getroot:
## Problem
The RTP receiver's jitter-buffer and track maps were declared with a `uint8_t` key, while every insertion and lookup uses the `uint32_t` track id. Any track whose id exceeds 255 is silently truncated, so it collides with another track or fails to resolve.

## Solution
Change the key type of these maps to `uint32_t` to match the track id.


dimiden (approved):
LGTM

PR #2179, opened 2026-06-06
refactor: use find()/iterator instead of map operator[] under shared_lock

getroot:
Several read paths held a std::shared_lock on a shared_mutex but indexed a guarded std::map with operator[], which inserts a default element when the key is absent, so a missing key turned a read into a concurrent write. In HlsStream::AppendMediaPacket and TranscoderStream::OnDecodedFrame the key can genuinely be missing (the latter indexed a different map than the one its guard checked).

Replace operator[] with find()/iterator access in these shared-lock sections so they are strictly read-only; an absent key now returns early or falls back to a default without mutating the map.

None of these are known to cause failures in practice (the missing-key paths are rare or effectively single-threaded), so this is closer to defensive hardening and cleanup than a critical fix.

The rtp_rtcp module has the same class of issue but is intentionally left out here, since its receive-path thread-safety is handled in #2177.

### DIFF
--- a/src/projects/base/publisher/stream.cpp
+++ b/src/projects/base/publisher/stream.cpp
@@ -117,13 +117,14 @@ namespace pub
 	std::shared_ptr<Session> StreamWorker::GetSession(session_id_t id)
 	{
 		std::shared_lock<std::shared_mutex> lock(_session_map_mutex);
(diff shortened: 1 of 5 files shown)


Keukhan (approved):
lgtm

PR #2181, opened 2026-06-08
fix: make socket and metrics timestamp fields atomic

getroot:
ThreadSanitizer reports data races on timestamp fields that are read and written across threads without synchronization: ov::Socket last recv/sent time (timeout timer vs socket worker) and mon::CommonMetrics last update/recv/sent time (provider input vs publisher output threads).

Make those time_point members std::atomic, matching the atomic counters CommonMetrics already uses, with relaxed loads and stores since each field is only an independent timestamp snapshot and guards no other state.


dimiden (approved):
LGTM

PR #2187, opened 2026-06-11
feat(webrtc): support AV1 ingest over WHIP

getroot:
WHIP ingest negotiates only H264/H265/VP8, so AV1 senders such as OBS and Chrome cannot publish.

Add AV1 to the WebRTC SDP codec negotiation and a new AV1 RTP depacketizer that reassembles OBU temporal units into the low-overhead AV1_OBU stream the MediaRouter normalizer already consumes.

### DIFF
--- a/src/projects/modules/rtp_rtcp/rtp_depacketizer_av1.cpp
+++ b/src/projects/modules/rtp_rtcp/rtp_depacketizer_av1.cpp
@@ -1,0 +1,203 @@
+#include "rtp_depacketizer_av1.h"
+
+#include <modules/bitstream/av1/av1_parser.h>
+
+#define OV_LOG_TAG "RtpDepacketizerAV1"
+
+// Aggregation header bit masks (AV1 RTP Specification).
+#define AV1_AGG_Z 0x80
+#define AV1_AGG_Y 0x40
+#define AV1_AGG_W 0x30
+#define AV1_AGG_W_SHIFT 4
+
+// obu_header: obu_has_size_field is bit 1 (forbidden|type(4)|extension_flag|has_size_field|reserved).
+#define AV1_OBU_HAS_SIZE_FIELD 0x02
+
+std::shared_ptr<ov::Data> RtpDepacketizerAV1::ParseAndAssembleFrame(std::vector<std::shared_ptr<ov::Data>> payload_list)
+{
+	if (payload_list.empty())
+	{
+		return nullptr;
+	}
+
+	// Step 1: reassemble OBU elements across packets. Z continues the previous packet's last
+	// element; Y holds the current element open for the next packet. Elements are stored without
+	// a size field, exactly as carried over RTP.
+	std::vector<std::shared_ptr<ov::Data>> obu_elements;
+	std::shared_ptr<ov::Data> current = nullptr;
+	size_t reserve_size = 0;
+
+	for (const auto &payload : payload_list)
+	{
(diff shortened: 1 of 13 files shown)


src/projects/providers/webrtc/webrtc_stream.cpp
```diff
@@ -10,6 +10,7 @@
 #include "webrtc_stream.h"
 
 #include "base/ovlibrary/converter.h"
+#include "base/ovlibrary/dump_utilities.h"
```
dimiden: Since the AV1 dump block is disabled in regular builds due to `#if 0`, this code will go unused.

Therefore, what do you think about cleaning this up in another way? For example, we could wrap this in `#if 0` as well, or make it active only when `OME_DUMP_WHIP` is defined.
(For reference, RTMP dumps only in DEBUG mode when the `OME_DUMP_RTMP` environment variable is defined.)

src/projects/modules/rtp_rtcp/rtp_depacketizer_av1.cpp
```diff
@@ -0,0 +1,203 @@
...
+		return nullptr;
+	}
+
+	// Step 2: rewrite each OBU into the low-overhead bitstream format (obu_has_size_field == 1).
+	auto bitstream = std::make_shared<ov::Data>(reserve_size);
+	ov::ByteStream stream(bitstream);
+
+	for (const auto &element : obu_elements)
+	{
+		if (WriteObuWithSizeField(stream, element->GetDataAs<uint8_t>(), element->GetLength()) == false)
```
dimiden, line 154: A zero-length OBU element (W=0 path with a LEB128 length of 0) produces an empty `ov::Data`, which makes `WriteObuWithSizeField()` return `false` and drops the entire temporal unit.
A single stray empty element shouldn't discard the whole frame - skip zero-length elements during assembly instead of failing the TU.

src/projects/providers/webrtc/webrtc_stream.cpp
getroot: applied it

src/projects/modules/rtp_rtcp/rtp_depacketizer_av1.cpp
getroot: fixed it

src/projects/providers/webrtc/webrtc_stream.cpp
```diff
@@ -881,11 +897,35 @@ namespace pvd
...
+		// <app>/dump/whip/<stream>_track<id>.obu. Each temporal unit is prefixed with a Temporal
+		// Delimiter OBU so the file is a standard low-overhead .obu stream that ffmpeg/dav1d can
+		// decode directly.
+		static const bool dump_whip = ov::Converter::ToBool(std::getenv("OME_DUMP_WHIP"));
+		if (dump_whip && bitstream_format == cmn::BitstreamFormat::AV1_OBU)
+		{
+			auto safe_name = GetName().Replace("/", "_").Replace("#", "_");
+			auto dump_path = ov::PathManager::Combine(
+				ov::PathManager::GetAppPath("dump/whip"),
+				ov::String::FormatString("%s_track%u.obu", safe_name.CStr(), track->GetId()));
```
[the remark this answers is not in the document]
getroot: no necessary

dimiden (approved):
LGTM

PR #2188, opened 2026-06-12
ci: add CLA Assistant GitHub Action workflow

choigilhoon:
## Summary

Replaces the externally hosted **cla-assistant.io** GitHub App with an **in-repo GitHub Action** for collecting Contributor License Agreement signatures. Contributors will be able to sign the CLA directly in the pull request by posting a comment, with no round-trip to an external service — this addresses the latency of the current SaaS-based flow.

### What changes
- Adds `.github/workflows/cla.yml` using [`contributor-assistant/github-action`](https://github.com/contributor-assistant/github-action), pinned to the **v2.6.1 commit** (`ca4a40a`) of our fork at **`OvenMediaLabs/cla-github-action`**. Upstream was archived in 2026-03, so we run from a fork pinned by commit SHA to avoid supply-chain/archive risk.
- Signature records are stored in the private **`OvenMediaLabs/cla-signatures`** repo. The `master` branch here is protected and cannot receive signature commits, so a dedicated remote repo is used.
- Reuses the existing `CLA.md` as the signed document.

### How signing works
1. A contributor opens a PR → the bot comments asking unsigned contributors to sign, and sets a failing status check.
2. The contributor comments: `I have read the CLA Document and I hereby sign the CLA`.
3. The signature is recorded in `cla-signatures` and the status check turns green. `recheck` re-runs the check.

Only bots (`*[bot]`) are exempt via the allowlist; maintainers sign like any other contributor.

## Prerequisites before merge
- [x] `CLA_SIGNATURES_TOKEN` repo secret set (PAT with write access to `cla-signatures`).
- [x] Existing signers migrated from cla-assistant.io into `signatures/version1/cla.json` (17 signers across both CLA document versions; the only diff between versions was the company rename AirenSoft → OvenMedia Labs).
- [x] Allowlist policy decided: bots only.

## Rollout
This runs **in parallel** with the existing cla-assistant.io app. After verifying the full flow on a test PR (and that migrated signers are not re-prompted), the cla-assistant.io app and its required status check will be removed.

## Test plan
- Open a test PR from a non-allowlisted account → confirm the bot requests a signature.
- Comment the sign phrase → confirm the check passes and an entry appears in `cla-signatures`.
- Confirm a migrated signer is not asked to re-sign.

PR #2192, opened 2026-06-12
fix(ci): grant actions:write so CLA check updates after signing

choigilhoon:
## Problem

After a contributor signs the CLA by commenting on a PR, the `CLA Assistant` status check stayed **failing** even though the signature was recorded in `cla-signatures`.

Root cause: the `CLA Assistant` check is the workflow job's own check-run. When someone signs via comment, the action records the signature and then **re-runs the `pull_request_target` workflow** so the job re-evaluates and the check turns green. That re-run API call requires the **`actions: write`** permission, which was dropped from the `permissions` block during review of #2188. Without it the action logs:

```
All contributors have signed the CLA ✅
##[error]Error occurred when re-running the workflow: HttpError: Resource not accessible by integration
```

So the signature is saved but the red check never clears.

## Fix

Add `actions: write` back to the workflow `permissions` block (it is part of the action's official example for exactly this reason). `contents` stays `read` because signatures are stored in the remote `cla-signatures` repo via the PAT, not in this repo.

## Test plan

After merge: with a contributor removed from the signatures file, open a PR → check fails & merge is blocked → comment the sign phrase → confirm the check flips to passing and the PR becomes mergeable.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -7,6 +7,9 @@ on:
     types: [opened, synchronize]
 
 permissions:
+  # actions: write is required so the action can re-run the PR workflow
+  # after a contributor signs, flipping the CLA Assistant check to passing.
+  actions: write
   contents: read
   issues: write
   pull-requests: write


Keukhan (approved):
looks good

PR #2194, opened 2026-06-12
fix: prevent Whisper STT encoder crash on stream stop

getroot:
Problem: With STT disabled the Whisper encoder never allocates its per-instance state, but on stream stop the audio-collection loop exits on the kill flag and falls through to the inference block, dereferencing the null state and crashing.

Solution: After the collection loop, skip the inference block when the encoder is stopping, muted, or has no allocated state.


Keukhan (approved):
No look approve

PR #2197, opened 2026-06-15
feat: warn when an unready track blocks stream preparation

getroot:
A stream is marked prepared only after every track has valid media. If one negotiated track never receives decodable media, such as a WebRTC simulcast layer or an audio track that never arrives, the stream stays unprepared indefinitely with no log indicating which track is responsible.

MediaRouter now monitors inbound streams that have started receiving media but are not yet prepared, and periodically logs a warning for each track that has stayed invalid beyond a timeout, identifying the track so the cause is visible.

example:

```
[06-15 15:23:48.625] W [InboundWorker:1106411] MediaRouter | mediarouter_stream.cpp:175  | [#default#app/stream] Track #3 (Video) has not received valid media for 3003 ms; the stream cannot be prepared until this track is ready
[06-15 15:23:51.626] W [InboundWorker:1106411] MediaRouter | mediarouter_stream.cpp:175  | [#default#app/stream] Track #3 (Video) has not received valid media for 6004 ms; the stream cannot be prepared until this track is ready
[06-15 15:23:54.627] W [InboundWorker:1106411] MediaRouter | mediarouter_stream.cpp:175  | [#default#app/stream] Track #3 (Video) has not received valid media for 9005 ms; the stream cannot be prepared until this track is ready
[06-15 15:23:57.642] W [InboundWorker:1106411] MediaRouter | mediarouter_stream.cpp:175  | [#default#app/stream] Track #3 (Video) has not received valid media for 12020 ms; the stream cannot be prepared until this track is ready
[06-15 15:24:00.656] W [InboundWorker:1106411] MediaRouter | mediarouter_stream.cpp:175  | [#default#app/stream] Track #3 (Video) has not received valid media for 15034 ms; the stream cannot be prepared until this track is ready
[06-15 15:24:03.660] W [InboundWorker:1106411] MediaRouter | mediarouter_stream.cpp:175  | [#default#app/stream] Track #3 (Video) has not received valid media for 18038 ms; the stream cannot be prepared until this track is ready
```

### DIFF
--- a/src/projects/mediarouter/mediarouter_application.cpp
+++ b/src/projects/mediarouter/mediarouter_application.cpp
@@ -937,9 +937,17 @@ void MediaRouteApplication::InboundWorkerThread(uint32_t worker_id)
 
 		// When the inbound stream is finished parsing track information,
 		// Notify the Observer that the stream is parsed
-		if (stream->IsStreamPrepared() == false && stream->IsStreamReady() == true)
+		if (stream->IsStreamPrepared() == false)
 		{
-			NotifyStreamPrepared(stream);
+			if (stream->IsStreamReady() == true)
+			{
+				NotifyStreamPrepared(stream);
+			}
+			else
+			{
+				// Warn if a track has not become valid in time, blocking the stream from being prepared
+				stream->CheckUnpreparedTrackTimeout();
+			}
 		}
 
 		std::shared_lock<std::shared_mutex> lock(_observers_lock);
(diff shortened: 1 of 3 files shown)


Keukhan (approved):
Looks good. I could not find any problem. 

but, I think preparing often takes longer than 3 second even in normal cases, so i'm worried the wanring message may appear to frequently.

getroot:
@Keukhan Since this is a very specific case where a stream isn't prepared for more than 3 seconds after being created, log spam is unlikely to happen. And in that situation, the log spam (one line every 3 seconds) might actually be useful.

PR #2199, opened 2026-06-15
fix(provider): stop OVT pull re-DESCRIBE storm from a reachable-but-silent origin

naanlizard:
The below is Claude-written (as is the change), from analysing crash dumps we've been seeing in prod, plus a back-and-forth with @getroot on the approach. The crash itself (the `OvtStream::_description` race) is already fixed on master by #2165; this fixes the OVT connection storm that was driving it.

Reworked to @getroot's suggested approach: instead of a re-pull backoff with new `PullStream` state, reset the no-input clock on a successful reconnect.

### Problem

When an edge relays a stream from an origin and the source stops sending media **while its TCP connection stays up** (a stalled encoder whose host keeps answering at the socket level), the origin keeps the stream alive. The edge's `WhiteElephantStreamCollector` then enters a tight re-pull loop:

- It runs every 100 ms. A pull stream with no incoming packets for `NoInputFailoverTimeout` is `Stop()`-ed; on the next tick it is `STOPPED`, so the collector re-pulls it — reconnecting and re-`DESCRIBE`ing the origin, with no delay.
- The reconnect *succeeds* (the stream still exists at the origin) but delivers no media, so `last_recv` stays old and the stream is stopped and re-pulled again on the very next tick.

Three edges relaying one silent stream produce a continuous **DESCRIBE storm** (~25/s per stream) at the origin's OVT publisher.

(The collector re-pull loop — and thus this storm — only engages when `RetryCount > 0`; with the default `RetryCount = 0`, `ResumeInternal` terminates a silent pull stream on its first re-pull instead of reconnecting. The reproduction and our deployment use `RetryCount = 2`.)

### Reproduction

Minimal Docker reproduction (origin + 3 edges + a publisher frozen with `docker pause` to model a silent-but-connected source): https://github.com/naanlizard/ome-ovt-flap-repro

### Fix

Measure the no-input timeout from `max(last_recv, last_reconnect)` instead of `last_recv` alone:

- The collector keeps a thread-local map of the last time it resumed each pull stream, stamped only on a *successful* reconnect.
- A successful reconnect to a reachable-but-silent origin delivers no media, so `last_recv` stays old; measuring from `max(last_recv, last_reconnect)` means the stream is not stopped again until a full `NoInputFailoverTimeout` has elapsed *since the reconnect*. The re-pull (re-DESCRIBE) cadence therefore settles at `NoInputFailoverTimeout` (default 3 s) instead of once per 100 ms tick.
- No new `PullStream` members and no lock — the map is touched only by the collector thread and pruned each tick for streams that no longer exist. The monitoring `last_recv` is never written, so reported stats stay honest. Both collector resume sites (the no-input retry and the failback URL switch) reset the clock.

### Logging

The silent-origin lifecycle is now visible in the edge log:

- **reconnect** — `debug`: each time the collector re-pulls and reconnects to the origin. Off at the default level; raise the edge to `debug` to watch the cadence.
- **no input → stop** — `warning` (unchanged): prints the true time since the last *media* packet, which keeps growing while the origin stays silent.
- **media resumed** — `info`: once, when media starts flowing again after a re-pull.

### Behaviour (from the reproduction, 3 edges, one silent stream, 20 s window)

| | before | after |
|---|---|---|
| origin OVT re-pulls | 506 (~25/s) | 36 (≈ one per `NoInputFailoverTimeout` per edge) |
| edge `no incoming packets` stops | 255 | 18 |
| failover to a healthy backup URL | immediate | immediate |
| recovery when the source returns | immediate (via viewer pull) | immediate (via viewer pull) |

`NoInputFailoverTimeout` / `RetryCount` / `UnusedStreamDeletionTimeout` semantics are unchanged.

### Behaviours to be aware of

- **While retries are enabled, a reachable-but-silent origin is re-DESCRIBEd once per `NoInputFailoverTimeout` indefinitely; it does not reach `TERMINATED`.** A successful reconnect resets `_restart_count`, so the relay is never abandoned while the origin is reachable — by design. This change caps the DESCRIBE rate at the timeout cadence; it does not end the relay. Tearing such a stream down is what `PacketSilenceTimeoutMs` on the origin is for — the origin-side fix is the root cause; this patches the storm in the window before the origin removes the silent stream.
- **The stop log reports true media-silence, not the value that triggered the stop.** The decision uses `max(last_recv, last_reconnect)`, but the log prints the real time since the last media packet (so it can exceed `NoInputFailoverTimeout` after a silent reconnect). This keeps the logged number honest.
- **The clock resets on real media, not on signaling.** `last_recv` advances only on media packets, not on the OVT DESCRIBE/PLAY exchange, so a silent reconnect never resets it.

### DIFF
--- a/src/base/provider/pull_provider/application.cpp
+++ b/src/base/provider/pull_provider/application.cpp
@@ -13,7 +13,9 @@
 #include "stream_props.h"
 #include "provider_private.h"
 
+#include <chrono>
 #include <cinttypes>
+#include <map>
 
 namespace pvd
 {
@@ -59,20 +61,49 @@ namespace pvd
 
 		auto global_no_input_timeout_ms = GetHostInfo().GetOrigins().GetProperties().GetNoInputFailoverTimeout(); 
 		auto global_unused_stream_timeout_ms = GetHostInfo().GetOrigins().GetProperties().GetUnusedStreamDeletionTimeout();
-		auto global_failback_timeout_ms = GetHostInfo().GetOrigins().GetProperties().GetStreamFailbackTimeout();	
-		
-		constexpr int64_t idle_wait_time_ms = 100; 
+		auto global_failback_timeout_ms = GetHostInfo().GetOrigins().GetProperties().GetStreamFailbackTimeout();
+
+		// The no-input timer below is measured from
+		// max(last_recv, last_reconnect) so a reachable-but-silent origin (the reconnect succeeds but no
+		// media follows) is not stopped and re-DESCRIBEd on every tick, which would flood the origin's
+		// OVT publisher. Touched only by this collector thread, so it needs no lock.
+		std::map<info::stream_id_t, std::chrono::steady_clock::time_point> last_reconnect_time;
+
+		constexpr int64_t idle_wait_time_ms = 100;
 		while (!_stop_collector_thread_flag.load())
 		{
 			auto streams = GetStreams();
+
+			// Drop reconnect timestamps for streams that no longer exist.
+			for (auto it = last_reconnect_time.begin(); it != last_reconnect_time.end();)
+			{
+				if (streams.find(it->first) == streams.end())
+				{
+					it = last_reconnect_time.erase(it);
+				}
+				else
+				{
+					++it;
+				}
+			}
+
 			for (auto const &x : streams)
 			{
 				auto stream = std::dynamic_pointer_cast<PullStream>(x.second);
 
+				auto current = std::chrono::steady_clock::now();
+
 				if (stream->GetState() == Stream::State::STOPPED || stream->GetState() == Stream::State::ERROR)
 				{
-					// Retry
-					ResumeStream(stream);
+					// Retry. On a successful reconnect, remember when: the no-input timer restarts from
+					// this reconnect, so a reachable-but-silent origin (the reconnect succeeds but
+					// delivers no media, leaving last_recv old) is not stopped and re-DESCRIBEd again on
+					// the very next tick.
+					if (ResumeStream(stream))
+					{
+						last_reconnect_time[stream->GetId()] = current;
+						logtd("Re-pulled %s/%s(%u): reconnected to the origin", stream->GetApplicationInfo().GetVHostAppName().CStr(), stream->GetName().CStr(), stream->GetId());
+					}
 				}
 				else if (stream->GetState() == Stream::State::TERMINATED)
 				{
@@ -81,8 +112,6 @@ namespace pvd
 				}
 				else
 				{
-					auto current = std::chrono::steady_clock::now();
-
 					// Default Properties of PullStream
 					auto is_persistent = false;
 					auto is_failback = false;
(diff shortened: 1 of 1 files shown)


naanlizard:
An aside, since this is a convenient way to ping y'all - @getroot @Keukhan 

I have been digging into color space (601/709, full/limited) and screenshots and how OME handles them. I'm in the final review phase of a PR to 

1. Fix color rendering on thumbnails generated by OME
2. Fix a bug with webP thumbnail encoding performance
3. Add AVIF image output format support
4. Add AV1 ingest thumbnail support

I'll hopefully push that in a few hours or tomorrow for review, including a big page I've been using to compare screenshots generated by various versions of OME (master, 0.20.5, my fork, and the ffmpeg upgrade commit that is part of my work)

getroot:
@naanlizard A very large refactoring is going to land on the Transcoder side this week. It would be better to put up any PRs that touch the transcoder after that.

naanlizard:
Is there somewhere I can see that work in progress?

getroot:
Nice analysis, and the backoff works. Personally I'd prefer to fix this with a more minimal change, since you already noted last_recv only advances on real media.

Instead of adding the backoff state, I'd just reset the no-input clock on a successful reconnect. The re-pull cadence then settles at NoInputFailoverTimeout on its own, with no new members and no lock, and existing viewers recover within the timeout instead of up to the 30s cap.

For the implementation, my idea is to keep a collector-local "last reconnect" timestamp and measure no-input from max(last_recv, last_reconnect) rather than stamping the monitoring last_recv itself, so the stats stay honest.

What do you think?

getroot:
> Is there somewhere I can see that work in progress?
@naanlizard It is the PR.
https://github.com/OvenMediaLabs/OvenMediaEngine/pull/2202

naanlizard:
I'd defer to your judgement, of course. That would be 3s (NoInputFailoverTimeout) between DESCRIBEs instead of the scaling-to-30s, right? 

The backoff seems better to me because 3s by default is still a bit tight, and if someone had a ton of edges, it could still cause a bunch of DESCRIBEs, but I guess that's an edge case. Your solution is certainly cleaner. 

An aside, PacketSilenceTimeoutMs defaults to 0 for RTMP, should that perhaps default to something like 5-10s? I'm not sure there are many situations where you have no media packets coming in for e.g. 5s and you should stay connected

getroot:
The root cause here is that a stream stays alive even when there is no data flow.
Setting PacketSilenceTimeoutMs on the origin resolves this, and the request storm from the edge goes away.
On the edge side, I think it would be good to apply the bug fix I described. It patches the storm during the window before the stream is removed by PacketSilenceTimeoutMs on the origin.
With that, I don't see a case that still needs the backoff. Other than this situation, are there cases where you think the backoff would still be needed?
And if some case intentionally needs to keep a silent stream alive, setting NoInputFailoverTimeout to a very large value covers it.

As for changing the default value of PacketSilenceTimeoutMs, let's handle that separately in a Discussion. There was a reason I set it to 0, but I don't remember it well right now, so I'll think it over.

naanlizard:
Updated

naanlizard:
Rebased onto current master (~ v0.21.0); the branch had gone stale against the src/projects flatten (#2248). The fix itself is unchanged — still the approach suggested above: no-input measured from max(last_recv, last_reconnect) via a collector-local map.

Current master doesn't crash any more, which is good, but there's still a ton of noise from the flapping.

getroot:
I pushed 174ac01b5 to this branch. It only rewords the "media resumed" log line.

The reconnect marker is set on every successful ResumeStream, so this log fires whenever media arrives after any re-pull, not only after a silent origin: an origin that went down and came back, a failback switch to the primary URL, or a move to the next URL in the list. The old wording described only the silent origin case. The new line, "stream has resumed receiving media after reconnecting", is accurate in all of them.

getroot (approved):

getroot:
Thank you for the PR. We had many Enterprise releases recently, so I did not have enough time to test this well and kept delaying it. Now I have time to test it fully, so I will merge it.

PR #2206, opened 2026-06-18
feat(llhls): add AV1 codec support to LLHLS fMP4 packaging

dimiden:
## Summary
- Add `av01`/`av1C` ISOBMFF box writers and AV1 OBU bitstream conversion with temporal delimiter stripping
- Build `ftyp` compatible brands dynamically per codec (H264/H265/AV1)
- Fix CENC member mutation bug that silently disabled DRM for subsequent tracks
- Update LLHLS doc to include AV1 in supported codec list


src/projects/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -602,7 +679,27 @@ namespace bmff
...
+
+				case cmn::MediaCodecId::H265:
+					stream.WriteText("hvc1");
+					break;
+
+				case cmn::MediaCodecId::Av1:
+					stream.WriteText("av01");
+					break;
+			}
+		}
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/projects/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -586,7 +589,81 @@ namespace bmff
...
+				{
+					logte("FMP4Packager::ConvertBitstreamFormat() - Failed to parse AV1 OBU at offset %zu", offset);
+					return nullptr;
+				}
+
+				if (obu.header.has_size_field == false)
+				{
+					logte("FMP4Packager::ConvertBitstreamFormat() - AV1 OBU at offset %zu missing obu_has_size_field", offset);
+					return nullptr;
+				}
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/projects/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -602,7 +679,30 @@ namespace bmff
...
+		{
+			switch (GetMediaTrack()->GetCodecId())
+			{
+				case cmn::MediaCodecId::H264:
+					stream.WriteText("avc1");
+					break;
+
+				case cmn::MediaCodecId::H265:
+					stream.WriteText("hvc1");
+					break;
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/projects/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -586,7 +589,88 @@ namespace bmff
...
+					logte("FMP4Packager::ConvertBitstreamFormat() - Failed to parse AV1 OBU at offset %zu", offset);
+					return nullptr;
+				}
+
+				// A missing size field is only valid for a TemporalDelimiter (stripped below) or for
+				// the terminal OBU of the sample. `ReadObu()` makes any other unsized OBU consume the
+				// rest of the buffer, so `next_offset == total` holds whenever it is genuinely terminal.
+				if (obu.header.has_size_field == false &&
+					obu.header.type != Av1ObuType::TemporalDelimiter &&
+					obu.next_offset != total)
```
getroot: I think, if OBU has no size_field, it is always obu.next_offset == total. So this condition is always false.

src/projects/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -586,7 +589,88 @@ namespace bmff
...
+
+				if (obu.header.type == Av1ObuType::TemporalDelimiter)
+				{
+					needs_filter = true;
+				}
+
+				offset = obu.next_offset;
+			}
+
+			if (needs_filter)
```
getroot: I think we could just run a loop to create the filtered variable instead of running a separate need_filter loop. What do you think? The need_filter loop seems like unnecessary extra work.

getroot: like this

```
while (offset < total)
			{
				if (Av1Parser::ReadObu(base, total, offset, obu) == false)
				{
					logte("FMP4Packager::ConvertBitstreamFormat() - Failed to parse AV1 OBU at offset %zu", offset);
					return nullptr;
				}

				if (obu.header.type == Av1ObuType::TemporalDelimiter)
				{
					stripped = true;
				}
				else
				{
					filtered->Append(base + obu.obu_offset, obu.next_offset - obu.obu_offset);
				}

				offset = obu.next_offset;
			}

			// No temporal delimiter: keep the original packet (converted_packet == media_packet).
			if (stripped)
			{
				if (filtered->GetLength() == 0)
				{
					return nullptr;
				}

				auto new_packet = std::make_shared<MediaPacket>(*media_packet);
				new_packet->SetData(filtered);
				converted_packet = new_packet;
			}
```

src/projects/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
dimiden: You're right. The original intent was a defensive check to reject OBUs that omit the size field, but `ReadObu()` makes any unsized non-TD OBU consume the rest of the buffer, so `next_offset == total` always holds and the condition can never fire. It was guarding against a case that can't occur, so I'll remove it.

src/projects/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
dimiden: Agreed. The separate scan loop was meant as an optimization to skip allocating/copying when there's nothing to strip, but AV1 emits a temporal delimiter almost every frame, so `needs_filter` is `true` nearly all the time and the fast path rarely triggers, while the extra pass is always paid. Merged it into a single loop that copies non-TD OBUs directly.

dimiden: Fixed.

src/projects/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
dimiden: Fixed.

getroot (approved):
Good

PR #2207, opened 2026-06-18
fix(llhls): block chunklist reload when _HLS_part is omitted

getroot:
Safari native LLHLS playback stalls a few seconds after start. The player requests a new segment's first part with `_HLS_msn` only (no `_HLS_part`), but the server skipped the blocking-reload check for part-less requests and returned a stale playlist immediately. This produced a tight request spin that saturated the connection pool, starved the video rendition, and froze playback. It regressed since #2037, which makes a completed segment emit its `EXTINF` before the next part exists and flips Safari from a part-indexed request to a part-less one.

Treat an omitted `_HLS_part` as part 0 in the chunklist blocking-reload check so the request blocks until the requested segment's first part is available. The existing pending-wake logic already handles a missing part index, so no other change is needed.


dimiden (approved):
LGTM

PR #2208, opened 2026-06-18
fix(transcoder): correct `OV_GUARDED_BY` placement on `_input_stream`

dimiden:
## What

Move the `OV_GUARDED_BY(_mutex)` annotation on `CompositeMap::_input_stream` to before the `= nullptr` initializer.


Keukhan (approved):
Thanks!!!

PR #2211, opened 2026-06-19
feat(thumbnail): AVIF thumbnails, AV1 ingest/decode, expanded image-encoder options, and a colour-managed thumbnail pipeline (on FFmpeg 8)

naanlizard:
## Summary

Like my previous PR, this is all coded by Claude but represents about 2 weeks of heavy back and forth and manual verification of rendered thumbs.

The primary motivation was to improve thumbnail color accuracy, as they can diverge pretty seriously from what the stream actually sends. I put together a comparison site for this — [OvenMediaEngine thumbnail colorimetry](https://naanlizard.github.io/ome-thumbnail-colorimetry/) — that renders the same sources through a variety of OME versions side by side; the default view is the current v0.20.5 output against this PR's. The repo it's built from ([naanlizard/ome-thumbnail-colorimetry](https://github.com/naanlizard/ome-thumbnail-colorimetry)) has the capture rig and the methodology: thumbnails are pulled back through a real browser canvas and compared (ΔRGB) against the colors the source actually authored, across each image format and a range of downscales.

Since I was also working on an AVIF output for thumbnails, I rolled everything into one PR commit ladder so as to avoid a storm of PRs. Just let me know if you'd rather I split it in some way (perhaps, color fixes, FFmpeg upgrade, AVIF support, something like that?). The commits stack in order rather than being fully independent — you could stop at any point and take a prefix, but a later group leans on the earlier ones (the color and AVIF work both sit on the FFmpeg 8 base, and AVIF builds on the image options). In order, in general terms they do the following:

1. Update FFmpeg through every breaking change for OME to the latest 8.0.2
2. Fix the SCTE-35 patch for FFmpeg to work with 8
3. Fix a small CMAKE build bug we found (rebuilding the project produces an unoptimized debug build)
4. Fix incorrect colors on thumbnails, fix WebP always encoding with the wrong preset.
5. Add quality config options for the thumbnail encoders
6. Add AVIF as an option for thumb output
7. Add per-thumb timing logging
8. Add AV1 decode for ingest, plus a zero-encode AV1 to AVIF passthrough

The most important for our use is the color fixes, but AVIF is a close second. The rest are not important to us but they were easy enough to add while we were in there.

The below is lightly edited by me but mostly Claude's original PR description.

---

This PR reworks the thumbnail scaler to do deterministic colour management across all four image codecs (JPEG/PNG/WebP/AVIF) and adds an AVIF thumbnail output. Supporting that, it brings in AV1 decode for ingest, uses the already-bundled libaom, and bumps the toolchain FFmpeg 5.1 → 8.0. AVIF is produced two ways: a full transcode path (decode → colour-managed rescale → libaom encode → AVIF mux), and a zero-encode **AV1→AVIF rewrap** that rewraps source AV1 key frames into single-image AVIF at the source resolution.

## What changes

- **Build**: FFmpeg 5.1.4 → 8.0.2 (sub-libs repinned) with source-compat fixes for the API removals (`pkt_duration`, `ticks_per_frame`, `AVFilter.flags_internal`, the `ff_filter_link` hw-frames accessor, const AVIO buffer); enables the `avif` muxer + `colorspace` filter (libaom is already bundled by #2205); re-rolls the SCTE-35 patch for FFmpeg 8; defaults `CMAKE_BUILD_TYPE` to Release when unset.
- **Media plumbing**: new `BitstreamFormat::AVIF` + `MediaCodecId::Avif` (an image codec; each packet is one complete single-image AVIF), threaded through the codec switches and module registration; `AV_CODEC_ID_AV1` becomes a decodable input.
- **Colour pipeline** (`filter_lavfi_rescaler`): AVIF converts to BT.709 full-range via the `colorspace` filter; JPEG/PNG/WebP get `out_color_matrix`/`out_range` on `scale`; untagged sources are stamped BT.709 at every resolution; PNG colour tags are stripped on encode.
- **Image-encoder options**: new `<Image>` options (`QScale`, `Quality`, `Method`, `Lossless`, `Preset`, `ChromaSampling`, `Speed`, `Crf`) registered, codec-applicability-checked, range-clamped, and applied to the JPEG/WebP/AVIF encoders.
- **AVIF encoder** (`encoder_avif`): a standalone encoder driving libaom-av1 all-intra, with deferred open to copy the first frame's CICP, av1C sequence-header extraction, and in-memory AVIF muxing.
- **AV1**: software decode for ingest (pre-framed, no parser needed), plus the rewrap path (`transcoder_stream`) that carries the source's CICP straight through.
- **Delivery + timing**: `.avif` served everywhere thumbnails are (new `image/avif` MIME); a debug-gated per-frame `ThumbStat` timer for thumbnail tracks.

## Notes for reviewers

- **`filters.h` install**: the FFmpeg 7.1+ `hw_frames_ctx` accessor needs `libavfilter/filters.h`, which FFmpeg doesn't install by default — the cmake step copies it manually.
- **Transcode vs rewrap differ on range by design**: both default unspecified matrix/primaries/transfer to BT.709; the transcode path forces full range (it range-expands samples), the rewrap path preserves the source's signalled range (it touches no pixels).
- **Untagged thumbnails are stamped BT.709 at all resolutions** (no SD/HD split) — this ensures thumbnails render with the same colors as the streams played via LLHLS/WebRTC
- **AVIF CRF `0` is a valid max-quality value**, so the unset sentinel is `-1` (defaulted to 30). A missing sequence header / av1C / mux failure drops that single thumbnail with a warning rather than erroring the stream.
- **The Xilinx `u30ma` patch is not re-rolled for FFmpeg 8**: it's opt-in (`ENABLE_XMA`, off by default) and a large vendor patch that can't be validated without U30 hardware. An `ENABLE_XMA` build warns and configures FFmpeg without it (upstream's existing warn-and-continue); re-roll `u30ma` for 8.0.2 if you need XMA.


Keukhan:
@naanlizard 

Thanks for all the work you put into this, and for adding the new color-related features in particular.
This PR is quite large and changes a few different things at once, so **I'd like to review it as a few smaller PRs, one per feature**, instead of one big one. Here's how we'd like to split it.

1. AV1 Decode + AVIF thumbnails 
I'm actually working on AV1 decoding right now to support AV1, and AVIF was already on my plan too, so this is great timing. It would be nice to have this as its own PR. One thing: instead of making a new `encoder_avif` file, could you put this code inside the existing avcodec-based `encoder_avcodec_image` class?


2. Color-managed image thumbnails
Honestly, I haven't paid much attention to color management until now. I'd like to use this chance to learn more about it first, and then review the PR. Once ( AV1 Decode + AVIF thumbnails) has been reviewed and merged, you can open this as a new PR.

3. FFmpeg upgrade
Updating FFmpeg would be nice, but because of some third-party libraries I depend on, we need to keep FFmpeg at 5.1.4 for now. Let's hold off on this part for now.

Thanks again for the great work! 👍

Keukhan:
Decided to split this into sperate PR

PR #2216, opened 2026-06-20
fix(mpegts): tolerate non-standard PTS/DTS prefix nibble in PES parser

### LINKED ISSUE #2215: SRT MPEG-TS ingest: PES parse fails (Invalid PTS_DTS start bits) for Moblin/libsrt sender — same stream ingests cleanly in SRS and srt-live-server
## Summary
A Moblin (libsrt, MPEG-TS H.264+AAC) SRT publish connects to OME and a stream object is created,
but OME's MPEG-TS PES parser immediately and continuously logs `Invalid PTS_DTS start bits: 10
(0A), expected: 2 (02)` and the stream never becomes playable. `RCV-DROPPED = 0` (no SRT loss).
The **byte-identical** Moblin stream ingests perfectly in `srt-live-server`, `SRS v6`, and via
ffmpeg — only Moblin → OME fails. A raw packet capture shows the cause precisely: the stream is
valid, 188-aligned MPEG-TS, but each PES carries a PTS whose 4-bit prefix is `0b1010` rather than
the spec-mandated `0b0010`. OME validates that prefix with strict equality and aborts; every other
demuxer ignores the prefix and reads only the timestamp bits.

## Environment
- **OME:** reproduced on **current `master`** (`v0.20.5 (bc43e18)`, `airensoft/ovenmediaengine:dev`, built 2026-06-19), on **v0.20.5** (latest release, 2026-03-26), and on **v0.20.0** — identical failure on all three (`Invalid PTS_DTS` floods within seconds; stream never publishes).
- **OS:** Linux x86_64, Docker.
- **Reproduced on two independent hosts:** a bare-metal node with OME on `hostNetwork` (no NAT) and a local Docker Desktop host on a clean 1 GbE LAN.
- **Config:** reproduces on **stock** config — a minimal provider is enough:
  ```xml
  <Providers>
    <SRT>
      <Port>9999</Port>
      <WorkerCount>1</WorkerCount>
    </SRT>
  </Providers>
  ...
  <Application><Providers><SRT /></Providers> ... </Application>
  ```
  (Also reproduces with `SRTO_RCVLATENCY=2000` set — latency is not the differentiator.)

## Encoder
[Moblin](https://github.com/eerimoq/moblin) (open-source IRL streamer, libsrt sender) — MPEG-TS over
SRT, **H.264 + AAC**, ~6.5 Mbit/s, SRT latency 2000 ms. Same result with a single SRT link and with
SRTLA-bonded links (`srtla_rec` → OME).

## Steps to reproduce
1. Run OME with the stock SRT provider above.
2. From Moblin, publish plain SRT to `srt://<host>:9999?streamid=default/app/test` (H.264 + AAC, latency 2000).
3. Watch the OME log.

## Expected
Stream publishes and plays (as it does in SRS and srt-live-server with the same encoder).

## Actual
SRT connects, stream is created, then PES parsing fails continuously and the stream never publishes:
```
SrtProvider | srt_provider.cpp:188 | The SRT client has connected : <... streamid: default/app/test>
Provider    | stream.cpp:51        | Unknown/test(100) has been started stream
MediaRouter | mediarouter_application.cpp:342 | [#default#app/test(100)] Trying to create a stream
Transcoder  | transcoder_stream.cpp:75 | [#default#app/test(100)] stream has been started
MPEG-2 TS   | mpegts_pes.cpp       | Invalid PTS_DTS start bits: 10 (0A), expected: 2 (02)
MPEG-2 TS   | mpegts_pes.cpp       | Could not parse PTS
MPEG-2 TS   | mpegts_pes.cpp       | Could not parse table header
MPEGTS_DEPACKETIZER | mpegts_depacketizer.cpp | Something wrong with parsing PES
... (repeats for every PES)
```

## Root cause (byte-confirmed)
Captured the raw SRT traffic with `tcpdump`, stripped the 16-byte SRT data-packet header, and
reassembled the carried MPEG-TS. Findings:

- The stream is **clean, standard, 188-byte-aligned** MPEG-TS — SRT payload size 1316 (= 7 × 188),
  first byte `0x47`, every 188-byte position is a sync byte. **This is not a desync / sync-loss
  problem.**
- The PES headers themselves parse fine (start code `00 00 01` present, `stream_id` valid). The
  failure is purely in the PTS field. Example first video PES (PID `0x100`, `stream_id 0xE0`):

  ```
  00 00 01 e0 64 e3 84 80 05 a3 86 cd 04 09
  └ start ┘ │  └len┘ │  │  │  └──── PTS ───┘
        code  E0     84 80 05   a3 86 cd 04 09
                     │  │  └ PES_header_data_length = 5
                     │  └ flags byte: pts_dts_flags = 0b10 (PTS only)  ✔
                     └ '10' marker bits  ✔
  ```
  The PTS field begins at `0xA3` = `1010 0011`. The first 4 bits (the PTS prefix) are **`0b1010`**,
  whereas ISO 13818-1 specifies **`0b0010`** for a PTS-only field. The remaining timestamp bits and
  all three marker bits are valid (`a3 86 cd 04 09` → markers all `1`). Only the prefix's high bit
  is set. This is consistent across **every** PES, video (`0xE0`) and audio (`0xC0`).

- `0b1010` = decimal `10` = `0x0A` — exactly the value in OME's error
  `Invalid PTS_DTS start bits: 10 (0A), expected: 2 (02)`.

OME's `Pes::ParseTimestamp` (`src/projects/modules/containers/mpegts/mpegts_pes.cpp`) rejects on
strict equality:
```cpp
auto prefix = parser->ReadBits<uint8_t>(4);
if (prefix != start_bits)           // start_bits = 0b0010 for PTS-only
{
    logte("Invalid PTS_DTS start bits: %d (%02X), expected: %d (%02X)", ...);
    return false;                   // <-- aborts the whole PES
}
```
By contrast, ffmpeg's `mpegts.c` reads the PTS via `(*buf & 0x0e) << 29 | ...` — it **masks the
prefix bits off entirely** and only consumes the timestamp bits. SRS and srt-live-server are
equally lenient. So those servers ingest Moblin's stream while OME alone rejects it.

(Moblin's prefix is arguably non-conformant and could be fixed upstream too, but since every other
SRT/TS receiver tolerates it, OME being strict here is an interop gap — Postel's law.)

## A/B controls (identical Moblin stream, same SRT ingest path)
| Receiver | Result |
|---|---|
| Moblin → **OME** (stock) | ❌ `Invalid PTS_DTS start bits: 10 (0A)`, never publishes |
| Moblin → **srt-live-server** | ✅ clean — valid 1080p H.264 + AAC, 0 PES/TS errors |
| Moblin → **SRS v6** | ✅ clean |
| **ffmpeg**-generated MPEG-TS → **OME** (same SRT path) | ✅ clean (ffmpeg emits the standard `0b0010` prefix) |

## Fix
Validate only the meaningful low 3 bits of the prefix (which still distinguish PTS-only `0b010`,
PTS-of-pair `0b011`, and DTS `0b001`) and tolerate a stray high bit — matching ffmpeg/SRS/SLS:

```cpp
auto prefix = parser->ReadBits<uint8_t>(4);
// Some encoders (e.g. Moblin/libsrt) set the high bit of the 4-bit PTS/DTS
// prefix, sending 0b1010 instead of the spec's 0b0010. ffmpeg, SRS and
// srt-live-server ignore the prefix entirely and read only the timestamp
// bits; match that leniency by validating only the meaningful low 3 bits so
// these otherwise-valid streams still ingest instead of being rejected.
if ((prefix & 0b0111) != (start_bits & 0b0111))
{
    logte("Invalid PTS_DTS start bits: %d (%02X), expected: %d (%02X)", prefix, prefix, start_bits, start_bits);
    return false;
}
if (prefix != start_bits)
{
    logtw("Tolerating non-standard PTS_DTS start bits: %d (%02X), expected: %d (%02X)", prefix, prefix, start_bits, start_bits);
}
```

## Verification
Built OME `master` with the patch and re-ran the exact repro (Moblin → patched OME, stock config):
- `Invalid PTS_DTS start bits` errors: **0** (was thousands).
- `Tolerating non-standard PTS_DTS start bits` warnings: fires once per PES — the parser now accepts them.
- Stream publishes (`#default#app/test` created) and **plays** (verified via SRT pull and LLHLS).

rostyslavrezn:
Fixes #2215.

## Problem
A Moblin (libsrt) MPEG-TS/SRT publish connects and a stream object is created, but `Pes::ParseTimestamp` continuously logs `Invalid PTS_DTS start bits: 10 (0A), expected: 2 (02)` and the stream never publishes. `RCV-DROPPED = 0` (no SRT loss). The byte-identical stream ingests cleanly in srt-live-server, SRS v6, and via ffmpeg.

## Root cause (byte-confirmed)
A raw SRT capture (strip the 16-byte SRT data header, reassemble the TS) shows the stream is clean, standard, **188-aligned** MPEG-TS — this is **not** a sync-loss / desync problem. The PES headers parse fine; only the PTS field's 4-bit prefix is non-standard. Example first video PES (PID `0x100`, `stream_id 0xE0`):

```
00 00 01 e0 64 e3 84 80 05 a3 86 cd 04 09
└ start  ┘ E0 └len┘ 84 80 05 └──── PTS ────┘
```

`pts_dts_flags = 0b10` (PTS-only), `PES_header_data_length = 5`, and the PTS begins at `0xA3` = `1010 0011`. The first 4 bits (the PTS prefix) are **`0b1010`** instead of the spec-mandated **`0b0010`** — and `0b1010` = `0x0A` = `10`, exactly the value in the error. The timestamp bits and all three marker bits are valid; only the prefix's high bit is set. This is consistent across **every** PES, both video (`0xE0`) and audio (`0xC0`).

ffmpeg's `mpegts.c` reads the PTS via `(*buf & 0x0e) << 29 | ...` — it masks the prefix off entirely and only consumes the timestamp bits. SRS and srt-live-server are equally lenient. OME's strict `prefix != start_bits` is the only one that rejects the stream.

## Fix
Validate only the meaningful low 3 bits of the prefix (which still distinguish PTS-only `0b010`, PTS-of-pair `0b011`, and DTS `0b001`) and warn instead of fail on a stray high bit — matching ffmpeg/SRS/srt-live-server.

## Verification
Built `master` + this patch and re-ran the exact repro (Moblin → patched OME, stock `<SRT/>` config):
- `Invalid PTS_DTS start bits` errors: thousands → **0**.
- The new `Tolerating non-standard PTS_DTS start bits` warning fires once per PES; the parser now accepts them.
- Stream publishes and **plays** (verified via SRT pull and LLHLS).

See #2215 for the full A/B control matrix and capture/parsing details.

### DIFF
--- a/src/modules/containers/mpegts/mpegts_pes.cpp
+++ b/src/modules/containers/mpegts/mpegts_pes.cpp
@@ -1,6 +1,8 @@
 #include "mpegts_pes.h"
 #include "mpegts_private.h"
 
+#include <atomic>
+
 namespace mpegts
 {
 	Pes::Pes(uint16_t pid)
@@ -376,10 +378,29 @@ namespace mpegts
 		// m: marker_bit (should be 1)
 		// T: timestamp
 		auto prefix = parser->ReadBits<uint8_t>(4);
-		if (prefix != start_bits)
+
+		// Some encoders (e.g. Moblin/libsrt) set the high bit of the 4-bit PTS/DTS
+		// prefix, sending 0b1010 instead of the spec's 0b0010. ffmpeg, SRS and
+		// srt-live-server ignore the prefix entirely and read only the timestamp
+		// bits; match that leniency by validating only the meaningful low 3 bits so
+		// these otherwise-valid streams still ingest instead of being rejected.
+		// The low 3 bits still tell PTS-only (0b010), PTS-of-pair (0b011) and
+		// DTS (0b001) apart, and the marker bits below still guard against corruption.
+		if ((prefix & 0b0111) != (start_bits & 0b0111))
 		{
 			logte("Invalid PTS_DTS start bits: %d (%02X), expected: %d (%02X)", prefix, prefix, start_bits, start_bits);
 			return false;
+		}
+
+		if (prefix != start_bits)
+		{
+			// A Pes is built per PES packet, so warn only once per process instead of
+			// ~70-100 times a second for as long as the stream stays connected.
+			static std::atomic<bool> warned{false};
+			if (warned.exchange(true) == false)
+			{
+				logtw("Tolerating non-standard PTS_DTS start bits: %d (%02X), expected: %d (%02X)", prefix, prefix, start_bits, start_bits);
+			}
 		}
 
 		int64_t ts = parser->ReadBits<int64_t>(3) << 30;
--- a/src/modules/containers/mpegts/mpegts_pes.h
+++ b/src/modules/containers/mpegts/mpegts_pes.h
@@ -111,6 +111,9 @@ namespace mpegts
 		}
 
 	private:
+		// Unit-test access to the private timestamp parser (see mpegts_pes_test.cpp)
+		friend class PesTest;
+
 		bool HasOptionalHeader() const;
 		bool HasOptionalData() const;
 		bool ParsePesHeader(BitReader *parser);
--- a/src/modules/containers/mpegts/mpegts_pes_test.cpp
+++ b/src/modules/containers/mpegts/mpegts_pes_test.cpp
@@ -1,0 +1,181 @@
+//==============================================================================
+//
+//  OvenMediaEngine
+//
+//  Created by Rostyslav Reznichenko
+//  Copyright (c) 2026 OvenMediaLabs. All rights reserved.
+//
+//==============================================================================
+#include "mpegts_pes.h"
+
+#include <gtest/gtest.h>
+
+#include <cstdint>
+
+namespace mpegts
+{
+	// A PTS/DTS field is 40 bits laid out as:
+	//
+	//  76543210  76543210  76543210  76543210  76543210
+	// [ssssTTTm][TTTTTTTT][TTTTTTTm][TTTTTTTT][TTTTTTTm]
+	//
+	// s: the 4-bit start_bits prefix, m: marker bit (must be 1), T: the 33-bit timestamp.
+	//
+	// Every vector below shares the tail 86 CD 04 09 taken from a real Moblin/libsrt
+	// capture and varies only the prefix nibble, so each accepted case must decode to
+	// the same value:
+	//
+	//   (1 << 30) | (0x86 << 22) | (0x66 << 15) | (0x04 << 7) | 0x04 == 1639121412
+	//
+	// which proves a stray high bit in the prefix does not perturb the timestamp.
+	constexpr int64_t CAPTURED_TIMESTAMP		 = 1639121412;
+
+	// PTS-only, spec-conformant prefix 0b0010.
+	constexpr uint8_t STANDARD_PTS[]			 = {0x23, 0x86, 0xCD, 0x04, 0x09};
+	// PTS-only, prefix 0b1010 - what Moblin/libsrt actually sends.
+	constexpr uint8_t NON_STANDARD_PTS[]		 = {0xA3, 0x86, 0xCD, 0x04, 0x09};
+	// PTS of a PTS/DTS pair, prefixes 0b0011 and 0b1011.
+	constexpr uint8_t STANDARD_PTS_OF_PAIR[]	 = {0x33, 0x86, 0xCD, 0x04, 0x09};
+	constexpr uint8_t NON_STANDARD_PTS_OF_PAIR[] = {0xB3, 0x86, 0xCD, 0x04, 0x09};
+	// DTS, prefixes 0b0001 and 0b1001.
+	constexpr uint8_t STANDARD_DTS[]			 = {0x13, 0x86, 0xCD, 0x04, 0x09};
+	constexpr uint8_t NON_STANDARD_DTS[]		 = {0x93, 0x86, 0xCD, 0x04, 0x09};
+
+	// Pes::ParseTimestamp is private; this fixture is friended by Pes so the parser can
+	// be driven directly and its boolean result asserted (see the friend declaration in
+	// mpegts_pes.h). Same arrangement as IcePort/IcePortTest in src/modules/ice.
+	class PesTest : public ::testing::Test
+	{
+	protected:
+		static bool Parse(const uint8_t *bytes, size_t length, uint8_t start_bits, int64_t &timestamp)
+		{
+			BitReader reader(bytes, length);
+			Pes pes;
+
+			return pes.ParseTimestamp(&reader, start_bits, timestamp);
+		}
+
+		template <size_t N>
+		static bool Parse(const uint8_t (&bytes)[N], uint8_t start_bits, int64_t &timestamp)
+		{
+			return Parse(bytes, N, start_bits, timestamp);
+		}
+	};
+
+	// Baseline: a spec-conformant PTS-only prefix still parses, and defines the value
+	// every tolerated variant below must match.
+	TEST_F(PesTest, ParsesStandardPtsPrefix)
+	{
+		int64_t timestamp = 0;
+
+		EXPECT_TRUE(Parse(STANDARD_PTS, 0b0010, timestamp));
+		EXPECT_EQ(timestamp, CAPTURED_TIMESTAMP);
+	}
+
+	// The regression this fix is about: Moblin/libsrt sets the high bit of the prefix
+	// (0b1010 instead of 0b0010). ffmpeg, SRS and srt-live-server mask the prefix off and
+	// read only the timestamp bits, so OME must accept it too - and decode the same value
+	// as the conformant encoding above.
+	TEST_F(PesTest, ToleratesStrayHighBitInPtsPrefix)
+	{
+		int64_t timestamp = 0;
+
+		EXPECT_TRUE(Parse(NON_STANDARD_PTS, 0b0010, timestamp));
+		EXPECT_EQ(timestamp, CAPTURED_TIMESTAMP);
+	}
+
+	// The same leniency applies to both halves of a PTS/DTS pair, not just PTS-only.
+	TEST_F(PesTest, ToleratesStrayHighBitInPtsOfPairPrefix)
+	{
+		int64_t timestamp = 0;
+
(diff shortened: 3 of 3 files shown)


rostyslavrezn:
I have read the CLA Document and I hereby sign the CLA

rostyslavrezn:
recheck

src/projects/modules/containers/mpegts/mpegts_pes.cpp
```diff
@@ -376,11 +376,20 @@ namespace mpegts
...
+		// bits; match that leniency by validating only the meaningful low 3 bits so
+		// these otherwise-valid streams still ingest instead of being rejected.
+		if ((prefix & 0b0111) != (start_bits & 0b0111))
 		{
 			logte("Invalid PTS_DTS start bits: %d (%02X), expected: %d (%02X)", prefix, prefix, start_bits, start_bits);
 			return false;
 		}
+		if (prefix != start_bits)
+		{
+			logtw("Tolerating non-standard PTS_DTS start bits: %d (%02X), expected: %d (%02X)", prefix, prefix, start_bits, start_bits);
```
getroot, line 391: `Pes` objects are created per PES packet, so this warning fires for every video and audio frame, roughly 70-100 lines per second for as long as the stream stays connected. Could you make it a one-shot instead? e.g.

```cpp
static bool warned = false;
if (warned == false)
{
	warned = true;
	logtw("Tolerating non-standard PTS_DTS start bits: %d (%02X), expected: %d (%02X)", prefix, prefix, start_bits, start_bits);
}
```

This keeps the operator-visible warning without flooding the log or adding per-packet formatting on the ingest path.

getroot (changes requested):
Thanks for the excellent analysis. The byte-level capture and the cross-checks against ffmpeg/SRS made this easy to verify, and the approach is right: validating only the low 3 bits keeps PTS/DTS discrimination while tolerating the stray high bit, and the marker bit checks still guard against corrupted data.

Two things before we can merge:

1. **Rebase onto current master.** The repository layout changed and this file moved from `src/projects/modules/containers/mpegts/mpegts_pes.cpp` to `src/modules/containers/mpegts/mpegts_pes.cpp`, so the PR is currently conflicting. `ParseTimestamp` itself is unchanged on master, so the patch should transplant cleanly.
2. **Please add a unit test.** We already have mpegts tests nearby. A `ParseTimestamp` case using the bytes from your capture (`A3 86 CD 04 09` accepted, and a low-3-bit mismatch still rejected) would protect this leniency from regressions.

rostyslavrezn:
Thanks for the review — all three points addressed.

**1. Rebased onto current master.** The patch transplanted verbatim; `mpegts_pes.cpp` is byte-identical between the old base and current master, so only the path changed to `src/modules/containers/mpegts/`.

**2. One-shot warning.** Done — good catch on the per-packet cost. One deviation from your snippet: I used `std::atomic<bool>` rather than a plain `static bool`, since MPEG-TS ingest runs on multiple threads and a plain `bool` written from several of them is a data race (benign in practice, but it trips ThreadSanitizer). Same shape and same one-shot behaviour:

```cpp
if (prefix != start_bits)
{
    // A Pes is built per PES packet, so warn only once per process instead of
    // ~70-100 times a second for as long as the stream stays connected.
    static std::atomic<bool> warned{false};
    if (warned.exchange(true) == false)
    {
        logtw("Tolerating non-standard PTS_DTS start bits: %d (%02X), expected: %d (%02X)", prefix, prefix, start_bits, start_bits);
    }
}
```

Happy to switch to the plain `bool` if you'd rather keep it simple.

**3. Unit test.** Added `src/modules/containers/mpegts/mpegts_pes_test.cpp` (picked up automatically by the existing `*_test.cpp` glob, no CMakeLists change). `ParseTimestamp` is private, so the fixture is friended — following the `IcePort`/`IcePortTest` arrangement already in `src/modules/ice`:

```cpp
// Unit-test access to the private timestamp parser (see mpegts_pes_test.cpp)
friend class PesTest;
```

Coverage — every vector shares the tail `86 CD 04 09` from the capture and varies only the prefix nibble, so each accepted case must decode to the same `1639121412`. That is the point: it shows the stray high bit does not perturb the timestamp.

| Case | Expected |
|---|---|
| `23 86 CD 04 09` as PTS-only | accept (spec-conformant control) |
| `A3 86 CD 04 09` as PTS-only | accept — the captured bytes |
| `33/B3 ...` as PTS-of-pair, `13/93 ...` as DTS | accept |
| `13 86 CD 04 09` as PTS-only | reject (`001` ≠ `010`) |
| `33 86 CD 04 09` as PTS-only | reject (`011` ≠ `010`) |
| `B3 86 CD 04 09` as PTS-only | reject — stray high bit *and* wrong low bits |
| marker bit cleared in byte 1 / 3 / 5 | reject |
| truncated field | reject |
| `2F/AF FF FF FF FF` | accept, `0x1FFFFFFFF` |

The `B3`-as-PTS-only case is the one that matters most for regressions: it pins down that masking the high bit did not degrade the check into accepting any prefix.

I also confirmed the test is load-bearing by reverting just the condition to `if (prefix != start_bits)` — exactly the four stray-high-bit cases fail and every rejection case still rejects, so the fix's scope is precisely the prefix nibble and nothing else.

---

Unrelated, but worth flagging while I was here: `cmake --build build/Debug --target tests` fails on a fresh tree with `fatal error: main/git_info.h: No such file or directory`. `ome_git_info` is an `add_custom_target(... ALL)`, so it only runs as part of the default `all` target and nothing declares a dependency on it — building `tests` alone skips the generation. Running `bash src/main/update_git_info.sh` once works around it. Happy to open a separate issue or PR if useful.

getroot:
Thank you for the PR!

I've made a few follow-up changes on top of your commits.

The main goal of this change is to improve flexibility, so I decided it would be better to allow any value in the 4-bit prefix and simply log a warning instead of rejecting it.

Some encoders don't follow the 4-bit prefix specification at all. While our previous implementation strictly validated these bits, I think we should allow not only cases like this one, where a single bit differs, but also cases where all four bits are incorrect.

getroot (approved):

PR #2218, opened 2026-06-23
feat(transcoder): add AV1 software decode and AVIF thumbnails

naanlizard:
## Summary

This adds two related capabilities to the software (FFmpeg) transcoder:

1. **AV1 software decode** — AV1 ingest can now be decoded to raw frames, so AV1 input can drive thumbnails and be transcoded to other codecs. Master already ingests AV1 over enhanced RTMP (#2204), packages it for LL-HLS (#2206), and installs libaom (#2205), but it does not register AV1 in the decoder, so an AV1 source could be packaged but never decoded. This wires AV1 into the existing software video decoder.
2. **AVIF thumbnails** — a fourth still-image thumbnail format alongside JPEG/PNG/WebP, encoded with the libaom AV1 encoder already present in the build and muxed into an AVIF (HEIF/MIAF) container. AVIF is the only one of the four that is colour-managed; JPEG/PNG/WebP are left exactly as they are today.

No new third-party library is introduced. libaom is already a dependency as of #2205. The only build change is correcting the FFmpeg configure component name and enabling the `avif` muxer and `colorspace` filter

## What's new

**AVIF codec plumbing**
- New `cmn::MediaCodecId::Avif` and `cmn::BitstreamFormat::AVIF`, threaded through the codec-id/string helpers, the thumbnail publisher/stream/interceptor, the segment writer, and the RTMP track/chunk handlers so an `<Image><Codec>avif</Codec></Image>` encode is recognised end to end.
- Registered in the transcoder module table under the existing `FFmpeg Image Codec` (DEFAULT) module next to JPEG/PNG/WebP.

**AV1 software decode** (`AVCodecVideoDecoder`)
- `Av1` added to the `FFmpeg Video Codecs` (DEFAULT) decoder module list (was `{H264, H265, Vp8}`).
- An `IsPreFramed()` path for AV1: ingest arrives as complete AV1 temporal units (`AV1_OBU`) and the bundled FFmpeg has no AV1 parser, so these tracks skip the bitstream framer and feed each packet to the decoder as-is. The decoder is selected by name (`libaom-av1`), and the framer-driven resolution-reinit step is skipped because libaom adapts to format changes internally.

**AVIF thumbnail encoder** (folded into `AVCodecImageEncoder`)
- AVIF is an AV1 still muxed into a HEIF/MIAF container. libaom emits only the raw AV1 bitstream, so the encoder drives the `AVCodecContext` directly (rather than the shared `FFmpegCodec` send/receive helpers) to do three things FFmpeg's image path can't do for us:
  - **Deferred open with verbatim CICP** — `avcodec_open2` is deferred to the first frame so the source's colour tags (primaries/transfer/matrix/range) are copied into the AV1 sequence header, making the AVIF CICP match the source. Any still-unspecified primaries/transfer/matrix field defaults to BT.709; an unspecified range defaults to full (JPEG).
  - **`av1C` extradata from an in-band sequence header** — with `AV_CODEC_FLAG_GLOBAL_HEADER` set, the `avif` muxer builds the `av1C` configuration box from the encoder's extradata. libaom can leave extradata empty, so as a fallback the encoder locates the `OBU_SEQUENCE_HEADER` in the first packet and supplies it. Without a populated `av1C`, libavif (Chrome's decoder) rejects the file.
  - **In-memory single-image mux** — each encoded AV1 temporal unit is wrapped into a one-image AVIF via libavformat's `avif` muxer over an in-memory AVIO buffer.
- Encoder settings: `usage=allintra` (every frame an intra still), GOP 0, single-threaded (thumbnails are ~1 fps), `cpu-used=8`, `crf=23`, `bit_rate=0` (CRF mode).
- **Graceful degradation**: if the deferred open fails, the track is marked dead and subsequent frames are dropped rather than erroring, so a failed thumbnail can never stall the stream.

**AVIF-only colour management** (`FilterLavfiRescaler`)
- For an AVIF output track only, the scale step is followed by `colorspace=all=bt709:range=pc:iprimaries=bt709:itrc=bt709`, converting the matrix to BT.709 full range. AVIF carries an explicit CICP tag that browsers honour, so the samples and the tag must agree; JPEG/PNG/WebP do not carry a matrix tag the same way and are intentionally left on the plain bilinear path.
- The `colorspace` filter rejects an unspecified input matrix. Rather than seed the buffersrc with colour options (those need FFmpeg 7.0+; this build is 5.1.4), each untagged frame is stamped before filtering: matrix → BT.709, range → JPEG for `yuvj*` pixel formats else MPEG. A genuinely-tagged BT.601 frame keeps its own tag and is converted; only untagged frames are assumed BT.709, which matches the default browsers apply to untagged content.

**Build** (`cmake/InstallPrerequisites.cmake`)
- FFmpeg configure: `libaom-av1` → `libaom_av1` in `--enable-encoder` and `--enable-decoder`. `libaom_av1` (underscore) is the component name FFmpeg's configure expects; with this the libaom AV1 encoder and decoder are actually compiled into the bundled FFmpeg.
- Added the `avif` muxer and the `colorspace` filter to the configure line.

## Build / dependency impact

- No new external library. libaom (3.12.0) is already installed as of #2205; this PR only makes FFmpeg actually build its libaom AV1 encoder/decoder and enables two more FFmpeg components (`avif` muxer, `colorspace` filter).

## Testing

A linear, self-contained test script is included at `misc/test_av1_avif.sh`. It builds nothing itself; given an image built from this branch it:
- starts OME with a JPEG+PNG+WebP+AVIF thumbnail profile,
- prints the bundled FFmpeg's enabled encoders/decoders/muxers/filters (to confirm `libaom_av1`, `avif`, `colorspace`),
- pushes a CPU H.264 (libx264) stream and fetches all four thumbnails — proving the H.264 decode path and all four image encoders,
- ffprobes the AVIF to confirm its colour tags (`av1` / `yuv420p` / `bt709` / `range=pc`),
- pushes a libaom AV1 stream, confirms OME ingests it as AV1 (`BSF(AV1_OBU)`), and produces thumbnails from it — a thumbnail can only be produced if the AV1 was decoded.

## Commits

1. `feat: AVIF codec id and AV1/AVIF build plumbing`
2. `feat: AV1 software decoder`
3. `feat: AVIF thumbnail encoder in the avcodec image encoder`
4. `feat: colour-managed AVIF thumbnails`
5. `test: AV1 decode + AVIF thumbnail reproduction script`

### DIFF
--- a/cmake/InstallPrerequisites.cmake
+++ b/cmake/InstallPrerequisites.cmake
@@ -566,13 +566,13 @@ set(_FFMPEG_CONFIGURE_CMD
     "${_FFMPEG_ADDI_EXTRA_LIBS}"
     "${_FFMPEG_ADDI_LICENSE}"
     "${_FFMPEG_ADDI_LIBS}"
-    "--enable-encoder=libvpx_vp8,libaom-av1,libopus,libfdk_aac,libopenh264,mjpeg,png,libwebp${_FFMPEG_ADDI_ENCODER}"
-    "--enable-decoder=aac,aac_latm,aac_fixed,mp2,mp2float,mp3float,mp3,h264,hevc,libaom-av1,opus,vp8,mjpeg,png${_FFMPEG_ADDI_DECODER}"
+    "--enable-encoder=libvpx_vp8,libaom_av1,libopus,libfdk_aac,libopenh264,mjpeg,png,libwebp${_FFMPEG_ADDI_ENCODER}"
+    "--enable-decoder=aac,aac_latm,aac_fixed,mp2,mp2float,mp3float,mp3,h264,hevc,libaom_av1,opus,vp8,mjpeg,png${_FFMPEG_ADDI_DECODER}"
     "--enable-parser=aac,aac_latm,aac_fixed,h264,hevc,mpegaudio,opus,vp8,png,jpg"
     "--enable-protocol=tcp,udp,rtp,file,rtmp,tls,rtmps,libsrt"
     "--enable-demuxer=rtsp,flv,live_flv,mp4,mp3,image2"
-    "--enable-muxer=mp4,webm,mpegts,flv,mpjpeg"
-    "--enable-filter=asetnsamples,aresample,aformat,channelmap,channelsplit,scale,transpose,fps,settb,asettb,crop,format${_FFMPEG_ADDI_FILTERS}"
+    "--enable-muxer=mp4,webm,mpegts,flv,mpjpeg,avif"
+    "--enable-filter=asetnsamples,aresample,aformat,channelmap,channelsplit,scale,transpose,fps,settb,asettb,crop,format,colorspace${_FFMPEG_ADDI_FILTERS}"
 )
 list(JOIN _FFMPEG_CONFIGURE_CMD " " _FFMPEG_CONFIGURE_LINE)
 
--- a/misc/test_av1_avif.sh
+++ b/misc/test_av1_avif.sh
@@ -1,0 +1,177 @@
+#!/usr/bin/env bash
+#
+# test_av1_avif.sh
+#
+# Reproduces the AV1-decode + AVIF-thumbnail tests for the OvenMediaEngine
+# branch  feature/av1-decode-avif-thumbnails.
+#
+# It is deliberately linear and verbose. Read it top to bottom: each "STAGE"
+# is one self-contained step that prints its own raw output. The output is
+# ugly on purpose -- the point is that anyone can run this and see what happens.
+#
+# Needs:
+#   - docker
+#   - the built image  ome-av1-avif:dev   (a USE_LOCAL docker build of the branch:
+#         docker build -t ome-av1-avif:dev --build-arg USE_LOCAL=true . )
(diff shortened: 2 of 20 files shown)


naanlizard:
Sort of obvious, but just to be clear - I wouldn't expect the test script to end up on master, it's just included for a quick sanity check that this didn't break anything in a very brute force sort of way - I'm not familiar enough with the test setup y'all use to trust Claude to orchestrate it, but I'm pretty handy with docker and bash. 

So, if everything else looks good, we can strip that file/commit from the PR and only then merge it.

Keukhan:
@naanlizard   Thank you for your PR. 
Our source code is currently in a code freeze period, so we only merge critical bug fixes.
Your code look good overall, i will review it carefully, and if anything needs to be chaged, i will let you know. 
Thanks you again. 

Is it okay if I make some changes and commit them if needed?

naanlizard:
Yes of course

FWIW I’m running our fork in production to test it, including the ffmpeg
update, timing logs, this av1 work, and thumbnail color corrections. So far
so good.

Keukhan:
@naanlizard 
The code changed too much after that PR, so it couldn't be used. I added the AVIF thumbnail feature based on your changes. Please take a look at this PR (#2268)

PR #2221, opened 2026-06-26
fix: preload Whisper model on the configured OME device, not the CUDA index

getroot:
## Problem
Whisper `<PreloadModel><Devices>` was interpreted as CUDA device indices, while `<Modules>nv:N` (video and STT encoders) uses OME device indices. On hosts where CUDA and OME order devices differently (CUDA defaults to fastest-first, OME uses PCI bus order), a single device index preloads the model onto a different GPU than the encoders use, so the preload is wasted and the per-stream STT encoder reloads its own context.

## Solution
Resolve `<Devices>` through the same OME-to-CUDA mapping as `<Modules>` (`TranscodeGPU::GetExternalDeviceId`) so preload and the encoders share one context on the configured device. Non-numeric tokens are rejected with a warning, and a selection that resolves to no device skips the model instead of falling through to load-on-all-GPUs. The config comment and STT docs are updated to match.

## Test
On a host where OME and CUDA enumerate GPUs in different orders (e.g. a mixed-GPU box where the faster card is OME device 1 but CUDA device 0), set `<Video><Modules>nv:1</Modules>` and `<PreloadModel><Devices>1</Devices>`, push a stream with audio, enable STT, then inspect the OME log (`Whisper model loaded successfully on device N`) and `nvidia-smi` per-GPU memory.

Pass conditions:
- The Whisper preload lands on the same physical GPU as the `nv:1` encoders, not the GPU at CUDA index 1.
- A non-numeric `<Devices>` token is skipped with a warning instead of silently mapping to device 0.
- A `<Devices>` value that resolves to no device skips the model with a warning rather than loading on every GPU.

### DIFF
--- a/docs/subtitles/realtime-speech-to-text.md
+++ b/docs/subtitles/realtime-speech-to-text.md
@@ -99,7 +99,7 @@ Each `<PreloadModel>` entry has the following fields:
 | Key | Description |
 |---|---|
 | Path | Path to the model file. Can be absolute or relative to the config directory. |
-| Devices | Comma-separated list of CUDA device indices to load the model onto (e.g. `0`, `0,1`, `2`). Set to `all` to load on every available GPU. If omitted, defaults to device 0. |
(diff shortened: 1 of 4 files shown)


dimiden (approved):
LGTM

PR #2223, opened 2026-06-27
fix: prevent crash in internal-queue stats API under concurrent queue updates

getroot:
## Problem
A read-only `GET /v1/stats/current/internals/queues` crashes the whole OME process with SIGSEGV when polled while streams/queues are being created or destroyed. The monitor thread reassigns `mon::QueueMetrics::_urn` / `_type_name` while the API thread reads them with no synchronization, corrupting the URN shared_ptr refcount (use-after-free).

## Solution
Guard `mon::QueueMetrics` fields with a per-instance `std::shared_mutex`: writers (`UpdateMetadata`/`UpdateMetrics`) take an exclusive lock, getters take a shared lock. `GetUrn()` and `GetTypeName()` now return by value so the reader holds its own reference for the `->ToString()` call.

## Test
Poll `/v1/stats/current/internals/queues` ~1/s while ramping and removing transcoded RTMP streams for several minutes.
Pass condition: no SIGSEGV; the endpoint keeps returning valid JSON throughout.

### DIFF
--- a/src/projects/modules/json_serdes/metrics.cpp
+++ b/src/projects/modules/json_serdes/metrics.cpp
@@ -83,7 +83,10 @@ namespace serdes
 		Json::Value value;
 
 		SetInt64(value, "id", metrics->GetId());
-		SetString(value, "urn", metrics->GetUrn()->ToString(), Optional::False);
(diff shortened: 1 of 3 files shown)


Keukhan (approved):
LGTM

PR #2227, opened 2026-06-30
fix: distribute MediaRouter streams across all workers via round-robin

getroot:
## Problem
With `AppWorkerCount` > 1, MediaRouter assigns each stream to a worker by `stream_id % worker_count`. The global stream-id counter is strided across applications (an origin stream and its transcoded output get interleaved ids), so for an even worker count the modulo lands every stream on only half the workers. The idle half wastes capacity while the active half carries double the load, building queue backlog on the hot workers at high stream counts.

## Solution
Assign each stream a worker round-robin at creation time (under the existing `_streams_lock`), store it immutably on the `MediaRouteStream` (set in the constructor), and route by the stored id instead of recomputing `stream_id % N`. The now-unused `GetWorkerIDByStreamID` is removed.

## Test
- Configure two `<Application>`s in `Server.xml`.
- Create streams sequentially in each app (any ingest method, e.g. RTMP/SRT).
- Observe how each stream is assigned to MediaRouter worker threads.

Pass condition:
- Streams are distributed evenly across all workers. Previously, in this sequential-assignment case, streams landed on only half the workers.

### DIFF
--- a/src/projects/mediarouter/mediarouter_application.cpp
+++ b/src/projects/mediarouter/mediarouter_application.cpp
@@ -415,6 +415,7 @@ std::shared_ptr<MediaRouteStream> MediaRouteApplication::CreateInboundStream(con
 		return nullptr;
 	}
 
+	new_stream->SetWorkerID((_inbound_worker_rr++) % _max_worker_thread_count);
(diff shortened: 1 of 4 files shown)


dimiden (approved):
LGTM

PR #2229, opened 2026-06-30
feat: expose internal-queue dwell-time percentiles in alerts, stats, and logs

getroot:
## Problem
Internal-queue congestion is surfaced only by instantaneous queue size and an EMA of waiting time. Both spike on bursty (jittery) arrival even when items drain in microseconds, so there is no way to tell a brief jitter spike (size touches the threshold but drains immediately) from a real backlog (items sit in the queue for a long time). Separately, the "Exceeded" queue log only prints after a queue has been over threshold for the cumulative log interval, so short bursts are never logged.

## Solution
Record a per-Dequeue dwell-time histogram (microseconds an item waited in the queue, 48 log2 buckets, relaxed atomics) on each ManagedQueue and expose min / avg / p50 / p90 / p99 / max through the congestion alert payload, the queues stats API, and the "Exceeded" log line. Change the log gate so a new peak always logs once (short bursts surface) while sustained congestion still rate-limits to the log interval.

## Test
- Run OME under load high enough to push an internal queue over its threshold (e.g. many streams through a transcoding app).
- Read the queues stats API (internalQueues) and the congestion alert webhook payload.

Pass condition:
- Each congested queue reports dwellMinUs / dwellAvgUs / dwellP50Us / dwellP90Us / dwellP99Us / dwellMaxUs, and the "Exceeded" log line includes the dwell distribution. A short burst that touches the threshold and drains immediately now produces a log line (previously suppressed by the cumulative-interval gate).


Keukhan (approved):
lgtm

PR #2236, opened 2026-07-07
feat: add last-interval dwell-time stats to internal queues

getroot:
Internal-queue dwell-time stats are cumulative over the queue's whole lifetime, so a single early spike pins the max and the high percentiles and they stop reflecting current behavior.

Add a parallel last-interval set that captures only the most recent completed stats interval, rolled over each tick the same way the per-second counters are, and expose it as dwellLatest{Min,Avg,P50,P90,P99,Max}Us in the queue stats API and the queue Exceeded log. The cumulative set is kept unchanged.

Test method: ran OME with a live stream, then polled /v1/stats/current/internals/queues and sampled an active queue once per second.
Pass condition: the queue exposes the dwellLatest* fields, they change every second to track the current interval, and the cumulative dwellMaxUs stays pinned at the lifetime peak.

### DIFF
--- a/src/projects/base/info/managed_queue.h
+++ b/src/projects/base/info/managed_queue.h
@@ -257,7 +257,9 @@ namespace info
 		}
 
 		// Dwell-time distribution: microseconds each item waited in the queue, recorded per
-		// Dequeue into log2 buckets so percentiles stay cheap on the hot path.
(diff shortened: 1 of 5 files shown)


Keukhan (approved):
There is no critical problem. looks good to me.

PR #2243, opened 2026-07-13
feat: add OME_LATENCY_PROBE build option for serving-path latency instrumentation

getroot:
## Problem

Diagnosing where request-serving time goes, or where epoll worker threads stall, currently means hand-patching temporary instrumentation into hot paths and rebuilding. There is no built-in, low-overhead way to trace when a request arrived in the kernel versus when OME actually served it, or to observe worker-loop and pipeline-creation stalls.

## Solution

This adds an opt-in build option `OME_LATENCY_PROBE`, OFF by default so no probe code is compiled and normal builds carry zero runtime cost. When enabled, the serving path and socket workers append time-ordered records to a single `latency_probe.log`: per-request kernel-arrival to serve latency (`LLHLS`, `HTTP`), epoll-wait and worker-iteration stalls (`EPOLL`, `SPW`), receive lateness and send backpressure (`RECV_LATE`, `SEND_STALL`), segment and chunk creation (`SEG`), and transcoder element creation time (`CREATE`). The output directory is selected at runtime with the `OME_LATENCY_PROBE_DIR` environment variable and defaults to `/dev/shm`.

## Test

Built with the option OFF (default): the binary contains no probe symbols and behavior is unchanged. Built with `-DOME_LATENCY_PROBE=ON`: ran the server, issued HTTP requests, and confirmed the one-time startup banner on stderr plus `latency_probe.log` populated with time-ordered records at the directory given by `OME_LATENCY_PROBE_DIR`.

Pass conditions: the OFF build links with zero probe symbols; the ON build writes serving-path and worker-stall records to the single file at the configured location.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,6 +144,9 @@ message(STATUS "  OME_WHISPER_STATIC: ${OME_WHISPER_STATIC}")
 # worker stalls to a single latency_probe.log (directory via the OME_LATENCY_PROBE_DIR
 # environment variable, default /dev/shm).
 if(OME_LATENCY_PROBE)
+    if(NOT CMAKE_SYSTEM_NAME STREQUAL "Linux")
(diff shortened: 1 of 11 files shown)


Keukhan (approved):
No critical. looks good to me.

PR #2244, opened 2026-07-13
feat: enable jemalloc background purge by default, configurable via Server.xml

getroot:
## Problem

jemalloc runs memory decay/purge on the allocating thread by default, so under load that reclamation work adds latency to the serving path. OME had no way to move it off the allocating thread or to tune it.

## Solution

Enable jemalloc's background purge thread at startup so decay/purge runs on a dedicated thread instead of the allocating thread. It is on by default and configurable via `<Modules><Jemalloc><BackgroundPurge>` in Server.xml, and it is effective only in jemalloc builds. This is fork-safe because OME forks only once at daemonize, before allocator initialization runs.

## Test

Built Release (jemalloc enabled) and ran the server. With `<BackgroundPurge>true</BackgroundPurge>` (the default) the startup log shows "Jemalloc background purge thread is enabled"; with `false` it shows "disabled by configuration"; omitting the element keeps it enabled.

Pass conditions: the config value toggles whether the background purge thread is enabled, as reflected in the startup log.

### DIFF
--- a/misc/conf_examples/Edge.xml
+++ b/misc/conf_examples/Edge.xml
@@ -34,6 +34,12 @@
 			<Enable>false</Enable>
 			<MaxClientPeersPerHostPeer>2</MaxClientPeersPerHostPeer>
 		</P2P>
+
+		<!-- jemalloc allocator tuning (effective only in jemalloc builds) -->
+		<Jemalloc>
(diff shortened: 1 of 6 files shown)


Keukhan (approved):
Emergency approve.

PR #2245, opened 2026-07-13
feat(ffmpeg): build filter graph input from generic host planes

Keukhan:
## What

Decouple the FFmpeg filter graph input from the FFmpeg backend so that frames downloaded from a non-FFmpeg backend can still be fed into the graph.

## Changes

- Add backend-agnostic host-plane accessors to the `MediaFrameData` interface:
  - `GetPlaneCount()`, `GetPlaneData()`, `GetStride()`, `GetPixelFormat()`
- Implement them in `FFmpegMediaFrameData` (host frames only — hardware frames report no host planes for consistency across all four accessors).
- In `FFmpegFilterGraph::PushFrame()`:
  - Reuse the existing `AVFrame` when the downloaded host frame is an FFmpeg backend (unchanged behavior).
  - Otherwise, build a temporary `AVFrame` from the generic host planes and free it after `av_buffersrc_write_frame()`.

## Why

This is a forward-looking change to prepare the transcoder for non-FFmpeg backends (NVIDIA / XMA / Netint). The current FFmpeg path is functionally unchanged, so there is no behavior change today.

## Notes

- Memory safety: the temporary `AVFrame` borrows plane pointers from the host holder, which stays alive until after `av_buffersrc_write_frame()` copies the data (KEEP_REF semantics). Since the temporary frame has no `AVBufferRef`, `av_frame_free()` releases only the struct, not the borrowed data — no use-after-free / double-free.
- The generic-host-plane branch is currently unreachable (only the FFmpeg backend exists) and is covered by a comment describing what a future backend must override.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/projects/modules/ffmpeg/ffmpeg_filter_graph.cpp
+++ b/src/projects/modules/ffmpeg/ffmpeg_filter_graph.cpp
@@ -130,12 +130,29 @@ namespace ffmpeg
 				local_frame->width	= media_frame->GetWidth();
 				local_frame->height = media_frame->GetHeight();
 
+				// Allocate owned buffers and copy the host planes into them, so the
+				// frame data outlives host_holder (which is released when this returns).
+				if (::av_frame_get_buffer(local_frame, 0) < 0)
+				{
+					::av_frame_free(&local_frame);
+					return CodecResult::NoMemory;
+				}
+
+				const uint8_t *src_data[AV_NUM_DATA_POINTERS] = { nullptr };
+				int src_linesize[AV_NUM_DATA_POINTERS]		  = { 0 };
+
 				int planes = host_holder->GetPlaneCount();
 				for (int i = 0; i < planes && i < AV_NUM_DATA_POINTERS; i++)
 				{
-					local_frame->data[i]	 = host_holder->GetPlaneData(i);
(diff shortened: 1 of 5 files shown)


Keukhan:
Thanks for the review. Addressed all four comments in the latest commit:

1. **`GetPlaneData()` const-correctness** — now returns `const uint8_t *` (interface + FFmpeg override).
2. **`GetPlaneData()` on hw frames** — now guards on `hw_frames_ctx` and returns `nullptr` for hardware frames.
3. **`GetStride()` on hw frames** — same guard, returns `0` for hardware frames. All four host-plane accessors are now consistent (no host planes for hw frames).
4. **`local_frame` lifetime / use-after-free** — the temporary frame now owns its buffers via `av_frame_get_buffer()` and the host planes are copied in with `av_image_copy()`, so the data no longer depends on `host_holder`'s lifetime. `av_buffersrc_write_frame()` (KEEP_REF) then just adds a reference.

Both changed translation units compile cleanly against FFmpeg 7.1 (libavutil 59.39.100).

Keukhan:
Closing: reworking this change locally. Will reopen a fresh PR later.

PR #2248, opened 2026-07-15
refactor: flatten `src/projects/*` into `src/`

dimiden:
## What

Move every subdirectory under `src/projects/` up one level to `src/` (`api_server`, `base`, `config`, `main`, `mediarouter`, `modules`, `monitoring`, `orchestrator`, `providers`, `publishers`, `third_party`, `transcoder`, `web_console`) and drop the now-empty `projects` layer.

## Why

The extra `projects` directory added a level of nesting without carrying any meaning: every buildable component already lived under it, so it was noise in every path.
Flattening to `src/` shortens paths and makes the layout match the include root.

## How

- `git mv` for all 13 subdirectories, so history follows each file as a rename.
- Merge the old `src/projects/CMakeLists.txt` into `src/CMakeLists.txt` (global compile definitions, the `ome_git_info` target, and the dependency-ordered `add_subdirectory` graph). `CMAKE_CURRENT_SOURCE_DIR` is now `src`, so `main/update_git_info.sh` and every `add_subdirectory` path stay valid.
- Update the include root in `cmake/CompilerOptions.cmake` (`OME_GLOBAL_INCLUDE_DIRS`): `src/projects` -> `src`, `src/projects/third_party/*` -> `src/third_party/*`. Because `#include` directives resolve against this root, no source `#include` lines change.
- Fix all remaining path references: `cmake/README.md` test-target table, root `CMakeLists.txt` comments, `.gitignore`, `README.md`, `CONTRIBUTING.md`, `docs/intro.md`, `docs/p2p-delivery.md`, `src/main/CMakeLists.txt`, and `src/base/ovlibrary/tsa/README.md`.

## Verification

- `git grep` confirms zero remaining `src/projects`/`projects/` path references outside `src/third_party`.
- `cmake -G Ninja` configure succeeds against the new layout; the `ovlibrary` static library compiles and links cleanly, confirming the include root change resolves at build time.
- Cross-checked every documentation and text-file link that points into the source tree (including the TSA guide and the rtc_signalling p2p link); all targets resolve. The GitHub `blob/master/src/...` links 404 until this branch merges to master, which is expected and self-heals on merge.


Keukhan (approved):
It works.

Issue #2250, opened 2026-07-17
Crash (SIGSEGV) in FILE publisher on concurrent :startRecord with the same record id


mbakumenkov:
**Describe the bug**

OME crashes with SIGSEGV (signal 11) inside the FILE publisher when multiple
concurrent `:startRecord` API requests are sent for the same record id.

The crash always happens in an API server thread (`SPAPISvr-t8082`) inside the
FFmpeg muxer, in one of two places:

- `avformat_write_header` - when a recording file is being opened
- `av_write_trailer` - when a recording file is being closed

Before the crash, the log shows the same record id being started several times
within the same millisecond, handled by different API worker threads:

```
[2026-07-17 08:33:21.101] I [SPAPISvr-t8082:100] FILE | file_application.cpp:111 | Recording Started.  id(stream_25-REC0001) stream(stream_25) ... variant_name(720p,pass)
[2026-07-17 08:33:21.102] I [SPAPISvr-t8082:102] FILE | file_application.cpp:111 | Recording Started.  id(stream_25-REC0001) stream(stream_25) ... variant_name(720p,pass)
[2026-07-17 08:33:21.103] I [SPAPISvr-t8082:101] FILE | file_application.cpp:111 | Recording Started.  id(stream_25-REC0001) stream(stream_25) ... variant_name(720p,pass)
```

Stack traces from two different crashes:

```
[2026-07-17 08:33:32.134] C [SPAPISvr-t8082:100] OvenMediaEngine | signals.cpp:239 |
    OME v0.20.5 (v0.20.5) received signal 11 (SIGSEGV), interrupt.
- Stack trace
#0   /lib/x86_64-linux-gnu/libc.so.6            0x700913610520 ? + 0x42520
#1   /opt/ovenmediaengine/lib/libavformat.so.59 0x700914cfd04b av_write_trailer + 0x4b
#2   OvenMediaEngine                            0x5c77f01f29f8 ? + 0x6569f8
#3   OvenMediaEngine                            0x5c77f01f2fea ? + 0x656fea
#4   OvenMediaEngine                            0x5c77efe0d04d ? + 0x27104d
...
```

```
[2026-07-17 08:42:57.127] C [SPAPISvr-t8082:103] OvenMediaEngine | signals.cpp:239 |
    OME v0.20.5 (v0.20.5) received signal 11 (SIGSEGV), interrupt.
- Stack trace
[...]

PR #2252, opened 2026-07-21
fix(file): serialize file writer lifecycle to fix races during concurrent recording control (#2250)

Keukhan:
## Summary
Related Issue : #2250

Fixes a crash in the File publisher where writing packets could race with recording control (Stop / Split / re-Start), causing a use-after-free on the `AVFormatContext`. All context access is now serialized under a single lock, muxer finalization is explicit, and the session control paths are guarded.

## Changes

**ffmpeg::Writer**
- serialize `SendPacket`/`Start`/`AddTrack` under one lock to fix the use-after-free
- move trailer/close into  `ReleaseAVFormatContext()`; 
- fix an `AVDictionary` leak in `Start()`.

**File publisher**
- guard session create/start/stop/split with mutexes.
- skip recording when no tracks are selected

**ov::IntervalGate** (new)
- thread-safe interval gate. replaces `StopWatch`  for the periodic session update.

**Tests**
- add `writer_test.cpp` (15 tests) for the Writer lifecycle

### 1. Unit tests

    cmake -S src -B build -DOME_BUILD_TESTS=ON
    cmake --build build --target ome_test_modules
    ctest --test-dir build -L modules -R FfmpegWriter --output-on-failure

→ 15/15 passed (incl. the SendPacket-vs-Stop concurrency and trailer-flush tests).

### 2. Integration race test (RTMP ingest + concurrent Record API)

Reproduces the real-world race the fix targets: the stream ingest thread.

1. Push a continuous RTMP source
2. While ingesting, start/stop record concurrently (several parallel loops to maximize contention):

       while true; do
         curl -s -X POST -H "Authorization: Basic <token>" \
           http://localhost:8081/v1/vhosts/default/apps/app:startRecord \
           -d '{"id":"rec-1","stream":{"name":"stream"}}'
         curl -s -X POST -H "Authorization: Basic <token>" \
           http://localhost:8081/v1/vhosts/default/apps/app:stopRecord \
           -d '{"id":"rec-1"}'
       done

   Alternatively, drive `startRecord` from an Admission Webhook handler and repeatedly connect/disconnect the RTMP publisher.

3. Also exercise segmentation (`interval`/`schedule`) so `Split()` overlaps with the API calls.

### DIFF
--- a/src/modules/ffmpeg/writer.cpp
+++ b/src/modules/ffmpeg/writer.cpp
@@ -66,6 +66,8 @@ namespace ffmpeg
 		return _error_message;
 	}
 
+	// Interrupt callback for FFmpeg I/O operations
+	// 1 = interrupt, 0 = continue
 	int Writer::InterruptCallback(void *opaque)
 	{
 		Writer *writer = (Writer *)opaque;
@@ -101,7 +103,19 @@ namespace ffmpeg
 		}
 		else if(writer->GetState() == WriterStateClosing)
 		{
-			// In closing state, Ignore interrupt to finish flushing and closing properly.
+			// Nothing opened or written yet: nothing to finalize, so interrupt the pending I/O now.
+			if (writer->_need_to_flush.load() == false && writer->_need_to_close.load() == false)
+			{
+				return 1;
+			}
+
+			// Close timeout
+			int64_t elapsed_ms = std::chrono::duration_cast<std::chrono::milliseconds>(elapsed).count();
+			if(elapsed_ms > writer->GetSendTimeout())
+			{
+				logae(writer, "Close timeout occurred. %" PRId64 " milliseconds. ", elapsed_ms);
+				return 1;
+			}
 		}
 
 		return 0;
@@ -113,6 +127,15 @@ namespace ffmpeg
 		{
 			logae(this, "Destination url is empty");
 			return false;
+		}
+
+		{
+			std::shared_lock<std::shared_mutex> mlock(_av_format_lock);
+			if (_av_format != nullptr)
+			{
+				logae(this, "URL is already set. Create a new Writer instead of reusing this one.");
+				return false;
+			}
 		}
 
 		_url = url;
@@ -149,14 +172,14 @@ namespace ffmpeg
 
 	std::shared_ptr<AVStream> Writer::CreateAVStream(const std::shared_ptr<const MediaTrack> &media_track)
 	{
-		auto av_format = GetAVFormatContext();
-		if (!av_format)
+		// Caller (AddTrack) must hold _av_format_lock
+		if (_av_format == nullptr)
 		{
 			logae(this, "Context is not available");
 			return nullptr;
 		}
 
-		AVStream *new_stream = avformat_new_stream(av_format.get(), nullptr);
+		AVStream *new_stream = avformat_new_stream(_av_format.get(), nullptr);
 		if (!new_stream)
 		{
 			logae(this, "Could not allocate stream");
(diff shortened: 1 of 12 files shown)


src/modules/ffmpeg/writer.cpp
```diff
@@ -101,7 +103,19 @@ namespace ffmpeg
...
-			// In closing state, Ignore interrupt to finish flushing and closing properly.
+			// Nothing opened or written yet: nothing to finalize, so interrupt the pending I/O now.
+			if (writer->_need_to_flush.load() == false && writer->_need_to_close.load() == false)
+			{
+				return 1;
+			}
+
+			// Close timeout
+			int64_t elapsed_ms = std::chrono::duration_cast<std::chrono::milliseconds>(elapsed).count();
+			if(elapsed_ms > writer->GetSendTimeout())
```
getroot: If Stop() is called while SendPacket() is blocked inside av_write_frame(), the sequence is: Stop() sets Closing, this callback interrupts the write after the send timeout, SendPacket() sets Error, and then ReleaseAVFormatContext() runs av_write_trailer()/avio_closep() while the state is Error. There is no Error branch in this callback, so it returns 0 with no timeout at all, and for a network target (the push publisher shares this Writer) a dead peer can hang Stop() indefinitely. Should this get the same timeout branch as Closing, or should ReleaseAVFormatContext() set the state back to Closing under the lock?

A related question on intent: the Closing flush used to ignore interrupts entirely, and now it is bounded by the send timeout (2s default). A slow final flush, for example a large mp4 trailer, can now be cut off and lose the trailer. Was that tradeoff intended? A separate, more generous close timeout might be safer.

src/modules/ffmpeg/writer.cpp
```diff
@@ -309,19 +341,37 @@ namespace ffmpeg
...
 	{
+		// Only a fresh (None) or cleanly stopped (Closed) writer may start; any other state
+		// means a Start() is still in effect and restarting would corrupt the context.
+		const WriterState current_state = GetState();
+		if (current_state == WriterStateConnected ||
+			current_state == WriterStateConnecting ||
+			current_state == WriterStateClosing ||
+			current_state == WriterStateError)
+		{
+			logae(this, "Writer is already started. state:%s", WriterStateToString(current_state));
```
getroot: The state check and SetState(WriterStateConnecting) are not atomic, so two concurrent Start() calls can both pass this guard. Since this PR makes the Writer responsible for its own thread safety, it seems worth moving the check inside _av_format_lock or using compare_exchange on _state.

Also, rejecting WriterStateError here means a writer that failed once can never be restarted, and the log prints "already started" for the Error case. Was that intended? All current callers create a fresh Writer per start, so nothing breaks today, but the message is misleading for Error.

src/modules/ffmpeg/writer.cpp
```diff
@@ -115,6 +129,15 @@ namespace ffmpeg
 			return false;
 		}
 
+		{
+			std::shared_lock<std::shared_mutex> mlock(_av_format_lock);
+			if (_av_format != nullptr)
```
getroot: This check runs under a shared lock, but the _url/_format assignment and SetAVFormatContext() happen after it is released. Two concurrent SetUrl() calls can both pass the null check, and the _url write is unsynchronized against GetUrl() readers. Covering the whole function with one exclusive lock (folding SetAVFormatContext's locking into it) would close both.

src/publishers/file/file_session.cpp
```diff
@@ -49,8 +49,18 @@ namespace pub
...
+
+		// If the session is already started, return true to avoid restarting it.
+		if (GetState() == SessionState::Started)
+		{
+			return true;
+		}
+
 		_is_splitting.store(false);
 
+		_found_first_keyframe = false;
```
getroot, line 62: _found_first_keyframe is written here under _record_control_mutex but read and written in SendOutgoingData() without a lock, so it is still a formal data race (TSan will flag it). std::atomic<bool> would settle it.

Resetting it on restart looks right, by the way. Previously a restarted recording could start mid-GOP.

getroot (commented):
The overall direction looks right. Moving the trailer/close out of the shared_ptr deleter into an explicit ReleaseAVFormatContext() fixes the actual root cause. I left a few questions below about remaining races and two behavior changes.

getroot (approved):
LGTM

PR #2255, opened 2026-07-21
feat(llhls): support runtime track configuration changes with discontinuity signaling

getroot:
## Problem

When a track configuration changed mid-stream (Scheduled Channel item transitions in bypass, WebRTC ingest resolution changes), the LLHLS output kept packaging against the initial configuration without any signaling, so players broke at every change.

## Solution

The LLHLS publisher now cuts an aligned segment boundary on every rendition at each configuration change and signals it per the HLS specification: EXT-X-DISCONTINUITY and EXT-X-DISCONTINUITY-SEQUENCE tags, versioned initialization segments with per-segment EXT-X-MAP switching, a TYPE=MAP preload hint so clients can fetch the upcoming initialization segment ahead of the boundary, and a master playlist CODECS attribute that covers every codec whose segments remain in the playlist window.

## hls.js limitation

hls.js (verified on 1.6.16, still present in current master) fatally errors with "discontinuity sequence mismatch" whenever a discontinuity appears at the live edge of an LL-HLS playlist. Its parser synthesizes the upcoming fragment with the current discontinuity sequence, and its playlist merge treats the sequence bump as a fatal parsing error, although rfc8216bis §4.4.5.3 explicitly allows the next part to start a new discontinuity domain. Until hls.js is patched, hls.js-based players including OvenPlayer can play streams containing discontinuities only with `?_HLS_legacy=YES`. Apple native players (iOS/macOS Safari) and Dolby OptiView play the boundaries seamlessly in LL-HLS mode.

## Test

- Unit: playlist writer tests covering discontinuity tagging, discontinuity sequence accounting, map switching, and preload hints.
- Manual: a Scheduled Channel alternating two bypass items every 10 seconds (640x360 H.264 Baseline with 44.1kHz AAC, and 1920x1080 High with 48kHz AAC, 1s keyframe interval).
- Pass conditions: every item transition produces EXT-X-DISCONTINUITY, a new EXT-X-MAP, and an increasing EXT-X-DISCONTINUITY-SEQUENCE with aligned boundaries on video and audio; the master CODECS attribute lists both codecs while segments of both configurations coexist in the window and returns to a single codec afterwards; iOS Safari and Dolby OptiView play across boundaries without interruption while staying in LL-HLS mode; hls.js plays with `?_HLS_legacy=YES`.

### DIFF
--- a/src/modules/containers/bmff/bmff_packager.cpp
+++ b/src/modules/containers/bmff/bmff_packager.cpp
@@ -36,6 +36,33 @@ namespace bmff
 		return _data_track;
 	}
 
+	bool Packager::UpdateMediaTrack(const std::shared_ptr<const MediaTrack> &media_track)
+	{
+		if (media_track == nullptr || media_track->GetId() != _media_track->GetId())
+		{
+			return false;
+		}
+
+		_media_track = media_track;
+
+		if (_cenc_property.scheme != CencProtectScheme::None && IsCencSupportedCodec(media_track->GetCodecId()) == false)
+		{
+			logte("CENC is not supported for the changed codec(%s), track(%u) will be excluded from CENC protection", cmn::GetCodecIdString(media_track->GetCodecId()), media_track->GetId());
+			_cenc_property.scheme = CencProtectScheme::None;
+		}
+
+		if (_media_track->GetMediaType() == cmn::MediaType::Audio)
+		{
+			_cenc_property.crypt_bytes_block = 0;
+			_cenc_property.skip_bytes_block = 0;
+		}
+
+		// The buffer was flushed at the boundary; rebuild it on the new track
+		_sample_buffer = SampleBuffer(_media_track, _cenc_property);
+
+		return true;
+	}
+
 	bool Packager::WriteFtypBox(ov::ByteStream &container_stream)
 	{
 		// ISO/IEC 14496-12 4.3
--- a/src/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
+++ b/src/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
@@ -100,17 +100,17 @@ namespace bmff
 		}
 
 		// The same codecs CreateInitializationSegment supports; validate before mutating
-		if ((media_track->GetCodecId() == cmn::MediaCodecId::H264) ||
-			(media_track->GetCodecId() == cmn::MediaCodecId::H265) ||
-			(media_track->GetCodecId() == cmn::MediaCodecId::Av1) ||
-			(media_track->GetCodecId() == cmn::MediaCodecId::Aac))
+		switch (media_track->GetCodecId())
 		{
-			// Supported codecs
-		}
-		else
-		{
-			logtw("FMP4Packager::UpdateTrack() - Unsupported codec id(%s)", cmn::GetCodecIdString(media_track->GetCodecId()));
-			return false;
+			case cmn::MediaCodecId::H264:
+			case cmn::MediaCodecId::H265:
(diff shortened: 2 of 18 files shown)


src/publishers/llhls/llhls_stream.cpp
```diff
@@ -1384,7 +1446,90 @@ void LLHlsStream::OnTrackChanged(int32_t track_id, const std::shared_ptr<const M
...
+		logte("LLHlsStream(%s/%s) - Track(%d) has changed to an unsupported codec(%s), the track is excluded from the output from this point", GetApplication()->GetVHostAppName().CStr(), GetName().CStr(), track_id, cmn::GetCodecIdString(new_track->GetCodecId()));
+
+		// Newly served master playlists stop advertising the excluded rendition
+		{
+			std::unique_lock<std::mutex> guard(_master_playlists_lock);
+			_master_playlists.clear();
+		}
+
+		return;
+	}
```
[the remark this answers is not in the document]
getroot: wrong

src/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -90,6 +92,91 @@ namespace bmff
...
+			(media_track->GetCodecId() == cmn::MediaCodecId::Av1) ||
+			(media_track->GetCodecId() == cmn::MediaCodecId::Aac))
+		{
+			// Supported codecs
+		}
+		else
+		{
+			logtw("FMP4Packager::UpdateTrack() - Unsupported codec id(%s)", cmn::GetCodecIdString(media_track->GetCodecId()));
+			return false;
+		}
```
dimiden: (Minor) Just a minor suggestion: how about using a `switch` instead of `if`? That way, it'll be easier to notice when a new codec is added in the future (e.g., compiler warnings for unhandled cases).

src/modules/containers/bmff/bmff_packager.cpp
```diff
@@ -36,6 +36,33 @@ namespace bmff
...
+			return false;
+		}
+
+		_media_track = media_track;
+
+		if (_cenc_property.scheme != CencProtectScheme::None && IsCencSupportedCodec(media_track->GetCodecId()) == false)
+		{
+			logte("CENC is not supported for the changed codec(%s), track(%u) will be excluded from CENC protection", cmn::GetCodecIdString(media_track->GetCodecId()), media_track->GetId());
+			_cenc_property.scheme = CencProtectScheme::None;
+		}
```
dimiden, line 52: On a CENC stream, changing a track to a codec CENC doesn't support (H265/AV1) makes `UpdateMediaTrack()` set the scheme to `None` and keep going, so segments go out in the clear while the playlist still advertises `EXT-X-KEY`.
Better to reject it up front in `OnTrackChanged()`, like the existing `IsSupportedMediaCodec()` guard.

dimiden (approved):
I’ve approved this since it doesn't look like a major issue, but please check the two points I mentioned above and make any necessary updates.

src/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
getroot: applied

src/modules/containers/bmff/bmff_packager.cpp
getroot: Currently, when DRM is not supported for a codec, the output goes out in the clear. So I applied a patch that removes the EXT-X-KEY in that case. If the policy later changes to block segment output when DRM is not supported, I will handle that in a separate PR.

dimiden (approved):

PR #2256, opened 2026-07-22
feat(hls): support runtime track configuration changes with discontinuity signaling

getroot:
## Problem
The TS-based HLS publisher freezes its tracks after start, so a mid-stream track configuration change (codec, resolution, audio format) delivered by a bypass source corrupts the affected rendition while the base publisher only logs a warning.

## Solution
This ports the LLHLS runtime-change support (#2255) to the TS-based HLS publisher. HlsStream::OnTrackChanged repackages the renditions carrying the changed track and requests an aligned discontinuity boundary on every other rendition. The mpegts Packetizer rebuilds the PMT and re-broadcasts the PSI only when the elementary stream type changes, while the Packager cuts a key-frame-aligned boundary and stamps each segment with a configuration generation. HlsMediaPlaylist emits EXT-X-DISCONTINUITY and EXT-X-DISCONTINUITY-SEQUENCE and derives the master CODECS as the union across the retained segments, and the HLS dump refreshes its master so the recorded output follows the change as well.

## Test
A Scheduled channel is the easiest way to exercise this: build a schedule that alternates between source items with different resolutions, so each item switch triggers a runtime track change on the bypass output. Enable the HLS (HLSv3) output and play playlist.m3u8 in OvenPlayer while the channel runs through several switches.

Pass conditions:
- OvenPlayer keeps playing smoothly across each item switch, with no stall, buffering break, or visible corruption, resuming at the new item's resolution.
- The media playlist shows #EXT-X-DISCONTINUITY before the first segment of each new item, and #EXT-X-DISCONTINUITY-SEQUENCE increases as those segments scroll out of the window.
- The master playlist RESOLUTION follows the current item as it changes.

Automated unit tests in the HLS publisher (target ome_test_publishers_hls) also cover the packetizer, packager, and playlist behavior.

### DIFF
--- a/src/modules/containers/mpegts/mpegts_packager.cpp
+++ b/src/modules/containers/mpegts/mpegts_packager.cpp
@@ -190,7 +190,8 @@ namespace mpegts
 			// and PSI (the swap happens after the flush). The new content starts at the
 			// new track's first key frame, which the packetizer key-frame gate ensures.
 			auto main_sample_buffer = GetSampleBuffer(_main_track_id);
-			if (main_sample_buffer != nullptr && main_sample_buffer->GetCurrentDurationMs() > 0.0)
+			bool has_buffered_content = (main_sample_buffer != nullptr && main_sample_buffer->GetCurrentDurationMs() > 0.0);
+			if (has_buffered_content == true)
 			{
 				Flush();
 			}
@@ -203,11 +204,15 @@ namespace mpegts
 			}
 			_codecs_parameter = MakeCodecsParameter();
 
-			// Start a new configuration generation so the next segment is a
-			// discontinuity point. An own cut supersedes a pending propagated one.
-			_track_config_generation++;
-			_pending_cut_timestamp_ms = -1.0;
-			_last_boundary_timestamp_ms = GetLastSampleEndTimestampMs();
+			// A change before any content is produced only establishes the initial
+			// configuration, so it is not a discontinuity. Otherwise start a new
+			// generation; an own cut supersedes a pending propagated one.
+			if (has_buffered_content == true || _last_segment_id > 0)
+			{
+				_track_config_generation++;
+				_pending_cut_timestamp_ms = -1.0;
+				_last_boundary_timestamp_ms = GetLastSampleEndTimestampMs();
+			}
 		}
 		else
 		{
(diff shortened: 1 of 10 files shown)


src/modules/containers/mpegts/mpegts_packetizer.cpp
```diff
@@ -263,11 +339,17 @@ namespace mpegts
...
 		_pmt._current_next_indicator = true;
 		_pmt._section_number = 0;
 		_pmt._last_section_number = 0;
 
+		// Reset per-build fields so the PMT can be rebuilt on a runtime track change.
+		// _version_number is preserved (bumped by UpdateTrack, 0 on the initial build).
+		_pmt._pcr_pid = 0x1FFF;
+		_pmt._program_info_length = 0;
+		_pmt._program_descriptors.clear();
+		_pmt._es_info_list.clear();
```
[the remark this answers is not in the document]
getroot: no problem

src/modules/containers/mpegts/mpegts_packager.cpp
```diff
@@ -9,6 +9,8 @@
 #include "mpegts_packager.h"
 #include "mpegts_private.h"
 
+#include <cmath>
```
Keukhan, line 12: Compile is possible without `include <cmath>`

src/publishers/hls/hls_media_playlist.cpp
```diff
@@ -10,6 +10,8 @@
 #include "hls_media_playlist.h"
 #include "hls_private.h"
 
+#include <algorithm>
```
Keukhan, line 13: Compile is possible without `include <algorithm>`

Keukhan (approved):
I can't find any critical bugs in this changed code. lgtm

src/modules/containers/mpegts/mpegts_packager.cpp
getroot: This include is intentional. Following the Include What You Use principle(IWYU), this file includes <cmath> because it directly uses std::abs, rather than relying on a transitive include that an unrelated header change could silently remove.

src/publishers/hls/hls_media_playlist.cpp
getroot: Same here, following IWYU: this file uses std::find directly, so it includes <algorithm> rather than depending on a transitive include that could disappear when another header changes.

PR #2257, opened 2026-07-22
feat(ovt): relay runtime track configuration changes to edges

getroot:
Problem

A source track can change its configuration at runtime (codec, resolution, decoder configuration). The OVT publisher had no way to tell a connected edge, so the edge kept serving the previous configuration, and for parameters not carried in the media bitstream such as AAC or Opus the edge could never recover. Separately, the WebRTC and SRT publishers logged every such change as a warning that reads like an error even when the change is harmless.

Solution

The OVT origin now relays the changed track's already parsed configuration to connected edges as a signaling message when the track version changes, and the edge applies it before forwarding the packets that follow. The WebRTC and SRT publishers now log a change that keeps the codec as an informational message, since the new parameter sets travel in the media itself, and keep a warning only for a codec change that a running session cannot adopt.

Test

Method: an OVT origin running a scheduled channel that alternates 640x360 AAC 44100 and 1920x1080 AAC 48000 every 10 seconds, an OVT edge pulling it and republishing LLHLS, plus the full_pipeline e2e that routes RTMP through an OVT pull into a transcoder and LLHLS.
Pass: the origin logs the relay on every change, the edge applies each change with zero protocol errors across repeated flips, the AAC sample rate change that is not carried in the bitstream is reflected on the edge, WebRTC and SRT emit the informational log instead of the warning, and the full_pipeline OVT integration stays healthy across ten streams.

### DIFF
--- a/src/modules/ovt_packetizer/ovt_depacketizer.cpp
+++ b/src/modules/ovt_packetizer/ovt_depacketizer.cpp
@@ -91,6 +91,16 @@ bool OvtDepacketizer::IsAvailableMediaPacket()
 	return !_media_packets.empty();
 }
 
+bool OvtDepacketizer::IsAvailable()
+{
+	return !_order.empty();
+}
+
+bool OvtDepacketizer::IsNextMessage()
+{
+	return !_order.empty() && _order.front() == ItemType::Message;
+}
+
 bool OvtDepacketizer::AppendMessagePacket(const std::shared_ptr<OvtPacket> &packet)
 {
 	//TODO(Getroot): Need to validate packet
@@ -108,7 +118,8 @@ bool OvtDepacketizer::AppendMessagePacket(const std::shared_ptr<OvtPacket> &pack
 
 		auto message = _message_buffer.Clone();
 		_messages.push(message);
-		
+		_order.push(ItemType::Message);
+
 		_message_buffer.Clear();
 	}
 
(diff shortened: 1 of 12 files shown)


src/modules/ovt_packetizer/ovt_depacketizer.cpp
```diff
@@ -160,7 +169,7 @@ bool OvtDepacketizer::AppendMediaPacket(const std::shared_ptr<OvtPacket> &packet
 		media_packet->SetFlag(media_flag);
 		media_packet->SetDuration(duration);
 
-		_media_packets.push(media_packet);
+		_items.push(Item{ItemType::MediaPacket, nullptr, media_packet});
```
dimiden: `media_packet` is at its last use here, so moving it into the `Item` avoids one `shared_ptr` copy (an atomic refcount inc/dec pair) per media packet:

```
		_items.push(Item{ItemType::MediaPacket, nullptr, std::move(media_packet)});
```

Micro-optimization only. It does not apply to the message push on line 119 since `_message_buffer.Clone()` already yields an rvalue.

dimiden:
I’ll continue my review once my comments and the issues raised by Copilot are addressed.

src/modules/ovt_packetizer/ovt_depacketizer.cpp
getroot: applied

dimiden (approved):
LGTM

PR #2258, opened 2026-07-23
feat(hevc): add H.265 slice header parser and switch sample entry to hvc1

Keukhan:
### Summary

Adds a slice segment header parser to the H.265 parser and switches the HEVC sample entry from hev1 to hvc1.

### Chnages
- Add H265Parser::ParseSliceHeader() to parse the slice segment header and measure its size; add H265SliceHeader, GetSPS()/GetPPS() accessors, and the SPS/PPS fields it needs.
- Replace `hev1` to `hvc1` in the packaging path and in GetCodecsParameter().

### Testing 
- Publish an HEVC stream to OME using OBS with E-RTMP or SRT.
- Generate an LLHLS stream from OME using h265 encoding (HW required) or bypass options.
- Verify that the stream plays correctly across various browsers and devices.

### Unit Test
```
cmake build/debug && ninja -C build/debug ome_test_modules
./build/debug/bin/ome_test_modules --gtest_filter='H265Parser.*:H265DecoderConfig.*'
```

### DIFF
--- a/src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
+++ b/src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
@@ -579,7 +579,7 @@ bool HEVCDecoderConfigurationRecord::AddVPS(const std::shared_ptr<ov::Data> &nal
 		return false;
 	}
 
-	_vps_map.emplace(vps.GetId(), vps);
+	_vps_map.emplace(vps.GetId(), std::make_shared<H265VPS>(vps));
 	_vps_data_list.push_back(nalu);
 
 	//_num_temporal_layers = std::max<uint8_t>(_num_temporal_layers, vps.GetMaxSubLayersMinus1() + 1);
@@ -627,7 +627,7 @@ bool HEVCDecoderConfigurationRecord::AddSPS(const std::shared_ptr<ov::Data> &nal
 	_constant_frame_rate = 0;
 
 	_sps_data_list.push_back(nalu);
-	_sps_map.emplace(sps.GetId(), sps);
+	_sps_map.emplace(sps.GetId(), std::make_shared<H265SPS>(sps));
 
 	return true;
 }
@@ -647,7 +647,7 @@ bool HEVCDecoderConfigurationRecord::AddPPS(const std::shared_ptr<ov::Data> &nal
 		return false;
 	}
 
-	_pps_map.emplace(pps.GetId(), pps);
+	_pps_map.emplace(pps.GetId(), std::make_shared<H265PPS>(pps));
 	_pps_data_list.push_back(nalu);
 
 	// TODO(Getroot) : Implement PPS parser for getting following values
@@ -665,30 +665,38 @@ bool HEVCDecoderConfigurationRecord::AddPPS(const std::shared_ptr<ov::Data> &nal
 	return true;
 }
 
-bool HEVCDecoderConfigurationRecord::GetSPS(int sps_id, H265SPS &sps) const
+std::shared_ptr<H265SPS> HEVCDecoderConfigurationRecord::GetSPS(int sps_id) const
 {
-	auto it = _sps_map.find(sps_id);
+	// sps_seq_parameter_set_id is in the range [0, 15] (Rec. ITU-T H.265 7.4.3.2.1). 
+	if (sps_id < 0 || sps_id > 15)
+	{
+		return nullptr;
+	}
+
+	auto it = _sps_map.find(static_cast<uint8_t>(sps_id));
 	if (it == _sps_map.end())
 	{
-		return false;
+		return nullptr;
 	}
 
-	sps = it->second;
-
-	return true;
+	return it->second;
 }
 
-bool HEVCDecoderConfigurationRecord::GetPPS(int pps_id, H265PPS &pps) const
+std::shared_ptr<H265PPS> HEVCDecoderConfigurationRecord::GetPPS(int pps_id) const
 {
-	auto it = _pps_map.find(pps_id);
-	if (it == _pps_map.end())
+	// pps_pic_parameter_set_id is in the range [0, 63] (Rec. ITU-T H.265 7.4.3.3.1).
+	if (pps_id < 0 || pps_id > 63)
 	{
-		return false;
+		return nullptr;
 	}
 
-	pps = it->second;
+	auto it = _pps_map.find(static_cast<uint8_t>(pps_id));
+	if (it == _pps_map.end())
+	{
+		return nullptr;
+	}
 
-	return true;
+	return it->second;
 }
 
 std::tuple<std::shared_ptr<ov::Data>, FragmentationHeader> HEVCDecoderConfigurationRecord::GetVpsSpsPpsAsAnnexB()
--- a/src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
+++ b/src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
@@ -60,7 +60,9 @@ ov::String HEVCDecoderConfigurationRecord::GetCodecsParameter() const
 
 	ov::String codecs_parameter;
 
-	codecs_parameter += "hev1";
+	// Use 'hvc1' (parameter sets in the sample entry) rather than 'hev1' (in-band allowed).
+	// This parallels AVC's use of 'avc1' and is required for encrypted HEVC
+	codecs_parameter += "hvc1";
 
 	if (_general_profile_space == 0)
 	{
@@ -659,6 +661,32 @@ bool HEVCDecoderConfigurationRecord::AddPPS(const std::shared_ptr<ov::Data> &nal
 	// else 
 	// 		parallelism_type = 1 // slice-based parallel decoding
 	_parallelism_type = 0;
+
+	return true;
+}
+
+bool HEVCDecoderConfigurationRecord::GetSPS(int sps_id, H265SPS &sps) const
+{
+	auto it = _sps_map.find(sps_id);
+	if (it == _sps_map.end())
+	{
+		return false;
+	}
+
+	sps = it->second;
+
+	return true;
+}
+
+bool HEVCDecoderConfigurationRecord::GetPPS(int pps_id, H265PPS &pps) const
+{
+	auto it = _pps_map.find(pps_id);
+	if (it == _pps_map.end())
+	{
+		return false;
+	}
+
+	pps = it->second;
 
 	return true;
 }
--- a/src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
+++ b/src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
@@ -470,9 +470,11 @@ void HEVCDecoderConfigurationRecord::RebuildNalUnitArray(H265NALUnitType nal_typ
 		v.push_back(nalu);
 	}
 
-	// The Annex B form is built from these lists and cached, so it no longer matches.
+	// Both cached forms are built from these lists, so neither matches any more: the Annex B
+	// bitstream here and the serialized record behind GetData() in the base class.
 	_vps_sps_pps_annexb_data = nullptr;
 	_vps_sps_pps_annexb_frag_header = FragmentationHeader();
+	UpdateData();
 }
 
 void HEVCDecoderConfigurationRecord::AddNalUnit(H265NALUnitType nal_type, const std::shared_ptr<ov::Data> &nal_unit)
--- a/src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
+++ b/src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
@@ -460,24 +460,42 @@ std::shared_ptr<const ov::Data> HEVCDecoderConfigurationRecord::Serialize()
 	return bit.GetDataObject();
 }
 
-void HEVCDecoderConfigurationRecord::AddNalUnit(H265NALUnitType nal_type, const std::shared_ptr<ov::Data> &nal_unit)
+void HEVCDecoderConfigurationRecord::RebuildNalUnitArray(H265NALUnitType nal_type, const std::map<uint8_t, std::shared_ptr<ov::Data>> &data_list)
 {
 	auto &v = _nal_units[static_cast<uint8_t>(nal_type)];
-	v.push_back(nal_unit);
-	
+	v.clear();
+	v.reserve(data_list.size());
+	for (const auto &[id, nalu] : data_list)
+	{
+		v.push_back(nalu);
+	}
+
+	// The Annex B form is built from these lists and cached, so it no longer matches.
+	_vps_sps_pps_annexb_data = nullptr;
+	_vps_sps_pps_annexb_frag_header = FragmentationHeader();
+}
+
+void HEVCDecoderConfigurationRecord::AddNalUnit(H265NALUnitType nal_type, const std::shared_ptr<ov::Data> &nal_unit)
+{
+	// Parameter sets are keyed by id, so Add*() maintains _nal_units itself.
 	if (nal_type == H265NALUnitType::VPS)
 	{
 		AddVPS(nal_unit);
+		return;
 	}
-	else if (nal_type == H265NALUnitType::SPS)
+	if (nal_type == H265NALUnitType::SPS)
 	{
 		// Set Info from SPS
 		AddSPS(nal_unit);
+		return;
 	}
-	else if (nal_type == H265NALUnitType::PPS)
+	if (nal_type == H265NALUnitType::PPS)
 	{
 		AddPPS(nal_unit);
+		return;
 	}
+
+	_nal_units[static_cast<uint8_t>(nal_type)].push_back(nal_unit);
 }
 
 uint8_t HEVCDecoderConfigurationRecord::Version()
@@ -574,13 +592,9 @@ bool HEVCDecoderConfigurationRecord::AddVPS(const std::shared_ptr<ov::Data> &nal
 		return false;
 	}
 
-	if (_vps_map.find(vps.GetId()) != _vps_map.end())
-	{
-		return false;
-	}
-
-	_vps_map.emplace(vps.GetId(), std::make_shared<H265VPS>(vps));
-	_vps_data_list.push_back(nalu);
+	_vps_map[vps.GetId()] = std::make_shared<H265VPS>(vps);
+	_vps_data_list[vps.GetId()] = nalu;
+	RebuildNalUnitArray(H265NALUnitType::VPS, _vps_data_list);
 
 	//_num_temporal_layers = std::max<uint8_t>(_num_temporal_layers, vps.GetMaxSubLayersMinus1() + 1);
 
@@ -593,11 +607,6 @@ bool HEVCDecoderConfigurationRecord::AddSPS(const std::shared_ptr<ov::Data> &nal
 	if (H265Parser::ParseSPS(nalu->GetDataAs<uint8_t>(), nalu->GetLength(), sps) == false)
 	{
 		logte("Could not parse H265 SPS unit");
-		return false;
-	}
-
-	if (_sps_map.find(sps.GetId()) != _sps_map.end())
-	{
 		return false;
 	}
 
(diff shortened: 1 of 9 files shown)


src/modules/bitstream/h265/h265_parser.cpp
```diff
@@ -1265,6 +1305,492 @@ bool H265Parser::ParsePPS(const uint8_t *nalu, size_t length, H265PPS &pps)
...
+				current_strps_valid = true;
+			}
+			else if (sps._num_short_term_ref_pic_sets > 1)
+			{
+				uint32_t H265_READ_BITS(short_term_ref_pic_set_idx, CeilLog2(sps._num_short_term_ref_pic_sets));
+				if (short_term_ref_pic_set_idx < sps._short_term_ref_pic_sets.size())
+				{
+					current_strps = sps._short_term_ref_pic_sets[short_term_ref_pic_set_idx];
+					current_strps_valid = true;
+				}
```
Keukhan: **Out-of-range `short_term_ref_pic_set_idx` should fail instead of silently continuing**

`short_term_ref_pic_set_idx` is read with `Ceil(Log2(num_short_term_ref_pic_sets))` bits, so when `num_short_term_ref_pic_sets` is not a power of two (e.g. `3` → 2 bits, allowing values `0..3` while the valid range is `0..2`), an out-of-range index is representable. In that case we skip the assignment and leave `current_strps_valid == false`, which under-counts `NumPicTotalCurr`. If `lists_modification_present_flag` is set and the real `NumPicTotalCurr > 1`, `ref_pic_lists_modification()` is then mis-parsed and the computed header size becomes wrong.

Every other unsupported/unexpected case in this function returns `false` as a fail-safe — this branch should be consistent:

```suggestion
				uint32_t H265_READ_BITS(short_term_ref_pic_set_idx, CeilLog2(sps._num_short_term_ref_pic_sets));
				if (short_term_ref_pic_set_idx >= sps._short_term_ref_pic_sets.size())
				{
					return false;
				}
				current_strps = sps._short_term_ref_pic_sets[short_term_ref_pic_set_idx];
				current_strps_valid = true;
```

src/modules/bitstream/h265/h265_parser.h
```diff
@@ -212,10 +337,84 @@ struct H265PPS
...
+		return _header_size_in_bits;
+	}
+
+	size_t GetHeaderSizeInBytes() const
+	{
+		return ((_header_size_in_bits + 7) / 8);
+	}
+
+private:
+	uint32_t _slice_type = 2;
```
Keukhan, line 348: **`_slice_type` stays `I` (default) for dependent slice segments**

When `dependent_slice_segment_flag == 1`, the block that reads `slice_type` is skipped, so `_slice_type` keeps its default of `2` (I-slice). This is harmless for the header-size use case (the field isn't needed there), but `GetSliceType()` will then report `ISlice` for what may actually be a P/B dependent slice — a silent trap for any future caller that trusts `GetSliceType()`.

Suggest making the "not parsed" state explicit, or at least documenting the limitation:

```cpp
enum class SliceType : uint8_t { BSlice = 0, PSlice = 1, ISlice = 2, Unknown = 0xFF };
...
uint32_t _slice_type = static_cast<uint32_t>(SliceType::Unknown);
```

(Alternatively, add a comment noting that `_slice_type` / `GetSliceType()` is only valid for independent slice segments.)

Keukhan: Not issue for now

getroot:
For the Test section, please write a slightly more detailed guide so that the QA manager can test this feature.

getroot (commented):
I reviewed the parser against Rec. ITU-T H.265. The slice header walk itself follows 7.3.6.1 correctly. I checked the flag placements, RPS selection, the long term ref loop, collocated_ref_idx, pred_weight_table, the deblocking inference from 7.4.7.1 and byte_alignment. The range/SCC extension fail safe is also the right call. A few things need to be addressed before merge.

### 1. Off-by-one in ProcessShortTermRefPicSet()

7.3.7 defines the loop as inclusive: `for (j = 0; j <= NumDeltaPocs[RefRpsIdx]; j++)`. The code loops `i < num_delta_pocs`, so it under-reads at least one bit. The resize just above already uses `num_delta_pocs + 1`, only the loop bound is wrong. Until now this only misaligned SPS parsing for streams using inter-predicted RPS. With this PR it becomes critical: a slice carrying an inline inter-predicted RPS (short_term_ref_pic_set_sps_flag == 0 with the inter flag set) gets a wrong header size, so the CENC clear/protected boundary lands in the middle of the slice header. This PR already fixes the adjacent delta_idx_minus1 condition, so please fix this loop bound together.

### 2. HEVC + DRM in OSS now produces full-sample encrypted video

Before this PR an H265 track with CENC configured was excluded by IsCencSupportedCodec() and produced in the clear. Now the track goes through the encryptor, and since GenerateSubSamplesFromH265() returns an empty subsample list, every HEVC sample is full-sample encrypted. CDMs such as Widevine require clear slice headers for video, so this output will most likely not play. It also leaves ParseSliceHeader() with no caller in this repo. Please wire GenerateSubSamplesFromH265() to ParseSliceHeader() the same way GenerateSubSamplesFromH264() works. If that is out of scope for this PR, keep H265 out of IsCencSupportedCodec() until it is wired. Either way, the comment in llhls_stream.cpp that lists H265 as a codec CENC cannot encrypt needs to be updated to match the new behavior.

### 3. Please add unit tests for the slice header parser

The value of this parser is exact bit accounting, and the playback test in the description does not execute it because the subsample path is stubbed. Please add unit tests that assert GetHeaderSizeInBytes() against reference values for a few known slice headers: I/P/B slices, SAO on and off, tiles or WPP for the entry point path, long term refs, and an inline inter-predicted RPS. The last case is the one that catches item 1.

### Smaller items

- slice_type is not validated. A malformed value above 2 silently continues with the I slice layout and returns a wrong size. Returning false is safer.
- GetSPS()/GetPPS() take int while the maps are keyed by uint8_t, so an id above 255 wraps and can match the wrong parameter set. Validating the spec ranges (SPS id up to 15, PPS id up to 63) would close this. They also copy the whole struct per slice on a per-frame path. Returning a const pointer avoids the copies.
- The comment above `_header_size_in_bits` says it gives the byte length of [NAL header + slice header], but the code subtracts the NAL header (same semantics as the H264 parser, where the caller adds it back). Please fix the comment so a future consumer does not double count.
- The new guard `delta_idx_minus1 + 1 > idx` wraps when delta_idx_minus1 is UINT32_MAX. `delta_idx_minus1 >= idx` is equivalent without the overflow.
- The logte/logtw calls in the per-slice path will spam at frame rate when triggered, for example by a source that updates PPS in-band with an id that is not in the hvcC record. Consider logtd there.
- The syntax tables added to h265_parser.h duplicate the spec. A one line pointer to the section number is enough.
- Trailing whitespace after the encv return in WriteHvc1Box().

The hev1 to hvc1 switch is the right direction. Apple requires hvc1 and Safari does not play hev1, and the sample entry, the ftyp brand, and the CODECS string all change consistently. The delta_idx_minus1 condition fix and the new bounds check are good catches as well.

src/modules/bitstream/h265/h265_parser.cpp
```diff
@@ -1928,7 +2458,15 @@ bool H265Parser::ProcessShortTermRefPicSet(uint32_t idx, uint32_t num_short_term
...
+		// malformed delta_idx_minus1 (>= idx) would underflow and index out of bounds; reject it.
 		uint32_t ref_rps_idx = idx - (rpset.delta_idx_minus1 + 1);
+		if (rpset.delta_idx_minus1 + 1 > idx || ref_rps_idx >= rpset_list.size())
+		{
+			return false;
+		}
+
 		uint32_t num_delta_pocs = 0;
 
 		if (rpset_list[ref_rps_idx].inter_ref_pic_set_prediction_flag)
```
getroot, line 2592: The flag loop below (`for (uint32_t i = 0; i < num_delta_pocs; i++)`) has an off-by-one. 7.3.7 defines it as inclusive: `for (j = 0; j <= NumDeltaPocs[RefRpsIdx]; j++)`, and the resize above the loop already uses `num_delta_pocs + 1`, only the loop bound is wrong. Until now this only misaligned SPS parsing for streams using inter-predicted RPS. With this PR it becomes critical: a slice carrying an inline inter-predicted RPS (short_term_ref_pic_set_sps_flag == 0 with the inter flag set) gets a wrong header size, so the CENC clear/protected boundary lands in the middle of the slice header. Since this PR already fixes the adjacent delta_idx_minus1 condition, please fix this loop bound together.

src/modules/bitstream/h265/h265_parser.cpp
```diff
@@ -1928,7 +2458,15 @@ bool H265Parser::ProcessShortTermRefPicSet(uint32_t idx, uint32_t num_short_term
 			return false;
 		}
 
+		// An inter-predicted RPS must reference a strictly earlier, already-parsed set
+		// (Rec. ITU-T H.265 7.4.8). ref_rps_idx is computed with unsigned arithmetic, so a
+		// malformed delta_idx_minus1 (>= idx) would underflow and index out of bounds; reject it.
 		uint32_t ref_rps_idx = idx - (rpset.delta_idx_minus1 + 1);
+		if (rpset.delta_idx_minus1 + 1 > idx || ref_rps_idx >= rpset_list.size())
```
getroot: `delta_idx_minus1 + 1` wraps to 0 when delta_idx_minus1 is UINT32_MAX, so the guard passes with ref_rps_idx == idx. `delta_idx_minus1 >= idx` is equivalent without the overflow.

src/modules/containers/bmff/cenc.cpp
```diff
@@ -233,6 +232,12 @@ namespace bmff
 		return true;
 	}
 
+	bool Encryptor::GenerateSubSamplesFromH265(const std::shared_ptr<const MediaPacket> &media_packet, std::vector<Sample::SubSample> &sub_samples)
+	{
+		// HEVC subsample encryption is not supported in the OSS version.
```
getroot: With H265 added to IsCencSupportedCodec(), this stub changes the OSS behavior. Before this PR an H265 track with CENC configured was excluded and produced in the clear. Now it goes through the encryptor with an empty subsample list, so every HEVC sample is full-sample encrypted. CDMs such as Widevine require clear slice headers for video, so this output will most likely not play, and it leaves ParseSliceHeader() with no caller in this repo. Please wire this to ParseSliceHeader() the same way GenerateSubSamplesFromH264() works. If that is out of scope for this PR, keep H265 out of IsCencSupportedCodec() until it is wired. Either way, the comment in llhls_stream.cpp that lists H265 as a codec CENC cannot encrypt needs to be updated to match.

src/modules/bitstream/h265/h265_parser.cpp
```diff
@@ -1265,6 +1305,494 @@ bool H265Parser::ParsePPS(const uint8_t *nalu, size_t length, H265PPS &pps)
...
+
+	if (dependent_slice_segment_flag == 0)
+	{
+		for (uint8_t i = 0; i < pps.num_extra_slice_header_bits; i++)
+		{
+			uint8_t H265_READ_BITS(slice_reserved_flag, 1);
+			(void)slice_reserved_flag;
+		}
+
+		uint32_t H265_READ_UEV(slice_type);
```
getroot, line 1553: slice_type is not validated. A malformed value above 2 silently continues with the I slice layout and returns a wrong size. Returning false is safer.

src/modules/bitstream/h265/h265_decoder_configuration_record.h
```diff
@@ -90,6 +90,10 @@ class HEVCDecoderConfigurationRecord : public DecoderConfigurationRecord
 	// Get NAL units by NAL unit type
 	std::vector<std::shared_ptr<ov::Data>> GetNalUnits(H265NALUnitType nal_type);
 
+	// Get parsed SPS/PPS by id (used e.g. by the slice header parser for CENC)
+	bool GetSPS(int sps_id, H265SPS &sps) const;
```
getroot: These take int while the maps are keyed by uint8_t, so an id above 255 wraps and can match the wrong parameter set. Validating the spec ranges (SPS id up to 15, PPS id up to 63) would close this. They also copy the whole struct per slice on a per-frame path. Returning a const pointer avoids the copies.

src/modules/bitstream/h265/h265_parser.cpp
```diff
@@ -1265,6 +1305,494 @@ bool H265Parser::ParsePPS(const uint8_t *nalu, size_t length, H265PPS &pps)
...
+			(void)slice_segment_header_extension_data_byte;
+		}
+	}
+
+	// byte_alignment()
+	uint8_t H265_READ_BITS(alignment_bit_equal_to_one, 1);
+	(void)alignment_bit_equal_to_one;
+
+	// The remaining alignment_bit_equal_to_zero bits pad to the next byte boundary.
+	// GetHeaderSizeInBytes() rounds up, giving the raw byte length of [NAL header + slice header].
```
getroot: This comment does not match the code. The stored value subtracts the NAL header, so GetHeaderSizeInBytes() returns the slice header bytes only, and the caller adds the NAL header back (same semantics as the H264 parser). Please fix the wording so a future consumer does not double count.

src/modules/bitstream/h265/h265_parser.cpp
```diff
@@ -1265,6 +1305,494 @@ bool H265Parser::ParsePPS(const uint8_t *nalu, size_t length, H265PPS &pps)
...
+		uint8_t H265_READ_BITS(no_output_of_prior_pics_flag, 1);
+		(void)no_output_of_prior_pics_flag;
+	}
+
+	uint32_t H265_READ_UEV(slice_pic_parameter_set_id);
+
+	H265PPS pps;
+	if (hvcc->GetPPS(slice_pic_parameter_set_id, pps) == false)
+	{
+		logte("H265 slice header: PPS(%u) not found", slice_pic_parameter_set_id);
```
getroot: This runs per slice, so when it triggers it will spam at frame rate. A source that updates PPS in-band with an id that is not in the hvcC record would hit this on every frame. Consider logtd here and for the logtw fail-safes below.

src/modules/bitstream/h265/h265_parser.h
```diff
@@ -106,6 +118,44 @@ struct H265VPS
 	uint8_t vps_video_parameter_set_id;
 };
 
+// Sequence Parameter Set (Rec. ITU-T H.265, 7.3.2.2)
+// Sequence-level parameters shared by every picture that references this SPS.
+// Variable-length bitstream: u(n)=n bits, ue(v)=Exp-Golomb unsigned, se(v)=signed. '*' = stored by OME.
```
getroot: These syntax tables duplicate the spec document. A one line pointer to the section number is enough. Same for the PPS and slice header tables below.

src/modules/containers/bmff/bmff_packager.cpp
```diff
@@ -913,7 +913,22 @@ namespace bmff
 			return false;
 		}
 
-		return WriteBox(container_stream, "hev1", *stream.GetData());
+		// CENC
+		if (_cenc_property.scheme != CencProtectScheme::None)
+		{
+			// Encrypted HEVC must use the 'hvc1' sample entry (parameter sets in the sample
+			// entry), matching the 'hvc1' codec string advertised by
+			// HEVCDecoderConfigurationRecord::GetCodecsParameter(). Browser EME / Widevine
```
getroot: This sentence is cut off after 'Browser EME / Widevine'. There is also a tab-only line left below the closing brace of this block.

getroot (commented):
Overall the slice header walk follows Rec. ITU-T H.265 7.3.6.1 correctly. I checked the flag placements, RPS selection, the long term ref loop, collocated_ref_idx, pred_weight_table, the deblocking inference from 7.4.7.1 and byte_alignment. The range/SCC extension fail safe is the right call, the delta_idx_minus1 condition fix is a good catch, and the hev1 to hvc1 switch is the right direction (Apple requires hvc1 and Safari does not play hev1). Details are in the inline comments.

One request beyond the inline comments: please add unit tests for the slice header parser. Its value is exact bit accounting, and the playback test in the description does not execute it because the subsample path is stubbed. Tests should assert GetHeaderSizeInBytes() against reference values for known slice headers: I/P/B slices, SAO on and off, tiles or WPP for the entry point path, long term refs, and an inline inter-predicted RPS (the case that catches the off-by-one I flagged inline).

src/modules/bitstream/h265/h265_parser.h
Keukhan: Changed

src/modules/bitstream/h265/h265_parser.cpp
Keukhan: Changed

src/modules/bitstream/h265/h265_parser.cpp
Keukhan: Fixed comment

src/modules/bitstream/h265/h265_decoder_configuration_record.h
Keukhan: Changed to a shared pointer

src/modules/bitstream/h265/h265_parser.cpp
Keukhan: Added validation

src/modules/containers/bmff/cenc.cpp
Keukhan: Excluded the hevc codec supported by IsCencSupportedCodec and changed the log and return values ​​here.

src/modules/bitstream/h265/h265_parser.cpp
Keukhan: Fixed

src/modules/bitstream/h265/h265_parser.cpp
Keukhan: Added validation to prevent underflow

src/modules/bitstream/h265/h265_parser.cpp
```diff
@@ -1265,6 +1305,509 @@ bool H265Parser::ParsePPS(const uint8_t *nalu, size_t length, H265PPS &pps)
...
+		for (uint32_t i = 0; i < slice_segment_header_extension_length; i++)
+		{
+			uint8_t H265_READ_BITS(slice_segment_header_extension_data_byte, 8);
+			(void)slice_segment_header_extension_data_byte;
+		}
+	}
+
+	// byte_alignment()
+	uint8_t H265_READ_BITS(alignment_bit_equal_to_one, 1);
+	(void)alignment_bit_equal_to_one;
```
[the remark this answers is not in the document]
Keukhan: I think this is not important. lets ignore it. 👍

src/modules/bitstream/h265/h265_decoder_configuration_record.cpp
```diff
@@ -458,24 +460,42 @@ std::shared_ptr<const ov::Data> HEVCDecoderConfigurationRecord::Serialize()
...
+	v.clear();
+	v.reserve(data_list.size());
+	for (const auto &[id, nalu] : data_list)
+	{
+		v.push_back(nalu);
+	}
+
+	// The Annex B form is built from these lists and cached, so it no longer matches.
+	_vps_sps_pps_annexb_data = nullptr;
+	_vps_sps_pps_annexb_frag_header = FragmentationHeader();
```
getroot, line 476: There is a second cache that also needs clearing here. `DecoderConfigurationRecord::GetData()` keeps the serialized record in `_data` and only re-serializes when `_updated` is set:

```cpp
std::shared_ptr<const ov::Data> GetData()
{
    if (_updated) { _data = Serialize(); _updated = false; }
    return _data;
}
```

There is no `UpdateData()` call anywhere in this file, so once `GetData()` has been called the cached blob is never rebuilt. Now that a parameter set with a reused id is an update rather than a rejection, the parsed state and the serialized state can diverge for the same id: `GetSPS()` and the slice header parser see the new SPS while `WriteHvccBox()` keeps writing the old hvcC bytes into the init segment.

Today this is latent rather than live, since `AddVPS`/`AddSPS`/`AddPPS` are private and `AddNalUnit()` is only reached from `Parse()`/`ParseV2Internal()`, which run before the first `GetData()`. But `AddNalUnit()` is public and this PR is what makes these methods mutating, so I would close it now. Calling `UpdateData()` alongside the Annex B invalidation should cover it.

Worth a test in the same shape as `AnnexBCacheFollowsSpsUpdate`: call `GetData()`, add an SPS that reuses the id with a different resolution, then assert `GetData()` changed.

getroot (approved):
The hardening and the test suite look good. I re-checked the new spec bounds against Rec. ITU-T H.265 and they all match, the off-by-one fix is right, and the inline inter-predicted RPS test really does discriminate the bug it guards. Nice catch on the CeilLog2 shift too.

One thing left, in the inline comment: the parameter set update clears the Annex B cache but not the serialized hvcC cache in the base class. A one line fix. Once that is in I think this is good to go.

PR #2261, opened 2026-07-25
feat(llhls): support runtime DRM key rotation

getroot:
A DRM protected LLHLS stream keeps one content key for its whole lifetime, so a key that leaks stays useful for as long as the stream runs, and a stream whose audience becomes restricted cannot make what it already handed out useless.

The DRM info file now takes an ordered list of content keys and an optional rotation period, and a stream moves to the next key from the next segment of each track, cycling through the list for as long as it runs. Each content version carries its own initialization segment and EXT-X-KEY, so segments already delivered stay playable with the key they were encrypted with, and no discontinuity is signaled because the content does not change. A version that a codec change leaves unencrypted now ends the key scope with METHOD=NONE, where before it stayed covered by the preceding key.

The DRM section of the LLHLS manual is restructured along the way, and its claim that only cbcs is supported is corrected.

Test:

Key rotation can be exercised without a DRM license server by using HLS SAMPLE-AES with `KEYFORMAT="identity"`, where the key URI returns the raw 16 byte key. Safari plays this natively. hls.js does not support SAMPLE-AES, so use Safari.

**1. Serve the keys.** Each key needs its own URL returning exactly the 16 raw bytes of the matching `<Key>`. Save the script below as `keyserver.py`.

The URL has to use HTTPS when the playlist is served over HTTPS, otherwise Safari blocks the key request as mixed content. Reuse the certificate that the LLHLS TLS port already serves, so that the host name in `<FairPlayKeyUrl>` matches a name in that certificate. In a default installation the certificate, its chain and its private key are the `<CertPath>`, `<ChainCertPath>` and `<KeyPath>` of the virtual host, and the first two are concatenated into the full chain the script loads.

```bash
cat certificate.crt certificate.ca-bundle > fullchain.pem
cp certificate.key privkey.pem
python3 keyserver.py
```

```python
import binascii, ssl
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

KEYS = {
    "/key/1": "16cf4232a86364b519e1982a27d90087",
    "/key/2": "0f1e2d3c4b5a69788796a5b4c3d2e1f0",
    "/key/3": "00112233445566778899aabbccddeeff",
}

class Handler(BaseHTTPRequestHandler):
    def do_GET(self):
        key_hex = KEYS.get(self.path)
        if key_hex is None:
            self.send_response(404); self.end_headers(); return
        body = binascii.unhexlify(key_hex)
        self.send_response(200)
        self.send_header("Content-Type", "application/octet-stream")
        self.send_header("Content-Length", str(len(body)))
        self.send_header("Access-Control-Allow-Origin", "*")
        self.end_headers()
        self.wfile.write(body)

httpd = ThreadingHTTPServer(("0.0.0.0", 18080), Handler)
ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
ctx.load_cert_chain("fullchain.pem", "privkey.pem")
httpd.socket = ctx.wrap_socket(httpd.socket, server_side=True)
httpd.serve_forever()
```

Confirm that each URL returns its key, and that the port is reachable from the machine running Safari.

```bash
curl -s https://<host>:18080/key/1 | xxd -p
# 16cf4232a86364b519e1982a27d90087
```

**2. Enable DRM** in the LLHLS publisher of the application, with `<DRM><Enable>true</Enable><InfoFile>DrmInfo.xml</InfoFile></DRM>`.

**3. Write the DRM info file.** The `<Key>` values must match what the key server returns, each `<ContentKey>` needs its own `<FairPlayKeyUrl>`, and the host must be reachable from the machine running Safari.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<DRMInfo>
    <DRM>
        <Name>IdentityRotation</Name>
        <VirtualHostName>default</VirtualHostName>
        <ApplicationName>app</ApplicationName>
        <StreamName>stream*</StreamName>

        <CencProtectScheme>cbcs</CencProtectScheme>
        <KeyRotationPeriod>30</KeyRotationPeriod>

        <Keys>
            <ContentKey>
                <KeyId>11111111111111111111111111111111</KeyId>
                <Key>16cf4232a86364b519e1982a27d90087</Key>
                <Iv>572547f914e34dc68ba9ba9ef91d4c4a</Iv>
                <Keyformat>identity</Keyformat>
                <FairPlayKeyUrl>https://example.com:18080/key/1</FairPlayKeyUrl>
            </ContentKey>
            <ContentKey>
                <KeyId>22222222222222222222222222222222</KeyId>
                <Key>0f1e2d3c4b5a69788796a5b4c3d2e1f0</Key>
                <Iv>112233445566778899aabbccddeeff00</Iv>
                <Keyformat>identity</Keyformat>
                <FairPlayKeyUrl>https://example.com:18080/key/2</FairPlayKeyUrl>
            </ContentKey>
            <ContentKey>
                <KeyId>33333333333333333333333333333333</KeyId>
                <Key>00112233445566778899aabbccddeeff</Key>
                <Iv>ffeeddccbbaa99887766554433221100</Iv>
                <Keyformat>identity</Keyformat>
                <FairPlayKeyUrl>https://example.com:18080/key/3</FairPlayKeyUrl>
            </ContentKey>
        </Keys>
    </DRM>
</DRMInfo>
```

**4. Publish an H.264 and AAC stream** and play it in Safari. Only H.264 and AAC are encrypted, so a stream of any other codec is not a valid subject for this test. A scheduled channel that switches items every 10 seconds is worth using as well, because it makes track changes and key rotations overlap.

**5. Watch the playlist across rotations.**

```bash
watch -n2 'curl -sk "https://<host>:<tls_port>/<app>/<stream>/chunklist_0_video_<stream_key>_llhls.m3u8" | grep -E "EXT-X-KEY|EXT-X-MAP|EXT-X-DISCONTINUITY"'
```

Pass conditions, over at least 4 rotations so that the list wraps back to the first key:

- Playback continues across every rotation, with no stall, no reload and no visible artifact.
- The `EXT-X-KEY` URI advances `/key/1`, `/key/2`, `/key/3` and then back to `/key/1`.
- A new `EXT-X-MAP` accompanies each rotation, and no `EXT-X-DISCONTINUITY` is emitted for it. A track change may still emit one.
- The video and audio chunklists report the same version and the same key.
- The key server log shows a request for each key as it comes into effect.
- Every playlist, initialization segment and media segment request returns 200.

**6. Confirm the content matches the advertised key.** The initialization segment of a version carries the key id it was encrypted with, so it has to match the key the playlist advertises for that version.

```bash
curl -sk "https://<host>:<tls_port>/<app>/<stream>/init_0_video_<stream_key>_v<version>_llhls.m4s" | xxd -p | tr -d '\n' | grep -o '22222222222222222222222222222222'
```

**7. Check the configuration errors are reported.** Each of these is a separate run:

- `<KeyRotationPeriod>1h</KeyRotationPeriod>`: the log warns that the value is not a number of seconds, the key is not rotated, and DRM still applies.
- `<Keys></Keys>` with no `<ContentKey>`: the log reports the DRM info file is invalid and the stream starts without DRM.
- A single `<ContentKey>`: the log warns once that only one key is configured, and the stream plays with that key.
- `<Pssh>` or `<FairPlayKeyUrl>` left on the `<DRM>` node while `<Keys>` is used: the log warns that the element is ignored.

### DIFF
--- a/src/base/event/command/rotate_drm_key.h
+++ b/src/base/event/command/rotate_drm_key.h
@@ -3,7 +3,7 @@
 //  MediaEvent Payload
 //
 //  Created by Getroot
-//  Copyright (c) 2026 AirenSoft. All rights reserved.
+//  Copyright (c) 2026 OvenMediaLabs. All rights reserved.
 //
 //==============================================================================
 #pragma once
--- a/src/base/event/command/rotate_drm_key.h
+++ b/src/base/event/command/rotate_drm_key.h
@@ -1,0 +1,39 @@
+//==============================================================================
+//
+//  MediaEvent Payload
+//
+//  Created by Getroot
+//  Copyright (c) 2026 AirenSoft. All rights reserved.
+//
+//==============================================================================
+#pragma once
+
+#include "../event_command.h"
+
+class EventCommandRotateDrmKey : public EventCommand
+{
+public:
+	EventCommandRotateDrmKey()
+	{
+	}
+
+	Type GetType() const override
+	{
+		return Type::RotateDrmKey;
+	}
+
+	ov::String GetTypeString() const override
+	{
+		return "RotateDrmKey";
+	}
+
+	bool Parse(const std::shared_ptr<ov::Data> &data) override
+	{
+		return true;
+	}
+
+	std::shared_ptr<ov::Data> ToData() const override
+	{
+		return std::make_shared<ov::Data>();
+	}
+};
--- a/src/publishers/llhls/llhls_stream.cpp
+++ b/src/publishers/llhls/llhls_stream.cpp
@@ -2323,8 +2323,8 @@ void LLHlsStream::OnMediaChunkUpdated(const int32_t &track_id, const uint32_t &s
 		// Advertise the EXT-X-KEY of the key this version was actually encrypted with.
 		// The packager recorded it when the version's initialization section was created,
 		// so this holds even when a rotation and a track change land close together. A
-		// version that was produced in the clear records a scheme of None and so
-		// registers no key.
+		// version produced in the clear is registered with a scheme of None, from which
+		// the chunklist ends the scope of the preceding key.
 		auto content_version = segment->GetTrackVersion();
 		auto registered_it = _last_registered_cenc_version.find(track_id);
 		bool already_registered = (registered_it != _last_registered_cenc_version.end() && registered_it->second >= content_version);
--- a/src/publishers/llhls/llhls_stream.cpp
+++ b/src/publishers/llhls/llhls_stream.cpp
@@ -8,6 +8,9 @@
 //==============================================================================
 #include "llhls_stream.h"
 
+#include <algorithm>
+#include <cctype>
+
 #include <base/ovlibrary/hex.h>
 #include <base/publisher/application.h>
 #include <base/publisher/stream.h>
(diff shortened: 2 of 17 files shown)


src/base/event/command/rotate_drm_key.h
```diff
@@ -0,0 +1,39 @@
+//==============================================================================
+//
+//  MediaEvent Payload
+//
+//  Created by Getroot
+//  Copyright (c) 2026 AirenSoft. All rights reserved.
```
dimiden: Let's use `OvenMediaLabs` here instead of `AirenSoft`

src/publishers/llhls/llhls_stream.cpp
```diff
@@ -153,7 +156,13 @@ bool LLHlsStream::Start()
...
 	if (drm_config.IsEnabled())
 	{
-		GetDrmInfo(drm_config.GetDrmInfoPath(), _cenc_property);
+		_drm_info_path = drm_config.GetDrmInfoPath();
+		if (GetDrmInfo(_drm_info_path, _cenc_key_list, _key_rotation_period_ms) == true && _cenc_key_list.empty() == false)
+		{
+			// The first key in the list is the current one; rotations advance the index
+			_current_key_index = 0;
+			_cenc_property = _cenc_key_list[_current_key_index];
+		}
```
dimiden, line 173: `Start()` passes the members `_cenc_key_list`/`_key_rotation_period_ms` straight into `GetDrmInfo()`, which clears the list up front and, on a mid-parse failure, returns `false` while leaving the keys it parsed so far behind. A later `RotateDrmKey()` event could then rotate into those leftover keys from a file that failed to load. `RotateDrmKey()` already avoids this by parsing into locals and committing only on success - could `Start()` do the same?

For example:
```
	uint64_t rotation_period_ms = 0;
	if (GetDrmInfo(_drm_info_path, key_list, rotation_period_ms) == true && key_list.empty() == false)
	{
		_cenc_key_list = std::move(key_list);
		_key_rotation_period_ms = rotation_period_ms;
		// The first key in the list is the current one; rotations advance the index
		_current_key_index = 0;
		_cenc_property = _cenc_key_list[_current_key_index];
	}
```

dimiden (approved):
Overall looks good! I left two nitpick comments, so please check them and make updates if necessary.

src/base/event/command/rotate_drm_key.h
getroot: Applied

src/publishers/llhls/llhls_stream.cpp
getroot: Applied

PR #2263, opened 2026-07-27
feat(alert): add server identification info (serverInfo) to alert notifications

choigilhoon:
## Why

When multiple OvenMediaEngine instances report to the same notification server, there is
no way to tell which server sent a given alert. The payload only contains `sourceUri`
(`#<vhost>#<app>/<stream>`), which is not unique across servers, so operators running
multi-server deployments cannot distinguish the origin of each notification.

## What

Add a top-level `serverInfo` object to every alert notification payload (INGRESS,
EGRESS, and INTERNAL_QUEUE types):

```json
{
	"serverInfo": {
		"serverID": "b8e2f6a1-58f6-4451-b6f6-97f2b3ee4c39",
		"serverName": "MyEdge-01",
		"hostname": "ome-edge-01",
		"ipAddresses": ["203.0.113.10", "192.168.0.220"]
	},
	"sourceUri": "#default#app/stream",
	...
}
```

- `serverID`: Unique server ID, generated at first startup and persisted in the
  `Server.id` file in the configuration directory.
- `serverName`: The value of `<Server><Name>` in the configuration. Omitted when not set.
- `hostname`: OS hostname of the machine running OvenMediaEngine (the pod name on
  Kubernetes, the container ID on Docker unless `--hostname` is specified).
- `ipAddresses`: IP addresses of the server, captured at server startup. The public IP
  addresses resolved from `<StunServer>` (if configured) come first, followed by the
  local interface addresses (excluding loopback and IPv6 link-local addresses).
  Omitted when no address is available.

Implementation notes:

- The server info is built once in `Alert::Start()` and injected at serialization time
  (`NotificationData::ToJsonString()`), so every notification type gets it without
  touching each `NotificationData` construction site.
- The `X-OME-Signature` HMAC is computed over the final body, so signature validation
  on the receiver side is unaffected.
- No new configuration is required; this is a backward-compatible payload addition.
- `docs/alert.md` is updated with the new payload example and field description.

### Setup

1. Add the following to `Server.xml` (the rest of the configuration can stay default):

   ```xml
   <Name>TestServer-01</Name>
   ...
   <Alert>
   	<Url>http://127.0.0.1:18090/alert</Url>
   	<SecretKey>test_secret</SecretKey>
   	<Timeout>3000</Timeout>
   	<Rules>
   		<Ingress>
   			<StreamStatus />
   			<MinBitrate>50000000</MinBitrate>
   		</Ingress>
   	</Rules>
   </Alert>
   ```

   `MinBitrate` is set to 50 Mbps on purpose so that any test stream triggers
   `INGRESS_BITRATE_LOW`.

2. Run an HTTP server on port 18090 that logs the body and the `X-OME-Signature`
   header of every POST request and responds with `200 OK`.

3. Start OvenMediaEngine and publish a test stream for about 15 seconds, then stop it:

   ```bash
   ffmpeg -re -f lavfi -i testsrc=size=640x360:rate=30 -f lavfi -i sine \
          -c:v libx264 -b:v 500k -g 30 -pix_fmt yuv420p -c:a aac \
          -t 15 -f flv rtmp://127.0.0.1:1935/app/stream
   ```

### Test 1 — `serverInfo` is present in every notification

The receiver must get 4 notifications in this order: `INGRESS_STREAM_CREATED`,
`INGRESS_STREAM_PREPARED`, `INGRESS_BITRATE_LOW`, `INGRESS_STREAM_DELETED`.

**Verify:** every payload contains a top-level `serverInfo` object like below, and
the three fields hold the same values across all 4 notifications:

```json
"serverInfo": {
	"serverID": "5e7f7a75-c16c-44f0-ac41-4cf25c5dbc82",
	"serverName": "TestServer-01",
	"hostname": "<hostname of the test machine>",
	"ipAddresses": ["<IP addresses of the test machine>"]
}
```

- `serverID` must equal the content of the `Server.id` file in the configuration
  directory.
- `serverName` must equal `<Server><Name>` in `Server.xml`.
- `hostname` must equal the output of the `hostname` command on the test machine.
- `ipAddresses` must equal the output of the `hostname -I` command on the test machine
  (IPv6 link-local addresses are excluded). If `<StunServer>` is configured in
  `Server.xml`, the public IP address resolved from it must appear first in the list.

### Test 2 — signature is still valid

**Verify:** for each notification, base64url(HMAC-SHA1(`test_secret`, body)) matches
the `X-OME-Signature` header. This confirms adding `serverInfo` does not break
receiver-side signature validation.

### Test 3 — `serverName` is omitted when `<Name>` is not set

Remove `<Name>TestServer-01</Name>` from `Server.xml`, restart OvenMediaEngine, and
publish the stream again.

**Verify:**
- `serverInfo` contains only `serverID` and `hostname` — no `serverName` key.
- `serverID` is identical to the value from Test 1 (persisted in `Server.id`
  across restarts).


src/monitoring/alert/alert.cpp
```diff
@@ -22,6 +25,26 @@ namespace mon::alrt
...
+		if (server_config->GetName().IsEmpty() == false)
+		{
+			server_info["serverName"] = server_config->GetName().CStr();
+		}
+
+		char hostname[HOST_NAME_MAX + 1] = {0};
+		if (::gethostname(hostname, sizeof(hostname) - 1) == 0)
+		{
+			server_info["hostname"] = hostname;
+		}
```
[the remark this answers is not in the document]
choigilhoon: This is intentional and consistent with the rest of the payload schema — `serverName` is likewise omitted when `<Server><Name>` is not configured, and existing fields such as `sourceUri` / `sourceInfo` / `extraInfo` are also conditionally included, so receivers should treat `serverInfo` fields as optional. In practice `gethostname()` cannot fail here since a valid `HOST_NAME_MAX`-sized buffer is always passed, and emitting an empty string would be worse for receivers than omitting the key. Clarified this in `docs/alert.md`.

src/monitoring/alert/alert.cpp
```diff
@@ -22,6 +27,51 @@ namespace mon::alrt
 		Stop();
 	}
 
+	static Json::Value MakeServerInfo(const std::shared_ptr<const cfg::Server> &server_config)
+	{
+		Json::Value server_info;
```
getroot, line 31: `serverID` is only stable when the configuration directory is writable. `LoadServerID()` ignores the `StoreServerID()` return value, so on a read-only conf directory (ConfigMap mount on Kubernetes, some Docker setups) a new UUID is generated on every restart and each restart looks like a different server to the receiver.

Since stable server identification is the whole point of this field, could we note that in `docs/alert.md` and log a warning when persisting fails?

src/monitoring/alert/alert.cpp
```diff
@@ -22,6 +27,51 @@ namespace mon::alrt
...
+
+		server_info["serverID"] = server_config->GetID().CStr();
+
+		if (server_config->GetName().IsEmpty() == false)
+		{
```
getroot, line 36: The analytics payload in `src/monitoring/event.cpp` always emits `serverName` even when it is empty, while here the key is dropped. A receiver that parses both payloads then sees two different shapes. Is omitting it the intended choice?

src/monitoring/alert/alert.cpp
```diff
@@ -22,6 +27,51 @@ namespace mon::alrt
...
+			server_info["serverName"] = server_config->GetName().CStr();
+		}
+
```
getroot, line 39: The same hostname lookup already exists in `src/config/engine/data_source.cpp` for the `${env:HOSTNAME}` macro, with a different buffer convention. A small shared helper such as `ov::GetHostname()` would keep the two from diverging, and would give us one place to deal with `HOST_NAME_MAX`, which is not defined on macOS.

src/monitoring/alert/notification_data.cpp
```diff
@@ -34,11 +34,17 @@ namespace mon::alrt
...
-	ov::String NotificationData::ToJsonString() const
+	ov::String NotificationData::ToJsonString(const Json::Value &server_info) const
 	{
 		// Make request message
 		Json::Value jv_root;
 
+		// Server info
+		if (server_info.isNull() == false)
+		{
+			jv_root["serverInfo"] = server_info;
```
getroot, line 45: `tests/framework/verifier.py` currently reads only `sourceUri` and tells instances apart by the alert URL path, which is exactly the limitation this field removes. Asserting `serverInfo.serverID` there, or using it for instance matching, would pin this payload contract down in e2e.

getroot (approved):
I approve merging after the current reviews are addressed.

src/monitoring/alert/alert.cpp
choigilhoon: Good catch — fixed. `LoadServerID()` now logs a warning when persisting fails, and `docs/alert.md` notes that the configuration directory must be writable for the ID to remain stable across restarts.

src/monitoring/alert/alert.cpp
choigilhoon: Yes, intentional. The alert payload uses conditional keys throughout (`sourceUri` / `sourceInfo` / `extraInfo`), so omitting an unset `serverName` follows the same convention and is documented as such. The analytics payload in `event.cpp` is a separate contract with a different consumer, so I'd rather not couple the two.

src/monitoring/alert/alert.cpp
choigilhoon: Done — added `ov::Platform::GetHostname()` with a `HOST_NAME_MAX` fallback for platforms that don't define it (e.g. macOS), and switched both this code and the `${env:HOSTNAME}` macro handling in `data_source.cpp` to use it.

src/monitoring/alert/notification_data.cpp
choigilhoon: That e2e framework is not part of this repository (it lives in the Enterprise side), so there is nothing to update in this PR — the `serverInfo.serverID` assertion can be added there when this change is ported.

src/monitoring/alert/alert.cpp
```diff
@@ -8,6 +8,11 @@
 //==============================================================================
 #include "alert.h"
 
+#include <climits>
+#include <unistd.h>
+
+#include <modules/address/address_utilities.h>
```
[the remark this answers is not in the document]
choigilhoon: Fixed — added the explicit `<algorithm>` include.

PR #2264, opened 2026-07-27
feat(drm): add HEVC (H.265) subsample encryption

Keukhan:
## Summary

#2258 added the H.265 slice-header parser and made encrypted HEVC use the `hvc1` sample entry, but the real HEVC CENC encryption was left as an empty stub (`Encryptor::GenerateSubSamplesFromH265` just returned `false`). This PR fills in that stub, so HEVC tracks can now be DRM-encrypted (cbcs/cenc) the same way H.264 already can.

## Changes

- **Implement `GenerateSubSamplesFromH265`** (`cenc.cpp`): it walks each NAL unit and decides which bytes stay clear and which are encrypted. The NAL length prefix, the 2-byte NAL header, and the slice header stay clear; the rest of the slice is encrypted. The slice-header size comes from `H265Parser::ParseSliceHeader` (added in #2258). At the end it checks that the byte total matches the sample size.
- **Add the `AppendCencSubSample` helper**: the clear-size field in `senc` is only 16-bit, so it cannot hold a value over 65535. This helper splits a large clear run into several subsamples. The H.264 path now uses it too, which fixes an old bug where a clear run over 65535 was silently truncated.
- **Allow H265 in `IsCencSupportedCodec`** (`cenc.h`) so CENC is offered for HEVC.
- **Add a null check for the AVC config record** in `GenerateSubSamplesFromH264` (the HEVC path already had one).

## Testing

Tested end-to-end:
1. An encoder (e.g. OBS) sends an H.265 stream to OME over Enhanced RTMP (E-RTMP) or SRT.
2. OME is set to Bypass, so the H.265 video is passed through without transcoding.
3. Played the encrypted stream back and confirmed it works across various platforms and browsers.

Note: desktop Chrome and Firefox cannot play encrypted HEVC, so DRM playback is verified on Apple (Safari / FairPlay), Android (Widevine), and TV / other devices.

### DIFF
--- a/src/modules/containers/bmff/cenc.cpp
+++ b/src/modules/containers/bmff/cenc.cpp
@@ -12,9 +12,23 @@
 #include "cenc.h"
 #include "modules/bitstream/h264/h264_parser.h"
 #include "modules/bitstream/h264/h264_decoder_configuration_record.h"
+#include "modules/bitstream/h265/h265_parser.h"
+#include "modules/bitstream/h265/h265_decoder_configuration_record.h"
 
 namespace bmff
 {
+    // senc BytesOfClearData is 16-bit, so split clear runs larger than 65535 across subsamples.
+    static void AppendCencSubSample(std::vector<Sample::SubSample> &sub_samples, uint32_t clear_bytes, uint32_t cipher_bytes)
+    {
+        while (clear_bytes > 0xFFFF)
+        {
+            sub_samples.emplace_back(static_cast<uint16_t>(0xFFFF), 0);
+            clear_bytes -= 0xFFFF;
+        }
+
+        sub_samples.emplace_back(static_cast<uint16_t>(clear_bytes), cipher_bytes);
+    }
+
     Encryptor::Encryptor(const std::shared_ptr<const MediaTrack> &media_track, const CencProperty &cenc_property)
     {
         _cenc_property = cenc_property;
@@ -119,6 +133,12 @@ namespace bmff
 	bool Encryptor::GenerateSubSamplesFromH264(const std::shared_ptr<const MediaPacket> &media_packet, std::vector<Sample::SubSample> &sub_samples)
 	{
 		auto avcc = std::static_pointer_cast<AVCDecoderConfigurationRecord>(_media_track->GetDecoderConfigurationRecord());
+		if (avcc == nullptr)
+		{
+			logte("AVC decoder configuration record is null");
+			return false;
+		}
+
 		auto nal_length_size = avcc->LengthMinusOne() + 1;
 
         uint32_t total_bytes = 0;
@@ -192,7 +212,7 @@ namespace bmff
 					cipher_bytes = nal_length - (H264_NAL_UNIT_HEADER_SIZE + slice_header.GetHeaderSizeInBytes());
                     total_bytes += clear_bytes + cipher_bytes;
                     
-					sub_samples.emplace_back(clear_bytes, cipher_bytes);
+					AppendCencSubSample(sub_samples, clear_bytes, cipher_bytes);
 
                     logtt("VCL NAL Unit Type : %d, Clear Bytes : %u, Protected Bytes : %u", ov::ToUnderlyingType(nal_header.GetNalUnitType()), clear_bytes, cipher_bytes);
 
@@ -217,7 +237,7 @@ namespace bmff
         if (clear_bytes > 0)
         {
             logtt("Last NAL Unit is not a video slice, clear bytes : %u", clear_bytes);
-            sub_samples.emplace_back(clear_bytes, cipher_bytes);
+            AppendCencSubSample(sub_samples, clear_bytes, cipher_bytes);
             total_bytes += clear_bytes;
         }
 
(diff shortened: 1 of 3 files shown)


src/modules/containers/bmff/cenc.cpp
```diff
@@ -234,10 +254,124 @@ namespace bmff
...
+
+		if (total_bytes != media_packet->GetDataLength())
+		{
+			logte("Total subsample bytes (%u) is not equal to sample data length (%zu)", total_bytes, media_packet->GetDataLength());
+			return false;
+		}
+
+		logtt("Subsample count : %zu Total subsamples : %u / %zu", sub_samples.size(), total_bytes, media_packet->GetDataLength());
+
+		return true;
```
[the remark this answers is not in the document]
Keukhan: Good catch, this is a real problem. Fixed in 1a08746a.

Rather than adding the check to both `GenerateSubSamplesFromH264` and `GenerateSubSamplesFromH265`, I put it in `Encryptor::Encrypt` right after `GenerateSubSamples` returns. That covers H.264, H.265 and any future codec from one place, and it has both the final subsample vector and the IV size available.

I also extracted the size formula into `SampleAuxInfo::CalcSencAuxInfoSize()` so the guard and the `saiz` writer cannot drift apart, and corrected `GetSencAuxInfoSize()`'s return type, which was `int8_t` rather than `uint8_t`.

Confirmed boundaries with the extracted formula:

```
cenc (16-byte IV)   : max 39 subsamples (252 bytes) | 40 -> 258, truncates to 2
cbcs (constant IV)  : max 42 subsamples (254 bytes) | 43 -> 260, truncates to 4
full-sample enc     : unaffected (no subsamples, guard does not fire)
```

One note on the failure mode: returning `false` here drops the sample, so a stream that consistently exceeds the limit will stall rather than produce a corrupt fragment. That seemed better than emitting a `saiz` entry that silently misaligns every following entry, but it means the error would repeat every frame, so the log is emitted only once per track.

src/modules/containers/bmff/cenc.cpp
```diff
@@ -217,7 +237,7 @@ namespace bmff
         if (clear_bytes > 0)
         {
             logtt("Last NAL Unit is not a video slice, clear bytes : %u", clear_bytes);
-            sub_samples.emplace_back(clear_bytes, cipher_bytes);
+            AppendCencSubSample(sub_samples, clear_bytes, cipher_bytes);
             total_bytes += clear_bytes;
```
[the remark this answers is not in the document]
Keukhan: Same issue as the H.265 path, fixed in 1a08746a by a single guard in `Encryptor::Encrypt` that covers both codecs. See the discussion on the H.265 thread for the details and the confirmed boundaries.

Worth noting that this one predates the PR: the H.264 path could already exceed the `saiz` limit with a multi-slice stream before `AppendCencSubSample` existed. The clear-run splitting only adds to it when a clear run is over 64 KB, which needs something like an unusually large SEI.

src/modules/containers/bmff/cenc.h
```diff
@@ -34,6 +34,7 @@ namespace bmff
 		switch(codec_id)
 		{
 			case cmn::MediaCodecId::H264:
+			case cmn::MediaCodecId::H265:
```
dimiden, line 37: `IsCencSupportedCodec` now returns true for H265, but `H265Parser::ParseSliceHeader` deliberately refuses several conforming HEVC classes: range/SCC extension PPS (`h265_parser.cpp:1512`), `slice_type > 2` (`:1557`), `inter_ref_pic_set_prediction_flag` + `lists_modification_present_flag` (`:1628`), `PicSizeInCtbsY == 0` (`:1531`).

For a Main 4:2:2 10 or Main 4:4:4 passthrough ingest with DRM enabled, `pps_range_extension_flag == 1` makes every frame fail at `cenc.cpp:353`, so every sample is dropped and no media segment is ever produced. Before this PR the same stream was downgraded to `scheme = None` at `bmff_packager.cpp:53` / `llhls_stream.cpp:1846` and played in the clear, so this is a regression from "plays unencrypted" to "does not play at all".

`IsCencSupportedCodec` only takes a codec id, so it cannot express "H265 but not this profile". A track-level preflight check at packager creation that excludes such a track from CENC once (same convention as the unsupported-codec path) would keep the pre-PR behaviour instead of failing per frame. Note also that the real cause is only logged at `logtd`, so operators just see "Failed to parse H265 slice header" repeated at frame rate.

src/modules/containers/bmff/cenc.cpp
```diff
@@ -69,6 +83,27 @@ namespace bmff
...
 
+        // saiz can describe up to MAX_SENC_AUX_INFO_SIZE bytes of aux info per sample
+        uint8_t stored_iv_size = 0;
+        if (_cenc_property.per_sample_iv_size != 0 && _cenc_property.iv != nullptr)
+        {
+            stored_iv_size = static_cast<uint8_t>(_cenc_property.iv->GetLength());
+        }
+
+        const auto aux_info_size = Sample::SampleAuxInfo::CalcSencAuxInfoSize(stored_iv_size, sub_samples.size());
+        if (aux_info_size > Sample::SampleAuxInfo::MAX_SENC_AUX_INFO_SIZE)
```
dimiden, line 94: The `MAX_SENC_AUX_INFO_SIZE` guard itself is correct (`saiz` sample_info_size is 8-bit), but the effective cap is 39 subsamples for `cenc` (16 + 2 + 6n <= 255) and 42 for `cbcs` (2 + 6n <= 255), and subsample count equals slice segments per access unit.
H.265 Table A.4 `MaxSliceSegmentsPerPicture` is 75 at level 4/4.1 and 200 at level 5 and above, so a fully conforming 1080p stream can carry roughly twice the cap. OME does not control the slice count of HEVC input (passthrough or hardware encoder only), and the 8-bit `saiz` field cannot be widened.
Exceeding the cap drops every frame permanently, and pre-PR the same track was downgraded to `scheme = None` and played in the clear. Deciding once per track and excluding it from CENC would be preferable to dropping frames. If a full-sample fallback is considered instead, it has to be latched for the whole track: `WriteSencBox` derives `UseSubSampleEncryption` from the fragment's first sample only (`bmff_packager.cpp:2305`), so a fragment mixing subsample and full-sample entries produces an unparseable `senc`.

src/modules/containers/bmff/cenc.cpp
```diff
@@ -234,10 +282,131 @@ namespace bmff
...
+
+			H265NalUnitHeader nal_header;
+			if (H265Parser::ParseNalUnitHeader(nalu->GetDataAs<uint8_t>(), nalu->GetLength(), nal_header) == true)
+			{
+				if (nal_header.IsVideoSlice() == true)
+				{
+					// Get Slice Header Size
+					// TODO: NAL header parsed twice (also inside ParseSliceHeader). Dedup later.
+					H265SliceHeader slice_header;
+					if (H265Parser::ParseSliceHeader(nalu->GetDataAs<uint8_t>(), nalu->GetLength(), slice_header, hvcc) == false)
```
dimiden, line 353: `GenerateSubSamplesFromH265` relies solely on the track's `hvcc`; in-band VPS/SPS/PPS carried in the sample are only counted as clear bytes and never folded back (`cenc.cpp:379-386`). Upstream does not repair this either: `HEVCDecoderConfigurationRecord::Equals()` compares only profile_idc, level_idc, width and height (`h265_decoder_configuration_record.cpp:366`), and on the AnnexB path the record is built only while `media_track->IsValid() == false` (`mediarouter_nomalize.cpp:712`). If the record pointer never changes, `StampTrack` does not republish the track and `Packager::UpdateMediaTrack` never rebuilds the Encryptor, so the stale `hvcc` persists.

Two consequences. A new `pps_id` at unchanged resolution/profile/level makes `ParseSliceHeader` fail with `PPS not found`, dropping every frame. Changed SPS content at unchanged resolution/profile/level makes `ParseSliceHeader` succeed with a wrong `CeilLog2(PicSizeInCtbsY)` and a wrong short_term_ref_pic_set bit count, so `GetHeaderSizeInBytes()` is wrong and the subsample boundary lands inside the slice header or leaves payload bytes in the clear, with no error reported.

Minimal in-scope fix: accumulate in-band VPS/SPS/PPS into a local record inside the loop and parse slice headers against that, so in-band-only parameter sets work without touching the shared track DCR. Fixing `Equals()` and the `IsValid()` gate affects H.264 as well and would fit better as a separate change.

src/modules/containers/bmff/cenc.cpp
```diff
@@ -69,6 +83,27 @@ namespace bmff
...
+        {
+            if (_aux_info_size_error_logged == false)
+            {
+                _aux_info_size_error_logged = true;
+                logte("SENC aux info size (%zu) exceeds the saiz limit (%zu). Subsample count (%zu) is too large to be packaged. "
+                      "Reduce the number of slices per frame of the input stream. (This message is logged only once)",
+                      aux_info_size, Sample::SampleAuxInfo::MAX_SENC_AUX_INFO_SIZE, sub_samples.size());
+            }
+
+            return false;
```
dimiden, line 104: Worth noting where this new `return false` (and every `return false` added in `GenerateSubSamplesFromH265`) actually ends up. `SampleBuffer::AppendSample` propagates it and `FMP4Packager::AppendSample` returns false, but the caller discards the result and returns true (`llhls_stream.cpp:1772` `packager->AppendSample(media_packet);`). There is no teardown, no segment discard and no error path, so an `Encryptor::Encrypt` failure becomes a silent frame drop with one `logte` per frame; if the cause persists, the track dies permanently while the stream still reports success.

The discarded return value is pre-existing and outside this diff, but this PR adds the first HEVC paths that actually reach it (see the range/SCC, saiz-cap and stale-hvcc comments), so every new failure return here is effectively "drop this frame forever, quietly". Deciding once per track instead of failing per sample avoids the problem without touching the publisher; handling the return value would fit as a separate change.

dimiden:
@Keukhan 
I’ve left a few comments, so please take a look when you can.

src/modules/containers/bmff/cenc.cpp
Keukhan: this is pre-existing behaviour shared with H.264  `GenerateSubSamplesFromH264` parses slice headers against the track `avcc` the same way.
I'd rather fix both codecs in a separate PR than patch only H.265 here. Note the AnnexB path needs more than `Equals()`: `hevc_config` is only built while `IsValid() == false` (`mediarouter_nomalize.cpp:712`).

src/modules/containers/bmff/cenc.cpp
Keukhan: the discarded return value is a separate change. 
The per-frame logging is the same in `GenerateSubSamplesFromH264`, so I'd rather rate-limit both paths in that separate PR than only H.265 here.

src/modules/containers/bmff/cenc.h
Keukhan: I don't think this is a regression. If DRM configured HEVC track was downgraded to `scheme = None ` and served in the clear.  that is the part that shouldn't happen, and no output is safer than unprotected output. So the fix is to make the encryption work, not to restore the clear fallback. 

range/SCC extension PPS parsing turn out to be a problem, we can improve that code later.

src/modules/containers/bmff/cenc.cpp
Keukhan: The saiz field is only 8 bits and we cannot make it bigger, so there is no good way around this limit. Dropping the sample is what we want here, and clear output (schme = None ) is not an option . same reason as the range/SCC thread.

getroot (approved):

PR #2265, opened 2026-07-27
feat(modules): add TaskPool for running short tasks on shared workers

getroot:
A module that needs to run something off its own thread has to create a thread for it, and a module that does this per stream ends up with one thread per stream. The work in question is usually a name lookup or a request to a remote service, which spends its time waiting rather than running, so a thread each is a poor trade.

TaskPool runs such tasks on workers shared by every module. `Post()` hands a task over and returns, `Submit()` does the same and gives back a `std::future` for the result. The workers start with the first task and stay until shutdown, so a server that never posts anything holds no threads, and a task that arrives once the queue is full is rejected rather than queued without end. The worker count and the queue limit are set under `<Modules><TaskPool>` in Server.xml.

## Test

Unit tests in `src/modules/task_pool/task_pool_test.cpp`, run with:

    cmake -B build/Debug -S . -DOME_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug -G Ninja
    ninja -C build/Debug ome_test_modules
    ./build/Debug/bin/ome_test_modules --gtest_filter='TaskPool.*'

Pass condition: all 13 TaskPool tests pass. They cover posting and running a task, taking a result through `Submit()`, submitting a callable that is not a temporary, reporting a task's exception through the future, starting the workers with the first task, running one task per worker at the same time, a worker surviving a task that throws, rejecting a task once the queue is full, running every task posted from many threads at once, `Stop()` waiting for a task that is still running, `Stop()` breaking the promise of a task it drops, taking no task after `Stop()`, and `Stop()` being safe to call twice.

Regression check: `./build/Debug/bin/ome_test_modules` and `./build/Debug/bin/ome_test_base` pass in full (265 and 195 tests).

Server check: start OME with the `<TaskPool>` block present in Server.xml and confirm the startup log line reporting the configured values, for example

    TaskPool is configured - threads: 4, max tasks: 128

Pass condition: the logged values match the configuration file, and they fall back to 4 threads and 128 tasks when the block is left out.


Keukhan (approved):
Well done! copilot. i'm counting on you

PR #2266, opened 2026-07-28
refactor(llhls): drive DRM key rotation through one path for every key source

getroot:
DRM key rotation walked the keys of the DRM info file directly: it re-read the file on the media thread at every rotation, advanced an index into the list, and applied the key in place. That shape only fits keys that are already on disk. It also left the file's stated rotation period as two states where the manual described three, so a stream could not be set up to rotate only when asked to.

Rotation now runs as one path, whatever serves the keys. A stream asks for the key of a period, that key is fetched ahead of the rotation on a shared worker rather than on the media thread, and a rotation applies what is already waiting. `DrmInfo` holds what the file states for the stream, `<DRMProvider>` decides who is asked, and the period index decides which key that is: for the keys the file lists, the index walks them as a cycle. `<KeyRotationPeriod>` now reads as three states, where the element left out keeps one key, `0` rotates only on request, and a value above `0` rotates on that period.

Filling in what a key does not carry, the FairPlay protection system header and the block layout of the protection scheme, moved to `CencProperty::Complete()`, next to the fields it sets, so every provider completes a key the same way.

### Setup

Enable DRM on the LLHLS publisher of an application, and give the DRM info file an entry that lists three keys:

    <DRM>
        <Name>Rotation</Name>
        <VirtualHostName>default</VirtualHostName>
        <ApplicationName>app</ApplicationName>
        <StreamName>*</StreamName>
        <CencProtectScheme>cbcs</CencProtectScheme>
        <KeyRotationPeriod>30</KeyRotationPeriod>
        <Keys>
            <ContentKey>
                <KeyId>11111111111111111111111111111111</KeyId>
                <Key>16cf4232a86364b519e1982a27d90087</Key>
                <Iv>572547f914e34dc68ba9ba9ef91d4c4a</Iv>
                <Keyformat>identity</Keyformat>
                <FairPlayKeyUrl>https://HOST:18080/key/1</FairPlayKeyUrl>
            </ContentKey>
            <ContentKey>... second key ...</ContentKey>
            <ContentKey>... third key ...</ContentKey>
        </Keys>
    </DRM>

Publish an H.264 + AAC stream to that application.

### 1. The keys are walked as a cycle

Sample the chunklist every few seconds for two minutes.

Pass condition: the `#EXT-X-KEY` line moves from the first key to the second to the third and back to the first, and the `URI` and `IV` of each match the `<ContentKey>` they came from. `DRM key rotation to key period N` appears about every 30 seconds with N increasing by one, and each rotation is followed by `Fetched the DRM key of period N+1`.

### 2. Rotation does not disturb playback

Pass condition: no `#EXT-X-DISCONTINUITY` is added at a rotation, the `KEYID` (or `URI` for `identity`) of the chunklist matches the `tenc` box of the initialization segment it references, and a player keeps playing across rotations.

### 3. The file is read once

Pass condition: the DRM info file is not read again while the stream runs. Keys added to the file are picked up by the streams that start from then on.

### 4. A single key, and no rotation

Pass condition: an entry with one key and a `<KeyRotationPeriod>` reports once that there is nothing to rotate to, and asks for no further key. An entry that leaves `<KeyRotationPeriod>` out keeps one key and never rotates.

### 5. `<DRMProvider>` values

Pass condition: `Manual` and the element left out are accepted in any letter case. Any other value is refused when the file is read, naming the value.

### Unit tests

    ninja -C build/Debug ome_test_publishers_llhls ome_test_modules ome_test_base
    ./build/Debug/bin/ome_test_publishers_llhls
    ./build/Debug/bin/ome_test_modules
    ./build/Debug/bin/ome_test_base

Pass condition: all pass (17, 266 and 195 tests).

### DIFF
--- a/src/publishers/llhls/llhls_stream.cpp
+++ b/src/publishers/llhls/llhls_stream.cpp
@@ -161,9 +161,10 @@ bool LLHlsStream::Start()
 		_drm_info_path = drm_config.GetDrmInfoPath();
 
 		// Parsed into a local and committed only on success, so a file that fails midway
-		// leaves no keys behind for a later rotation to pick up
+		// leaves no keys behind for a later rotation to pick up. A file that says nothing
+		// about this stream leaves the provider unset and the stream unprotected.
 		DrmInfo drm_info;
-		if (GetDrmInfo(_drm_info_path, drm_info) == true)
+		if ((GetDrmInfo(_drm_info_path, drm_info) == true) && (drm_info.provider.IsEmpty() == false))
 		{
 			_drm_info = std::move(drm_info);
 
@@ -470,7 +471,7 @@ bool LLHlsStream::GetDrmInfo(const ov::String &file_path, DrmInfo &drm_info)
 
 				if (is_seconds == false)
 				{
-					logtw("LLHlsStream(%s/%s) - DRM info file(%s) has KeyRotationPeriod(%s), which is not a number of seconds. The key is not rotated automatically.", GetApplication()->GetVHostAppName().CStr(), GetName().CStr(), final_path.CStr(), rotation_period_value.CStr());
+					logtw("LLHlsStream(%s/%s) - DRM info file(%s) has KeyRotationPeriod(%s), which is not a number of seconds, so it is read as if it were not stated and the key is not rotated.", GetApplication()->GetVHostAppName().CStr(), GetName().CStr(), final_path.CStr(), rotation_period_value.CStr());
 				}
 				else
 				{
--- a/src/publishers/llhls/llhls_stream.cpp
+++ b/src/publishers/llhls/llhls_stream.cpp
@@ -22,6 +22,8 @@
 
 #include <pugixml-1.9/src/pugixml.hpp>
 
+#include <modules/task_pool/task_pool.h>
+
 #include "llhls_application.h"
 #include "llhls_private.h"
 #include "llhls_session.h"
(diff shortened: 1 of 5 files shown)


src/publishers/llhls/llhls_stream.cpp
```diff
@@ -158,18 +160,35 @@ bool LLHlsStream::Start()
...
+			}
+			else
+			{
+				logte("LLHlsStream(%s/%s) - Could not get the first DRM key, the stream is not protected", GetApplication()->GetVHostAppName().CStr(), GetName().CStr());
+
+				// Cleared so that a stream left without a key does not keep looking for a
+				// rotation it has nothing to rotate to
+				_drm_info = DrmInfo();
+			}
 		}
```
dimiden, line 192: `GetDrmInfo()` returns true with an empty `DrmInfo` when no `<DRM>` entry matches this stream, so every unprotected stream of a DRM-enabled app logs two `logte` on start.
Guard on a matched entry: `&& drm_info.provider.IsEmpty() == false`

dimiden (changes requested):
Even for streams with a pattern not configured in `DrmInfo.xml`, starting the stream can potentially trigger the error `Could not get the first DRM key, the stream is not protected`.

This could cause operational issues, so it looks like it needs to be fixed.

src/publishers/llhls/llhls_stream.cpp
getroot: This was already resolved earlier.

dimiden: Okay!

dimiden (approved):

PR #2267, opened 2026-07-28
fix(stream): keep runtime state shared between stream copies

getroot:
Since #2251 every module works on its own copy of the stream info, and state written on the source after that copy was taken never reaches it. The published time is one of them: it is stamped when the stream goes on air, which is always after the copies exist, so every module reads an epoch value. Anything that consumes it is affected, and it surfaced as LLHLS and HLS emitting EXT-X-PROGRAM-DATE-TIME from 1970. The source URL, rewritten on every pull-stream failover, goes stale the same way.

Add info::StreamStats, one instance shared by every copy of a stream, and move the state that changes after the stream was created into it: the created time, the published time and on-air state, the prepared time, and the media source. This is the per-stream counterpart of TrackStats, which is already shared per track. The stream description keeps travelling with the copies as immutable snapshots.

## Test

Publish an RTMP stream and play the LLHLS playlist of the output stream. EXT-X-PROGRAM-DATE-TIME must show the wall clock time at which the input stream started, and must keep advancing in real time. Run ome_test_base and the InfoStreamSourceUrl cases must pass.

### DIFF
--- a/src/mediarouter/mediarouter_stream.cpp
+++ b/src/mediarouter/mediarouter_stream.cpp
@@ -41,6 +41,7 @@ using namespace cmn;
 MediaRouteStream::MediaRouteStream(const std::shared_ptr<info::Stream> &stream, cmn::MediaRouterStreamType type, uint32_t worker_id)
 	: _worker_id(worker_id),
 	  _stream(stream),
+	  _stats(stream->GetStats()),
 	  _packets_queue(nullptr, 600)
 {
 	SetType(type);
@@ -91,7 +92,7 @@ void MediaRouteStream::OnStreamPrepared(bool completed)
 {
 	_is_stream_prepared = completed;
 
-	_stream->GetStats()->SetPrepared(completed);
+	_stats->SetPrepared(completed);
 }
 
 bool MediaRouteStream::IsStreamPrepared()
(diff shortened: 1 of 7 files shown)


src/mediarouter/mediarouter_stream.cpp
```diff
@@ -152,14 +154,18 @@ bool MediaRouteStream::IsStreamReady()
...
 {
-	auto now = std::chrono::steady_clock::now();
+	if (_stream->IsOnAir() == false)
+	{
+		return;
+	}
 
 	// Anchor at the first media packet so connection/handshake delay is excluded; full silence is the provider's job.
-	if (_first_media_recv_time_set == false)
+	auto media_start_time = _stream->GetStats()->GetPublishedTimeSteady();
```
Keukhan: `GetPublishedTimeSteady() `value is set through SetOnAir() for every stream. 
 in` MediaRouterApplicationConnector::SendFrame()`, on the first media packet. 
Multiplex is the only exception: **MultiplexStream::Publish() calls SetPublishedTime()** directly (src/providers/multiplex/multiplex_stream.cpp:347), before any media flows.

So for a multiplex channel this anchor is the publish time, not the first packet, and the timeout can behave differently there.

If this is out of scope for this PR, it can be improved in a separate PR.

Keukhan (approved):

src/mediarouter/mediarouter_stream.cpp
getroot: Applied

PR #2268, opened 2026-07-28
feat(thumbnail): add AVIF thumbnail support

Keukhan:
## Summary

Adds AVIF as a thumbnail codec. Supersedes #2218 (original work by @naanlizard).

Unlike JPEG/PNG/WEBP, libaom-av1 emits only the AV1 bitstream, so each encoded still is wrapped into a HEIF container before it leaves the encoder. Known limitation: a non-BT.709 source is mistagged, since nothing in the filter graph converts the frames (TODO in `SetParamsAvif()`).

## Changes

- **`modules/containers/avif`** (new): `avif::Packager` writes the one-image AVIF boxes on top of the existing `Av1Parser` / `AV1DecoderConfigurationRecord` - no libavformat muxer. `av1C`, `ispe` and `colr` are all derived from the in-band sequence header, so the container cannot disagree with what was encoded.
- **`AVCodecImageEncoder`**: AVIF encoder parameters (allintra, CRF, CICP), and `ReceivePacket()` wraps each still through the packager.
- **`Av1SequenceHeaderSummary`**: now reports the `color_config()` CICP fields the parser already read and discarded.
- **Plumbing**: `MediaCodecId::Avif` / `BitstreamFormat::AVIF` through the transcoder, the thumbnail publisher and the codec-id switches.

## Testing

**Manual** - add an `Image` encode with the `avif` codec and enable the Thumbnail publisher:

```xml
<Encodes>
    <Image>
        <Codec>avif</Codec>
        <Width>320</Width>
        <Height>180</Height>
        <Framerate>1</Framerate>
    </Image>
</Encodes>
```

Then open `http://<host>:<port>/<app>/<stream>/thumb.avif` in a browser. The thumbnail rendering is the pass condition; a broken/blank image means the still was served without a valid container.

**Unit tests** - opt-in, as documented in `cmake/README.md`:

```bash
cmake -B build/Debug -S . -DCMAKE_BUILD_TYPE=Debug -DOME_BUILD_TESTS=ON -G Ninja
cmake --build build/Debug --target tests
ctest --test-dir build/Debug -L modules
```

The new packager and AV1 parser tests run in `ome_test_modules` (label `modules`).

**Also verified** - the packager's output decodes with libavif, and 8 consecutive live fetches over 24s produced 8 distinct decodable files, confirming libaom repeats the sequence header in every temporal unit - the assumption the packager rests on.

### DIFF
--- a/src/transcoder/codec/encoder/encoder_avcodec_image.cpp
+++ b/src/transcoder/codec/encoder/encoder_avcodec_image.cpp
@@ -8,7 +8,7 @@
 //==============================================================================
 #include "encoder_avcodec_image.h"
 
-#include <unistd.h>
+#include <modules/containers/avif/avif_packager.h>
 
 #include "../../transcoder_private.h"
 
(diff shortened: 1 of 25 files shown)


src/transcoder/codec/encoder/encoder_avcodec_image.cpp
```diff
@@ -86,6 +115,9 @@ bool AVCodecImageEncoder::OpenCodec()
 		case cmn::MediaCodecId::Png:
 			result = SetParamsPng();
 			break;
+		case cmn::MediaCodecId::Avif:
+			result = SetParamsAvif();
```
getroot, line 119: If other AV1 encoders are added down the road (SVT-AV1, av1_nvenc, etc.), this part will likely need to be extended. Of course, since there's only one encoder for now and there's no immediate issue, please just keep it in mind for future extension.

getroot (approved):
LGTM

naanlizard:
Already merged of course, but my thoughts:

Good progress! It's a bit out of scope but the big differences between what we're running ourselves and this

1. WebP generation is bugged (preset overrides compression levels) so webp CPU usage is higher than necessary. On my test machine, about 3x cheaper when you fix this and drop from compression level 4 to 1.
2. Color handling of course, which is a TODO. Proper color handling will increase CPU usage a bit, about 20% for total CPU usage creating a thumbnail. 
3. Tuning options we added would be nice - WebP Preset, Quality, Method, and lossless. JPEG/AVIF ChromaSampling, JPEG QScale, AVIF Speed, Crf, PassthroughAV1. 

The passthrough for AV1 ingest is pretty slick but I get putting it off. We don't even use it, since nobody streams with AV1 hah

PR #2270, opened 2026-07-28
fix(http2): fix HPACK stale indexes and header block interleaving

getroot:
When an HTTP/2 connection lives long enough for the HPACK dynamic table to fill up, the server keeps sending table indexes for header entries that were already evicted. The client treats this as a connection level compression error and closes the connection, so LLHLS playback stalls and reconnects in a loop. With the ETag module enabled this happens within seconds on Safari, and even without it the connection breaks about once a minute because the Date header fills the table. Responses of one connection are also produced by multiple threads, so two header blocks could interleave on the wire and corrupt the compression state the same way.

Evicted entries are now removed from the lookup state, an index that no longer exists is treated as a miss and logged instead of being sent, each header block is encoded and written to the socket as one unit per connection, header frames respect the HTTP/2 default frame size limit, and headers with a unique value per response (currently ETag) are no longer inserted into the dynamic table.

### Setup

- Run this build with HTTP2 and LLHLS modules enabled and TLS configured on the LLHLS publisher port. Enable the ETag module in Server.xml. It makes the bug reproduce within seconds instead of a minute:

```xml
<Modules>
    <ETag>
        <Enable>true</Enable>
    </ETag>
</Modules>
```

- Publish a test stream:

```sh
ffmpeg -re -f lavfi -i "testsrc2=size=1280x720:rate=30" -f lavfi -i "sine=frequency=1000" \
  -c:v libx264 -preset ultrafast -tune zerolatency -g 30 -c:a aac \
  -f flv "rtmp://<host>:1935/app/stream"
```

### Test 1. Safari playback with ETag enabled

- Play the stream over HTTPS in Safari (macOS or iOS) for at least 10 minutes.
- Pass: no stall and no periodic hiccup. On master the stream visibly stutters or stalls every 5 to 10 seconds because Safari reconnects in a loop.

### Test 2. Default configuration (ETag disabled)

- Disable the ETag module and play in Safari for at least 5 minutes.
- Pass: no stall around the 1 minute mark and none after. On master the connection silently breaks about once per minute. On a fast network playback may still look fine, so check the server log for repeated disconnect and reconnect of the playback connection instead of relying on the picture alone.

### Log check, applies to both tests

- The server log must contain no `STALE INDEX` warning. This warning comes from a guard added in this change that fires whenever the encoder is about to reference a table entry that no longer exists, so even a single occurrence is a failure regardless of how playback looks.

### DIFF
--- a/src/modules/http/hpack/encoder.cpp
+++ b/src/modules/http/hpack/encoder.cpp
@@ -16,6 +16,9 @@ namespace http
 	{
 		bool Encoder::UpdateDynamicTableSize(size_t size)
 		{
+			// A size update must sit at the start of a header block, so it cannot be
+			// signalled while another stream is in the middle of encoding one
+			ov::LockGuard<ov::Mutex> block_lock(_header_block_lock);
 			ov::LockGuard<ov::Mutex> lock(_encoder_lock);
 			if (_table_connector.UpdateDynamicTableSize(size) == false)
 			{
--- a/src/modules/http/hpack/encoder.h
+++ b/src/modules/http/hpack/encoder.h
@@ -9,6 +9,9 @@
 #pragma once
 
 #include <base/ovlibrary/ovlibrary.h>
+
+#include <unordered_set>
+
 #include "data_structure.h"
 #include "table_connector.h"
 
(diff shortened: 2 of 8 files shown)


Keukhan (approved):
There don't seem to be any critical issues beyond what Copilot reported. LGTM

PR #2272, opened 2026-07-29
fix(transcoder): handle media property changes in filters correctly

getroot:
## Problem

When a Scheduled Channel switches between items with different media properties, the video filter logs `Changing video frame properties on the fly is not supported by all filters` at every transition because the buffer source is created without color space and range information.

There is also a race in the filter rebuild path. The rebuild flag is not tied to the frame that triggered it, so when a format boundary is crossed while older frames are still queued (for example, a channel joining playback right before an item boundary), the queued old-format frames are pushed into the rebuilt graph. The audio buffer source rejects them with an error that puts the filter into the ERROR state, and all audio input is dropped until the next format change rebuilds the filter again.

## Solution

The color matrix and range of decoded frames now flow into the input track and the buffer source arguments, and they are preserved across the GPU to host transfer, so the filter graph is always created with the same color tags as the frames it receives.

Input format changes (video resolution, pixel format, and color tags, audio sample rate/layout/format) are now detected per frame at the consumption position in the filter worker, so the filter is rebuilt exactly at the boundary frame regardless of queue depth. The pixel format is compared with the decoder-side format label, so hardware frames do not falsely mismatch the downloaded host format. The audio resampler also drops a mismatched frame instead of latching the ERROR state, matching the video rescaler behavior.

## Test

Preparation: a schedule channel with a transcoding output profile (video encode + AAC audio), whose program alternates two files, where the files differ in both color metadata and audio sample rate (for example, one bt709/tv 48kHz file and one 44.1kHz file without color metadata). The difference can be checked with `ffprobe -show_entries stream=color_space,color_range,sample_rate <file>`.

1. Play the stream and watch the server log across at least 10 item transitions.
   - Pass: video and audio play back normally across every transition, and the log shows no `Changing video frame properties on the fly is not supported by all filters` warning. This also holds when the two files have the same resolution and differ only in color metadata. Before this fix, the warning appeared at every transition between the two files.
2. Set the item durations short (2 to 3 seconds), restart the server about 10 times, and check the log and playback right after each startup.
   - Pass: no `Changing audio frame properties on the fly is not supported` error and no filtering error alert right after startup, and audio is audible from the beginning every time. This targets a startup timing issue that only occurs when playback joins the schedule just before an item boundary, so short items and repeated restarts are needed to hit it. Before this fix, when it occurred, audio stayed silent until the first item ended.

Unit tests covering the color conversion and the filter rebuild logic are included and run in CI.

### DIFF
--- a/src/base/info/media_track.cpp
+++ b/src/base/info/media_track.cpp
@@ -92,6 +92,8 @@ bool MediaTrack::Update(const MediaTrack &media_track)
 	_resolution_conf = media_track._resolution_conf;
 	_b_frames = media_track._b_frames.load();
 	_colorspace = media_track._colorspace.load();
+	_color_matrix = media_track._color_matrix.load();
+	_color_range = media_track._color_range.load();
 	_preset = media_track._preset;
 	_profile = media_track._profile;
 	_thread_count = media_track._thread_count.load();
--- a/src/base/info/media_track_test.cpp
+++ b/src/base/info/media_track_test.cpp
@@ -1,0 +1,29 @@
+//==============================================================================
+//
+//  OvenMediaEngine
+//
+//  Created by Jeheon Han
+//  Copyright (c) 2026 OvenMediaLabs. All rights reserved.
+//
+//==============================================================================
+#include <base/info/media_track.h>
+#include <gtest/gtest.h>
+
+TEST(MediaTrackUpdate, CopiesColorFields)
+{
+	MediaTrack source;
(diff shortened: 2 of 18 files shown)


Keukhan (approved):
It looks fine overall. But right now, it only fixed the warning log that was occurring in the filter. Later, it needs to improve this so it can specify Color Matrix and Color Range on the encoder, and have the filter convert the color to match the encoder's settings.

Thanks.

PR #2273, opened 2026-07-30
fix(build): retry dependency downloads instead of aborting on transient server errors

choigilhoon:
## Why

`cmake -P cmake/InstallPrerequisites.cmake` downloads every dependency with
`curl -sSLf <url> | tar -xz`. A single transient server error kills the whole
build — for example, GitHub briefly answering HTTP 500 for a release asset:

```
curl: (22) The requested URL returned error: 500

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
CMake Error at cmake/InstallPrerequisites.cmake:229 (message):
  [OME Prerequisites] FAILED: libpcre2
```

This aborted a release Docker build on both amd64 and arm64 at the same
minute; the same URL served HTTP 200 shortly after. Since curl is piped
straight into tar, `--retry` alone cannot fix it: a retry restarts the
transfer from the beginning while tar has already consumed the bytes of the
failed attempt, corrupting the stream.

## What

- Add an `ome_fetch` shell helper, injected into every script generated by
  `ome_run`: download the archive **to a file** with
  `curl -sSLf --retry 8`, then extract it with
  `tar -x --strip-components=1` and remove it.
- Replace all 18 `curl ... | tar ...` call sites with `ome_fetch <url>`.

Details:

- `--retry 8` with curl's default exponential backoff (1s, 2s, 4s, ...)
  rides out outages of up to ~4 minutes. curl only retries transient
  failures (HTTP 408/429/500/502/503/504 and timeouts) — genuine errors
  such as a 404 from a wrong version still fail immediately.
- tar auto-detects the compression when reading from a file, so the helper
  covers both the `.tar.gz` and the `.tar.bz2` (x264, jemalloc) archives.
- The success path is unchanged: retries only add time when a download
  actually fails.

## Test

1. Normal build: `cmake -P cmake/InstallPrerequisites.cmake` (or a full
   Docker image build).
   - Pass: every dependency downloads, extracts, and builds exactly as
     before.
2. Transient-failure simulation: point one `*_SOURCE_URL` at a local HTTP
   server that answers 500 for the first two requests and then serves the
   archive.
   - Pass: curl logs the two 500s, retries with backoff, and the build
     continues to completion instead of aborting.
3. Genuine-failure behavior: point a `*_SOURCE_URL` at a non-existent path
   (404).
   - Pass: the build fails immediately without retrying.

Both the retry path (gz and bz2) and archive-file cleanup were verified
locally with the simulation from step 2.

### DIFF
--- a/cmake/InstallPrerequisites.cmake
+++ b/cmake/InstallPrerequisites.cmake
@@ -206,10 +206,26 @@ message(STATUS "[OME Prerequisites] Prefix: ${PREFIX}")
 # ==============================================================================
 # Helper: run shell command and fail loudly
 # ==============================================================================
+# Download a source archive into the current directory and extract it there.
+# Retries make transient server errors (e.g. GitHub sporadically answering
+# HTTP 500) non-fatal instead of aborting a long build: 8 retries with
+# curl's default exponential backoff (1s, 2s, 4s, ...) ride out outages of
+# up to ~4 minutes, while genuine errors such as 404 still fail immediately.
+# The archive must be fetched to a file first: piping curl straight into tar
+# cannot be retried, because a retry restarts the transfer while tar has
+# already consumed the bytes of the failed attempt. tar auto-detects the
+# compression (gz/bz2) when reading from a file.
+set(_OME_FETCH "ome_fetch()
+{
+    curl -sSLf --retry 8 -o ome_source_archive \"$1\" &&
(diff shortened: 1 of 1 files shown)


Keukhan (approved):
lgtm

PR #2274, opened 2026-07-30
fix(multiplex): fix tap queue overflow warning and unbounded growth

getroot:
**Problem**

When a Multiplex channel taps an already running source, the whole 2 second mirror buffer is injected into the tap queue at once, which immediately triggers the MediaRouterStreamTap threshold warning on multi-track sources. Independently, the channel worker pops one packet per source per round and blocks up to 100ms on an empty tap, so when sources have different packet rates the faster tap queues grow without bound, continuously increasing latency and memory.

**Solution**

The worker now drains each tap in bounded batches without blocking and backs off its idle sleep, so one source can no longer throttle the others. The mirror buffer retains per track in media time: a video track holds only its current GOP from its latest keyframe, so the backfill always starts decodable; non-video is cut at the earliest included keyframe DTS and the result is interleaved by DTS. The backfill is delivered through a separate tap buffer, so its one-shot burst cannot trip the live queue's congestion warning, which now purely means the consumer is falling behind. The tap's need-past-data flag is now atomic, and the misleading buffer_size parameter is renamed to threshold.

**Test**

*Unit tests*

```bash
cmake -B build/Debug -S . -DOME_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug -G Ninja
cmake --build build/Debug
ctest --test-dir build/Debug -L mediarouter
```

Pass condition: all mediarouter tests pass (9 mirror buffer cases cover keyframe-led retention, DTS-aligned cuts, interleaving, and stale-track handling).

*Environment setup*

Application `mux`: enable the Multiplex provider with `<MuxFilesDir>mux_files</MuxFilesDir>`, enable the LLHLS publisher, and use an OutputProfile with only `bypass_video` (`<Bypass>true</Bypass>`) and `bypass_audio` (`<Bypass>true</Bypass>`) encodes.

`mux_files/muxed.mux` (the file name must match the OutputStream Name) combines 3 sources into one output stream:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Multiplex>
    <OutputStream>
        <Name>muxed</Name>
        <BypassTranscoder>true</BypassTranscoder>
    </OutputStream>
    <SourceStreams>
        <SourceStream>
            <Name>input1</Name>
            <Url>stream://default/mux/input1</Url>
            <TrackMap>
                <Track>
                    <SourceTrackName>bypass_video</SourceTrackName>
                    <NewTrackName>input1_video</NewTrackName>
                </Track>
                <Track>
                    <SourceTrackName>bypass_audio</SourceTrackName>
                    <NewTrackName>input1_audio</NewTrackName>
                </Track>
            </TrackMap>
        </SourceStream>
        <SourceStream>
            <Name>input2</Name>
            <Url>stream://default/mux/input2</Url>
            <TrackMap>
                <Track>
                    <SourceTrackName>bypass_video</SourceTrackName>
                    <NewTrackName>input2_video</NewTrackName>
                </Track>
                <Track>
                    <SourceTrackName>bypass_audio</SourceTrackName>
                    <NewTrackName>input2_audio</NewTrackName>
                </Track>
            </TrackMap>
        </SourceStream>
        <SourceStream>
            <Name>input3</Name>
            <Url>stream://default/mux/input3</Url>
            <TrackMap>
                <Track>
                    <SourceTrackName>bypass_video</SourceTrackName>
                    <NewTrackName>input3_video</NewTrackName>
                </Track>
                <Track>
                    <SourceTrackName>bypass_audio</SourceTrackName>
                    <NewTrackName>input3_audio</NewTrackName>
                </Track>
            </TrackMap>
        </SourceStream>
    </SourceStreams>
    <Playlists>
        <Playlist>
            <Name>Mux ABR</Name>
            <FileName>abr</FileName>
            <Rendition>
                <Name>input1</Name>
                <Video>input1_video</Video>
                <Audio>input1_audio</Audio>
            </Rendition>
            <Rendition>
                <Name>input2</Name>
                <Video>input2_video</Video>
                <Audio>input2_audio</Audio>
            </Rendition>
            <Rendition>
                <Name>input3</Name>
                <Video>input3_video</Video>
                <Audio>input3_audio</Audio>
            </Rendition>
        </Playlist>
    </Playlists>
</Multiplex>
```

Feed 3 sources with different frame rates (different rates are required to reproduce the unbounded growth):

```bash
ffmpeg -re -f lavfi -i testsrc2=size=640x360:rate=30 -f lavfi -i "sine=frequency=440:sample_rate=48000" -c:v libx264 -preset veryfast -tune zerolatency -g 60 -b:v 800k -pix_fmt yuv420p -c:a aac -b:a 128k -ar 48000 -ac 2 -f flv rtmp://localhost:1935/mux/input1 -loglevel error &
ffmpeg -re -f lavfi -i testsrc2=size=640x360:rate=15 -f lavfi -i "sine=frequency=880:sample_rate=48000" -c:v libx264 -preset veryfast -tune zerolatency -g 30 -b:v 500k -pix_fmt yuv420p -c:a aac -b:a 128k -ar 48000 -ac 2 -f flv rtmp://localhost:1935/mux/input2 -loglevel error &
ffmpeg -re -f lavfi -i testsrc2=size=640x360:rate=24 -f lavfi -i "sine=frequency=1760:sample_rate=48000" -c:v libx264 -preset veryfast -tune zerolatency -g 48 -b:v 600k -pix_fmt yuv420p -c:a aac -b:a 128k -ar 48000 -ac 2 -f flv rtmp://localhost:1935/mux/input3 -loglevel error &
```

*Scenarios and pass conditions*

1. Channel start: start OME, then the 3 inputs. Pass: the `muxed` stream starts with 6 tracks and `grep -c "MediaRouterStreamTap size has exceeded" ovenmediaengine.log` stays 0.
2. Re-tap: run `touch mux_files/muxed.mux` 10 times with 10 or more seconds between touches (each touch recreates the channel and re-injects past data). Pass: warning count still 0.
3. Sustained run: keep all 3 inputs running 30 minutes or more. Pass: warning count 0 and OME memory stable. Before this fix the faster tap queues grow without bound in this scenario and the warning repeats every 5 seconds with a rising queue value.
4. Source loss and recovery: kill the input2 ffmpeg. The channel logs the untapped source and terminates, then waits for input2. Restart input2. Pass: the channel restarts automatically, warning count still 0.
5. Playback: open `http(s)://<host>:<llhls-port>/mux/muxed/abr.m3u8`. Pass: 3 renditions (30/15/24 fps) are listed and play; `ffprobe` on the URL shows 3 video and 3 audio streams.

Addresses: OvenMediaLabs/OvenMediaEngineEnterprise#125

### DIFF
--- a/src/mediarouter/mediarouter_stream.cpp
+++ b/src/mediarouter/mediarouter_stream.cpp
@@ -29,7 +29,6 @@
 #include "mediarouter_stream.h"
 
 #include <algorithm>
-#include <limits>
 #include <map>
 
 #include <base/event/media_event.h>
@@ -437,21 +436,8 @@ std::shared_ptr<MediaPacket> MediaRouteStream::PopAndNormalize()
 	MediaRouterStats::Update(static_cast<uint8_t>(_type), IsStreamPrepared(), _packets_queue, GetStream(), media_track, pop_media_packet);
 
 	// Mirror Buffer
-	_mirror_buffer.emplace_back(std::make_shared<MirrorBufferItem>(pop_media_packet));
-
-	// Delete old packets
-	for (auto it = _mirror_buffer.begin(); it != _mirror_buffer.end();)
-	{
-		auto item = *it;
-		if (item->GetElapsedMilliseconds() > MEDIA_ROUTE_STREAM_MAX_MIRROR_BUFFER_SIZE_MS)
-		{
-			it = _mirror_buffer.erase(it);
-		}
-		else
-		{
-			++it;
-		}
-	}
+	int64_t dts_us = (int64_t)((double)pop_media_packet->GetDts() * 1000000.0 * media_track->GetTimeBase().GetExpr());
+	RetainMirrorBuffer(_mirror_buffers, pop_media_packet, dts_us);
 
 	return pop_media_packet;
 }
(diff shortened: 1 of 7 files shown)


src/mediarouter/mediarouter_stream_test.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+#include "mediarouter_stream.h"
+
+namespace
+{
+	std::shared_ptr<MediaRouteStream::MirrorBufferItem> MakeItem(cmn::MediaType media_type, uint32_t track_id, bool keyframe)
+	{
+		auto packet = std::make_shared<MediaPacket>(media_type, track_id, nullptr, 0, 0, 0,
+													keyframe ? MediaPacketFlag::Key : MediaPacketFlag::NoFlag,
+													cmn::BitstreamFormat::Unknown, cmn::PacketType::Unknown);
```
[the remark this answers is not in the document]
getroot: wrong

src/mediarouter/mediarouter_stream.cpp
```diff
@@ -675,3 +679,85 @@ std::vector<std::shared_ptr<MediaRouteStream::MirrorBufferItem>> MediaRouteStrea
...
+				{
+					continue;
+				}
+			}
+
+			trimmed.push_back(item);
+		}
+
+		selected = std::move(trimmed);
+	}
```
dimiden: When the cap binds this silently discards packets, and it can discard one video track's past data entirely, with no log at any level - the operator only sees a rendition starting late with nothing to trace it to.
Could we log at debug/warn here when the trim loses packets or drops a whole track?

src/providers/multiplex/multiplex_stream.cpp
```diff
@@ -134,40 +141,57 @@ namespace pvd
...
             if (break_loop)
             {
                 break;
             }
+
+            if (any_packet_popped == false)
+            {
+                // Every tap was empty in this round
+                std::this_thread::sleep_for(std::chrono::milliseconds(1));
+            }
```
dimiden, line 211: On my dev box (24-core Linux), benchmarking this loop shape standalone (a model, not a real OME run, so indicative only): at 3 sources x 77 pkt/s ~92% of rounds find every tap empty, so this sleep fires ~940 times/sec per channel at +0.45pp of one core, almost all of it nanosleep wake overhead rather than queue work. It is a low-rate-only fixed cost - at 3 x 500 pkt/s this loop is already cheaper than master - so an exponential backoff on consecutive empty rounds, capped around 10ms, would drop it without touching the fix itself.

dimiden (approved):
I’ve approved this as I don't see any major issues. Please check my previous comments before merging.

src/mediarouter/mediarouter_stream.cpp
getroot: I changed it to a much better structure.

src/providers/multiplex/multiplex_stream.cpp
getroot: I changed it to a much better structure.

PR #2276, opened 2026-07-30
fix(api): emit track timebase and align the stream info spec with the actual response

dimiden:
## Summary

`GET /v1/vhosts/{vhost}/apps/{app}/streams/{stream}` never returned `timebase`, and the published spec had drifted from the serializer in field coverage, JSON types and required lists.
This fixes the inverted guard that suppressed `timebase` and rewrites the stream info schemas in `api/ome.yml` and in the REST API document directly from `serdes` code.

## Motivation

1. `serdes::SetTimebase()` returned early when the timebase was valid (`GetDen() > 0`), so `timebase` reached the response only for a track whose denominator was 0 or less. The field has been documented since the API was introduced and never shipped. The `OV_ASSERT2` inside `CONVERTER_RETURN_IF` receives the stringified condition, so the assertion is always true and never surfaced the mistake.
2. `api/ome.yml` and the document still described the response as it looked before v0.19.0 and v0.20.0: the `bitrate` family as string, a bypass track carrying only `bypass`, and no `maxWidth`, `maxHeight`, `maxFramerate`, `codecModule`, `codecStatus`, `language`, `characteristics`, `configChangeCount` or `lastConfigChanged`. Data tracks, subtitle tracks and playlists were not modelled at all.
3. The gap is customer facing. An integrator reading the spec was promised a field that never arrives (`timebase`) and types that do not match the payload.

## Changes

Code:

- `SetTimebase()` now returns early on `timebase.IsValid() == false`, which requires both numerator and denominator to be positive, matching `cmn::Timebase::IsValid()` and `MediaTrack::IsValidTimeBase()`. A valid track now carries `timebase`, and a degenerate one omits the field instead of emitting `den: 0`.
- The serializer is shared, so the same field also appears in the transcode webhook body, the alert webhook `sourceInfo` and `parentSourceInfo`, the WebRTC playlist signaling message and the local event log. The change is additive for every consumer.

Spec (`api/ome.yml`):

- `Stream.outputs` becomes an array, its relay behavior is described, and `response` is marked required on the 200 response.
- `VideoTrack` and `AudioTrack` rewritten from the serializer: integer bitrate family, `max*` fields, `codecModule`, `codecStatus`, `language`, `characteristics`, `configChangeCount`, `lastConfigChanged`, each with its presence condition.
- New `DataTrack`, `SubtitleTrack` including the Whisper `stt` object, and `OtherTrack` for track kinds that carry no payload object.
- Response only `StreamPlaylist`, `StreamRendition` and `StreamPlaylistOptions`. The configuration schemas cannot be reused because they declare `webRtcAutoAbr` while the response emits `webrtcAutoAbr`.
- `required` lists reduced to keys the serializer always writes, and `sourceType` enumerated with the values `StringFromStreamSourceType()` returns, including `SRT` and `MPEGTS`.

Document:

- 200 OK sample regenerated for a reproducible RTMP scenario, H264 passed through and AAC transcoded to OPUS. Every value is traced to the code that produces it: the 1/1000 ingest timebase of the RTMP provider, the timebase a bypass track copies from its input, codec module ids of the decoder and encoder, `codecStatus` on a bypass track, and configured versus measured bitrate.
- Field notes added for configured versus measured values, presence conditions, high water marks and float serialization.
- The embedded OpenAPI block now uses the real path with its path parameters and basic auth, matches `api/ome.yml`, and its 404 schema no longer pins a message string the server never sends.
- Response bodies are valid JSON again: the field legends moved out of the `json` fences, and one block also had a trailing comma.

## Limitations

- `Tracks` models `Attachment` and `Unknown` through `OtherTrack`, which only carries `id`, `name` and `type`. No in-tree producer emits those media types today; the branch exists for an OVT origin that sends one over the wire.
- The `framerate` and `keyFrameInterval` families are narrowed to single precision before serialization, so `29.97` is rendered as `29.969999313354492`. This is documented rather than changed, since fixing it would alter values that clients already parse.

## Known issues

- `info::Stream::_playlists` is an unguarded `std::map`. Every `GET .../streams/{stream}` adds the publishers' default playlist to the shared monitoring object through `AddPlaylist()`, while `GetPlaylist()`, `GetPlaylists()` and the copy constructor read the same map without a lock, so concurrent requests can race. This is not introduced or touched here: the write on the GET path dates back to `3b3e7cc69` and the unguarded member to `8ea69e851`. **It will be handled as a separate task.**

### Guard behavior

The `CONVERTER_RETURN_IF` macro and the `SetTimebase()` body were extracted from the tree and compiled against the in-tree `jsoncpp` with the writer settings `ov::Json::Stringify()` uses, to confirm the before and after behavior:

```plain
before: den=90000 -> {}            den=48000 -> {}      den=0 -> {"timebase":{"den":0,"num":1}}
after : den=90000 -> {"timebase":{"den":90000,"num":1}}  num=0 -> {}
```

### Spec and document

- `api/ome.yml` and the embedded OpenAPI block: YAML parse, `$ref` resolution, and every `required` entry checked against the declared properties.
- 200 OK sample parses as JSON, and each field value was cross-checked against the code path that produces it.
- Markdown structure verified: `details` pairs and code fences balanced, `git diff --check` clean.

### DIFF
--- a/api/ome.yml
+++ b/api/ome.yml
@@ -86,7 +86,7 @@ components:
       type: object
       required: [schema,urls]
       properties:
-        schme:
+        schema:
           type: string
         urls:
           $ref: '#/components/schemas/Urls'
@@ -255,27 +255,36 @@ components:
           type: string
         input:
           type: object
-          required: [sourceType,tracks]
+          required: [createdTime,sourceType,tracks]
           properties:
             createdTime:
               type: string
               format: date-time
             sourceType:
               type: string
+              description: Provider that ingested this stream.
+              enum: [WebRTC,Ovt,Rtmp,RtmpPull,Rtsp,RtspPull,Transcoder,SRT,MPEGTS,Scheduled,Multiplex,File,Unknown]
             sourceUrl:
               type: string
+              description: Source address of the input stream. Omitted when the provider has no source address.
             tracks:
               $ref: '#/components/schemas/Tracks'
         outputs:
-          type: object
-          required: [name,tracks]
-          properties:
-            name:
-              type: string
-            tracks:
-              $ref: '#/components/schemas/Tracks'
-            playlists:
-              $ref: '#/components/schemas/Playlists'
+          type: array
+          description: For a source or transcoded stream, the output streams created by the output profiles, or an empty array until one exists. For a relay stream with no child output stream, the relay stream itself as the single entry.
+          items:
+            type: object
+            required: [name,tracks,playlists]
+            properties:
+              name:
+                type: string
+              tracks:
+                $ref: '#/components/schemas/Tracks'
+              playlists:
+                type: array
+                description: Playlists of this output stream. Empty when the output stream has no playlist.
+                items:
+                  $ref: '#/components/schemas/StreamPlaylist'
     StreamPush:
       type: object
       required: [stream,protocol,url]
@@ -639,6 +648,8 @@ components:
           type: boolean
     Timebase:
       type: object
+      required: [num,den]
+      description: Timebase of the track. The owning object omits this field when the track has no valid timebase.
       properties:
         num:
           type: integer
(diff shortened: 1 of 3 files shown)


api/ome.yml
```diff
@@ -86,7 +86,7 @@ components:
       type: object
       required: [schema,urls]
       properties:
-        schme:
+        schema:
```
getroot: it may be 'scheme' ?

getroot (approved):
LGTM, 
plz check my in-line comment before merge it

api/ome.yml
dimiden: Thanks for catching that! I matched it with the previous documentation where `schema` was used, but `scheme` is indeed correct in this context.

Reference: https://ovenmedia.com/docs/ome-enterprise/features/high-availability/clustering#edge

PR #2278, opened 2026-08-01
fix(transcoder): fix out of memory caused by unbounded frame buffering

Keukhan:
## Summary

On a busy server the transcoder's memory kept growing until the process was OOM-killed. The filter held far more frames at once than it needed to:

- `FilterVideoBase`/`FilterAudioBase` collected **everything a single push produced** before any of it was delivered. `ProcessFrameInternal()` drained the backend pipeline into an intermediate `_output_frames` queue, and only after that did the worker start handing frames downstream. Every frame of that batch is alive at the same time, and for hardware pipelines each one pins a GPU surface. `v0.20.7.1` pulled one frame and delivered it right away; the queue was introduced later.
- `SendBuffer()` enqueued frames even when the filter was not created yet, in the `ERROR` state, or already stopped, so frames piled up for a filter that would never drain them.
- The filter input queue allowed 5 frames per track.

The branch also keeps its original fix: the decoder, encoder, and filter worker threads drain their component in a `while (!_kill_flag)` loop, but each loop only recognized a subset of the results its component can return. Unhandled values fell through to the next iteration, so the loop asked again immediately and got the same answer — pinning a core and flooding the log when the value persisted.

## Changes

- **One frame in flight instead of a whole batch** — `_output_frames` is gone. `PopCompletedFrameInternal()` now returns a single completed frame, and the worker calls it until the filter reports `Again`. On the video side it also drives the push: when the backend pipeline has nothing left, it takes the next frame from the FPS filter and feeds it in, so `ProcessFrameInternal()` only hands the input frame to the FPS filter. A frame is delivered downstream as soon as it exists, so a push that produces several frames no longer keeps all of them alive at once.
- **Frames are rejected for a filter that cannot use them** — `SendBuffer()` checks the filter up front and returns `false` when it is missing, in `ERROR`, or `STOPPED`, instead of queueing the frame.
- **Smaller filter input queue** — `MAX_QUEUE_SIZE` 5 → 2.
- **One error report per frame** — errors are reported from the pull side only. A push failure leaves the filter state behind and `PopCompletedFrameInternal()` reports it exactly once, so a single failing frame no longer counts as two `FILTERING_ERROR`s. Failures that only drop a frame and leave the filter usable (FPS filter full, a frame the backend refuses) are logged instead of being raised as errors.
- **Frame processing time is measured where the work happens** — libavfilter does the filtering while the frame is pulled out of the sink, so the weighted average now covers the push *and* `ReceiveFrame()`. Without it the average collapsed to a few microseconds and automatic frame skipping never engaged under load. Delivery to the encoder stays outside the measured region, as before.
- **XMA first-frame device lock dropped** — XMA support is scheduled for removal, so the XMA-only path that processed the first frame under the device mutex is not carried over.
- **Drain loops handle every documented result** — the decoder, encoder, and filter loops now handle every value in `DecodeResult`, `EncodeResult`, and `FilterResult`, and exit with a warning on anything unrecognized.
  - **Decoder** — `NoData`/`DataError` are forwarded once and then end the loop instead of being retried immediately.
  - **Encoder** — `NoData`/`EndOfFile` no longer fall through with no exit; `DataError` is forwarded through `Complete()` before exiting, so `OnEncodedPacket()` counts it toward the encoding-error alert (previously it was only logged).
  - **Filter** — `DataError` is reported once and then ends the loop.

  Exiting a drain loop is not a lost-work path: the thread returns to its blocking input dequeue and drains again on the next packet or frame.

## Testing

1. Configure an encoding profile that uses hardware acceleration and publish a stream to OME (RTMP or another provider).
2. Publish more streams the same way, increasing the number of input streams.
3. Once the added load starves the transcoder of resources and frames start piling up in the queues, watch the GPU memory usage with `nvidia-smi`.
4. **Pass** — GPU memory usage stops growing. **Fail** — it keeps climbing.

### DIFF
--- a/src/transcoder/filter/filter_audio_base.cpp
+++ b/src/transcoder/filter/filter_audio_base.cpp
@@ -20,29 +20,21 @@ FilterResult FilterAudioBase::ProcessFrameInternal(const std::shared_ptr<MediaFr
 		return FilterResult::Error();
 	}
 
-	// Drain every frame produced by this push into the output queue.
-	while (auto completed_frame = ReceiveFrame())
-	{
-		_output_frames.push(std::move(completed_frame));
-	}
-
-	if (GetState() == State::ERROR)
-	{
-		return FilterResult::Error();
-	}
-
 	return FilterResult::NoOutput();
 }
 
 FilterResult FilterAudioBase::PopCompletedFrameInternal()
 {
-	if (_output_frames.empty())
+	if(GetState() == FilterBase::State::ERROR)
+	{
+		return FilterResult::Error();
+	}
+
+	auto completed_frame = ReceiveFrame();
+	if(completed_frame == nullptr)
 	{
 		return FilterResult::NoOutput();
 	}
 
-	auto output_frame = std::move(_output_frames.front());
-	_output_frames.pop();
-
-	return FilterResult::Ready(std::move(output_frame));
+	return FilterResult::Ready(std::move(completed_frame));
 }
--- a/src/transcoder/filter/filter_audio_base.cpp
+++ b/src/transcoder/filter/filter_audio_base.cpp
@@ -15,9 +15,14 @@
 
 FilterResult FilterAudioBase::ProcessFrameInternal(const std::shared_ptr<MediaFrame> &media_frame)
 {
+	if (media_frame == nullptr)
+	{
+		return FilterResult::Error("Received frame is null");
+	}
+
 	if (SendFrame(media_frame) == false)
 	{
-		return FilterResult::Error();
+		logtw("[%s] Dropped a frame that could not be pushed into the backend resampler.", GetLogPrefix().CStr());
 	}
 
 	return FilterResult::NoOutput();
@@ -25,14 +30,21 @@ FilterResult FilterAudioBase::ProcessFrameInternal(const std::shared_ptr<MediaFr
 
 FilterResult FilterAudioBase::PopCompletedFrameInternal()
 {
-	if(GetState() == FilterBase::State::ERROR)
+	if (GetState() == FilterBase::State::ERROR)
 	{
-		return FilterResult::Error();
+		return FilterResult::Error("The filter is in the error state");
 	}
 
 	auto completed_frame = ReceiveFrame();
-	if(completed_frame == nullptr)
+	if (completed_frame == nullptr)
 	{
+		// ReceiveFrame() also returns nullptr when it fails, so the state decides whether
+		// this is an empty pipeline or an error that has to be reported right away.
+		if (GetState() == FilterBase::State::ERROR)
+		{
+			return FilterResult::Error("Failed to receive frame from backend resampler");
+		}
+
 		return FilterResult::NoOutput();
 	}
 
--- a/src/transcoder/filter/filter_video_base.cpp
+++ b/src/transcoder/filter/filter_video_base.cpp
@@ -95,17 +95,6 @@ FilterResult FilterVideoBase::ProcessFrameInternal(const std::shared_ptr<MediaFr
 			}
 		}
 
-		// Drain every frame produced by this push into the output queue.
-		while (auto completed_frame = ReceiveFrame())
-		{
-			_output_frames.push(std::move(completed_frame));
-		}
-
-		if (GetState() == State::ERROR)
-		{
-			return FilterResult::Error();
-		}
-
 		auto elapsed_time_us = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::steady_clock::now() - start_time).count();
 
 		// Update the weighted average frame processing time
@@ -122,15 +111,18 @@ FilterResult FilterVideoBase::ProcessFrameInternal(const std::shared_ptr<MediaFr
 
 FilterResult FilterVideoBase::PopCompletedFrameInternal()
 {
-	if (_output_frames.empty())
+	if(GetState() == FilterBase::State::ERROR)
+	{
+		return FilterResult::Error();
+	}
+
+	auto completed_frame = ReceiveFrame();
+	if(completed_frame == nullptr)
 	{
 		return FilterResult::NoOutput();
 	}
 
-	auto output_frame = std::move(_output_frames.front());
-	_output_frames.pop();
-
-	return FilterResult::Ready(std::move(output_frame));
+	return FilterResult::Ready(std::move(completed_frame));
 }
 
 #if _SKIP_FRAMES_ENABLED
--- a/src/transcoder/filter/filter_video_base.cpp
+++ b/src/transcoder/filter/filter_video_base.cpp
@@ -11,7 +11,6 @@
 
 #include <base/ovlibrary/ovlibrary.h>
 
-#include "../transcoder_gpu.h"
 #include "../transcoder_private.h"
 #include "../transcoder_stream_internal.h"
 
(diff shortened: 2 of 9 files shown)


src/transcoder/transcoder_encoder.cpp
```diff
@@ -505,16 +505,19 @@ void TranscodeEncoder::ThreadLoop()
...
 				{
-					break;
+					Complete(received.result, std::move(received.packet));
+
+					// Keep draining to check whether more packets are pending.
+					continue;
 				}
 
-				Complete(received.result, std::move(received.packet));
+				break;
```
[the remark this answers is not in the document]
Keukhan: It is intentional

src/transcoder/transcoder_filter.cpp
```diff
@@ -295,6 +316,31 @@ void TranscodeFilter::Stop()
...
+		{
+			logtw("[%s] Filter is not created yet. track:%u",
+				  _input_stream_info->GetUri().CStr(),
+				  GetInputTrack()->GetId());
+
+			return false;
+		}
+
+		auto state = _filter_base->GetState();
+		if (state == FilterBase::State::ERROR || state == FilterBase::State::STOPPED)
```
dimiden: This guard runs before the filter error state check in `IsNeedUpdate()` (the one labeled `Check #2` (`transcoder_filter.cpp:395`), which recreates the filter when an XMA filter is in ERROR), which makes the only self-recovery path unreachable.
`v0.20.7.1` also rejected by state, but it ran the recovery check first, so please either drop `ERROR` from the rejected states or set `_setup_pending = true` here and let the frame through.

How it used to recover:

1. On the next frame after ERROR, `IsNeedUpdate()` matches `Check #2` and returns `true`.
2. `_setup_pending` is set and the frame is queued (`v0.20.7.1` called `CreateInternal()` right there instead).
3. The worker picks that frame up, runs `Initialize()`, and swaps in a fresh filter object.
4. That same frame is processed normally and the track survives.

Scenario with this PR:

1. An XMA resource allocation failure puts the filter into `ERROR`.
2. Every later frame is rejected in `SendBuffer()`, so the input queue drains.
3. The worker parks in dequeue and `_setup_pending` is never set.
4. The track produces no output and never recovers until a restart.

src/transcoder/filter/filter_video_base.cpp
```diff
@@ -95,17 +95,6 @@ FilterResult FilterVideoBase::ProcessFrameInternal(const std::shared_ptr<MediaFr
...
-		{
-			_output_frames.push(std::move(completed_frame));
-		}
-
-		if (GetState() == State::ERROR)
-		{
-			return FilterResult::Error();
-		}
-
 		auto elapsed_time_us = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::steady_clock::now() - start_time).count();
```
dimiden: The `while (auto completed_frame = ReceiveFrame())` loop this PR removes from `ProcessFrameInternal()` sat inside the `start_time` measurement, so this average now only measures `SendFrame()`.
libavfilter does the actual filtering at `av_buffersink_get_frame` (that is, in `ReceiveFrame()`), which pins `UpdateSkipFrames()`'s adaptation to 0, so please extend the measured region to cover the `PopCompletedFrameInternal()` drain in `ThreadLoop()`.

Scenario:

1. The filter is genuinely bottlenecked, but the average only reflects the push and stays at a few microseconds.
2. `max_frames_per_second` is computed thousands of times larger than reality.
3. `next_skip_frames` clamps to 0.
4. Automatic frame skipping never engages under load and the backlog piles up in the queues.

src/transcoder/transcoder_filter.cpp
```diff
@@ -264,12 +264,33 @@ void TranscodeFilter::ThreadLoop()
...
+				OnComplete(recv.result, std::move(recv.frame));
+				break;
+			}
+			else
+			{
+				// Unhandled result; leave the loop rather than spinning forever.
+				logtw("[%s] Unexpected result while draining filtered frames. result(%d)", _input_stream_info->GetUri().CStr(), static_cast<int32_t>(recv.result));
+				break;
+			}
 		}
```
dimiden, line 294: The fix itself is sound: it restores the `v0.20.7.1` design of pulling one frame and delivering it right away.
But "it grew whenever the consumer popped slower" and "the input queue was rarely full, so no backpressure reached the provider" do not hold, since the same worker thread empties the queue every iteration and exceed-wait is enabled, so it would be better to lean on the claim that is actually correct: everything a single push produces is collected before any of it is delivered.

src/transcoder/transcoder_filter.cpp
```diff
@@ -264,12 +264,33 @@ void TranscodeFilter::ThreadLoop()
...
 				break;
 			}
+			else if (recv.result == TranscodeResult::DataError)
+			{
+				logte("[%s] Error occurred while draining filtered frames. reason(%s)", _input_stream_info->GetUri().CStr(), recv.error.CStr());
 
-			OnComplete(recv.result, std::move(recv.frame));
+				// Report the error, then stop draining rather than asking the filter
+				// again - it would keep returning the same error.
+				OnComplete(recv.result, std::move(recv.frame));
```
dimiden, line 285: Most `SendFrame()` failure paths also call `SetState(ERROR)`, so this `OnComplete` overlaps with the one from `ProcessFrameInternal()` and each frame is counted as two `FILTERING_ERROR`s.

Scenario:

1. `SendFrame()` fails and calls `SetState(ERROR)` along the way.
2. `ProcessFrameInternal()` returns `Error()`, producing one `OnComplete(DataError)`.
3. The first `PopCompletedFrameInternal()` of the following drain hits the state check and produces another.
4. A single frame is counted as two `FILTERING_ERROR`s, reaching the alert threshold twice as fast.

src/transcoder/transcoder_filter.cpp
```diff
@@ -295,6 +316,31 @@ void TranscodeFilter::Stop()
 
 bool TranscodeFilter::SendBuffer(std::shared_ptr<MediaFrame> buffer)
 {
+	// Check if the filter is ready to process frames
+	{
+		ov::SharedLockGuard lock(_mutex);
+		if (_filter_base == nullptr)
+		{
+			logtw("[%s] Filter is not created yet. track:%u",
```
dimiden: This is on the per-frame path, so it prints dozens of lines per second per track during teardown.
Lowering it to `logtd`, or logging once on the state transition, would be better.

src/transcoder/filter/filter_audio_base.cpp
```diff
@@ -20,29 +20,21 @@ FilterResult FilterAudioBase::ProcessFrameInternal(const std::shared_ptr<MediaFr
...
 	return FilterResult::NoOutput();
 }
 
 FilterResult FilterAudioBase::PopCompletedFrameInternal()
 {
-	if (_output_frames.empty())
+	if(GetState() == FilterBase::State::ERROR)
 	{
-		return FilterResult::NoOutput();
+		return FilterResult::Error();
```
dimiden: nit: `PopCompletedFrameInternal()` returns `FilterResult::Error()` with no message.

dimiden (changes requested):
Reviewed! A few items need to be resolved prior to merging - please check them out.

src/transcoder/transcoder_filter.cpp
Keukhan: This will be deprecated soon. It doesn't matter even if this code is dead code.  thanks.

Keukhan: You're right, Check #2 can't be reached anymore. I'm not restoring it — XMA is going to be removed from the code. I left a TODO there saying it is dead code and will be deleted with the rest of XMA. The XMA first-frame device lock is dropped for the same reason.

src/transcoder/filter/filter_video_base.cpp
Keukhan: Fixed. `ProcessFrameInternal()` now only pushes the frame into the FPS filter. `PopCompletedFrameInternal()` does the rest: pull one finished frame, and if the pipeline is empty, push the next FPS frame and try again. The timer is inside that call, so `ReceiveFrame()` is measured again.

One note: the encoder delivery wait is still not measured, same as before this PR.

src/transcoder/transcoder_filter.cpp
Keukhan: You're right, my description was wrong. The queue could not grow without limit and backpressure was fine. I rewrote it: the real problem is that all frames from one push were collected before any of them was delivered.

src/transcoder/transcoder_filter.cpp
Keukhan: Fixed. Only the pull side reports errors now. A failed push just leaves the filter state behind, and the next `PopCompletedFrameInternal()` reports it once. If the failure only dropped a frame, it is a log line, not an error.

src/transcoder/transcoder_filter.cpp
Keukhan: Changed both to `logtd`.

src/transcoder/filter/filter_audio_base.cpp
Keukhan: Fixed. Every `FilterResult::Error()` now carries a message.

src/transcoder/transcoder_encoder.cpp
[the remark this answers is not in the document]
Keukhan: Intentional. In this drain I only take the packets that come back as `DataReady`. Any other result means there is nothing left to drain, so the loop ends there.

src/transcoder/transcoder_filter.cpp
```diff
@@ -303,6 +349,7 @@ bool TranscodeFilter::SendBuffer(std::shared_ptr<MediaFrame> buffer)
 		_setup_pending = true;
 	}
 
+	// Enqueue the buffer to the input buffer queue for processing by the worker thread.
```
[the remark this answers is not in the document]
Keukhan: Good catch. `IsNeedUpdate()` writes `_last_timestamp`, so it takes an exclusive `ov::LockGuard` now.

src/transcoder/transcoder_filter.cpp
```diff
@@ -295,6 +316,31 @@ void TranscodeFilter::Stop()
 
 bool TranscodeFilter::SendBuffer(std::shared_ptr<MediaFrame> buffer)
 {
+	// Check if the filter is ready to process frames
+	{
+		ov::SharedLockGuard lock(_mutex);
+		if (_filter_base == nullptr)
+		{
+			logtd("[%s] Filter is not created yet. track:%u",
```
dimiden: Lowering these was my suggestion, but combined with the other changes it removes every way to tell that a track has died.
On master a dead filter kept re-raising the alert every 5s; that repetition is now gone and this line is at debug level, invisible under the shipped `info` config, so logging once on the state transition at `logtw` would be better.

Scenario:

1. The filter enters `ERROR`, producing one `logte` line and one alert.
2. The input queue drains and every further signal stops.
3. The track stays dark with nothing left in the production log.

src/transcoder/filter/filter_audio_base.cpp
```diff
@@ -15,34 +15,38 @@
...
-	while (auto completed_frame = ReceiveFrame())
-	{
-		_output_frames.push(std::move(completed_frame));
-	}
-
-	if (GetState() == State::ERROR)
+	if (SendFrame(media_frame) == false)
 	{
-		return FilterResult::Error();
+		logtw("[%s] Dropped a frame that could not be pushed into the backend resampler.", GetLogPrefix().CStr());
```
dimiden, line 25: The audio side only logs a `SendFrame()` failure and moves on, while the equivalent video failure is raised as `FilterResult::Error` at `filter_video_base.cpp:126`.
Audio returned an error before this change, so it would be good to align the two paths.

dimiden (approved):
Overall looks great! Just addressed two minor points - please take another look before merging.

src/transcoder/transcoder_filter.cpp
Keukhan: Good point. A failed filter is now reported once at warning level, at the moment it stops accepting frames, so a dark track still leaves a line in the production log. A planned stop (`STOPPED`) stays at debug, so shutting down a stream with many tracks does not print a burst of warnings. The check moved into `IsReadyToProcess()`.

src/transcoder/filter/filter_audio_base.cpp
Keukhan: Aligned them the other way: the video path now logs the dropped frame and keeps going, like audio does. If the failure was fatal, the filter is left in the `ERROR` state and the check at the top of the next round reports it once.

I did not make audio raise the error instead, because `SendFrame()` returns false both for a fatal failure and for a frame the backend simply refuses (a parameter mismatch). Raising it would count those drops as filtering errors, and on the audio side the same frame would be reported twice — once from the push, once from the drain.

PR #2279, opened 2026-08-03
refactor(trs): AVFrame pixel format helpers in ffmpeg::compat

Keukhan:
## Summary

No functional change: `PixelFormat` and `PlaneCount` resolution is consolidated into `ffmpeg::compat` helpers.

One regression is fixed along the way: the rescaler filter converted a frame's duration to the output track timebase a second time, after `settb` had already done it.

## Changes

- `ffmpeg::compat`: add `GetSampleAVPixelFormat()` / `GetSampleVideoPixelFormat()` / `GetPlaneCount()` and drop the unused `GetHWFramesConstraintsValidSWFormat()`, so `FFmpegMediaFrameData` resolves format, plane count and stride through one place — including hardware frames, whose layout comes from the frames context's `sw_format`.
- `MediaFrameData::GetPlaneData()`: drop the `const` qualifier and return `uint8_t *`, so in-place stages do not need a `const_cast`. Read-only callers are unaffected.
- `FFmpegFilterGraph::PushFrame()`: use `compat::GetPlaneCount()`.
- `FilterLavfiRescaler::ReceiveFrame()`: drop the duplicate duration timebase conversion.

## Test

Not needed. Refactoring with no behavior change.

### DIFF
--- a/src/modules/ffmpeg/ffmpeg_media_frame.cpp
+++ b/src/modules/ffmpeg/ffmpeg_media_frame.cpp
@@ -99,20 +99,12 @@ namespace ffmpeg
 
 	int FFmpegMediaFrameData::GetPlaneCount() const
 	{
-		// Host pixel planes only;
-		if (_frame == nullptr || _frame->hw_frames_ctx != nullptr || _frame->width <= 0 || _frame->height <= 0)
-		{
-			return 0;
-		}
-
-		int planes = ::av_pix_fmt_count_planes(static_cast<AVPixelFormat>(_frame->format));
-		return (planes < 0) ? 0 : planes;
+		return ffmpeg::compat::GetPlaneCount(ffmpeg::compat::GetSampleAVPixelFormat(_frame));
 	}
 
-	const uint8_t *FFmpegMediaFrameData::GetPlaneData(int plane) const
+	uint8_t *FFmpegMediaFrameData::GetPlaneData(int plane)
 	{
-		// Host pixel planes only; a hardware frame has no CPU-side data.
-		if (_frame == nullptr || _frame->hw_frames_ctx != nullptr || _frame->width <= 0 || _frame->height <= 0 ||
+		if (_frame == nullptr || _frame->width <= 0 || _frame->height <= 0 ||
 			plane < 0 || plane >= AV_NUM_DATA_POINTERS)
 		{
 			return nullptr;
@@ -123,8 +115,7 @@ namespace ffmpeg
 
 	int FFmpegMediaFrameData::GetStride(int plane) const
 	{
-		// Host pixel planes only; a hardware frame has no CPU-side data.
-		if (_frame == nullptr || _frame->hw_frames_ctx != nullptr || _frame->width <= 0 || _frame->height <= 0 ||
+		if (_frame == nullptr || _frame->width <= 0 || _frame->height <= 0 ||
 			plane < 0 || plane >= AV_NUM_DATA_POINTERS)
 		{
 			return 0;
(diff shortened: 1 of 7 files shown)


src/transcoder/media_frame.h
```diff
@@ -43,7 +43,7 @@ class MediaFrameData
 	virtual void FillZero() = 0;
 
 	virtual int GetPlaneCount() const { return 0; }
-	virtual const uint8_t *GetPlaneData(int plane) const { return nullptr; }
+	virtual uint8_t *GetPlaneData(int plane) { return nullptr; }
 	virtual int GetStride(int plane) const { return 0; }
```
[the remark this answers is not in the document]
Keukhan: I intentionally removed `const`. Do not raise an issue. thanks.

src/modules/ffmpeg/ffmpeg_media_frame.cpp
```diff
@@ -99,20 +99,12 @@ namespace ffmpeg
...
-		return (planes < 0) ? 0 : planes;
+		return ffmpeg::compat::GetPlaneCount(ffmpeg::compat::GetSampleAVPixelFormat(_frame));
 	}
 
-	const uint8_t *FFmpegMediaFrameData::GetPlaneData(int plane) const
+	uint8_t *FFmpegMediaFrameData::GetPlaneData(int plane)
 	{
-		// Host pixel planes only; a hardware frame has no CPU-side data.
-		if (_frame == nullptr || _frame->hw_frames_ctx != nullptr || _frame->width <= 0 || _frame->height <= 0 ||
+		if (_frame == nullptr || _frame->width <= 0 || _frame->height <= 0 ||
```
getroot, line 107: Dropping the hw_frames_ctx guard changes the failure mode for hardware frames. On a CUDA frame, data[0] and data[1] are non-null device pointers and linesize is the device pitch, and GetPlaneCount() now reports the sw_format plane count, so a host side consumer passes every check here and then reads GPU memory. The old code failed safe with nullptr and 0.

No current caller reaches this path (the only consumer runs on the result of DownloadToHost()), but this PR is where the contract for the planned in-place stages gets defined. I suggest keeping the hardware guard in GetPlaneData() and GetStride(), or stating in the header that IsHardwareFrame() must be checked before calling them. Reporting the sw_format from GetPixelFormat() and GetPlaneCount() as metadata looks fine.

src/transcoder/filter/filter_lavfi_rescaler.cpp
```diff
@@ -329,8 +329,6 @@ std::shared_ptr<MediaFrame> FilterLavfiRescaler::ReceiveFrame()
 			continue;
 		}
 
-		// Convert duration to output track timebase
-		output_frame->SetDuration((int64_t)((double)output_frame->GetDuration() * _input_track->GetTimeBase().GetExpr() / _output_track->GetTimeBase().GetExpr()));
```
getroot, line 333: Verified against FFmpeg n7.1 that settb rescales frame->duration, so removing this line is correct for the version we build against. One note: docs/troubleshooting.md still guides FFMPEG_VERSION=5.1.4, which does not even have AVFrame::duration, so the doc is stale. Not this PR's scope, but worth updating since this fix depends on the newer settb behavior.

src/modules/ffmpeg/ffmpeg_media_frame.cpp
Keukhan: thanks. this function called from enterprise version.

src/transcoder/filter/filter_lavfi_rescaler.cpp
Keukhan: The troubleshooting document will be handled in a separate PR. thanks.

getroot (approved):
LGTM if you check my two comments

PR #2281, opened 2026-08-03
fix(thumbnail): a thumbnail request delays other sockets

Keukhan:
## Summary

A thumbnail request waited up to 5 seconds for an image to arrive on the socket worker thread. Since a physical port uses a single worker by default, every other socket on that port — including the HLS/LLHLS publishers sharing the TLS port — was delayed as well. Requesting a codec that is not configured (e.g. `.png` when only JPEG/WEBP are enabled) hit the full 5 seconds every time, because the image would never arrive.

The policy is now to return immediately: if no image is cached for the requested codec, the request is answered with 404 right away and the worker thread is never held.

Related: OvenMediaLabs/OvenMediaEngineEnterprise#166

## Changes

- The request handler no longer waits for an image (`GetVideoFrameByCodecId(codec_id, 0)`).
- `ThumbnailStream::GetVideoFrameByCodecId()` returns `nullptr` immediately if the stream has no track for the requested codec, and logs which codec is missing.
- Guarded against a null track slot in the codec scan, matching `Stream::GetFirstTrackByType()`.
- Dropped a redundant `Clone()` before `AppendData()`, which already clones.
- Added a `TODO(Keukhan)`: `PullStream()` and the AdmissionWebhooks call still block the socket worker thread. Once the exchange is handed over to a worker thread, waiting for an image can be restored.

## Note

A request issued before the first image has been encoded now returns 404 instead of the image. This is an intentional trade-off — never holding the shared socket worker takes priority.

## Testing

1. HLS + LLHLS + Thumbnail (JPEG/WEBP only, no PNG) on a shared TLS port.
2. Start HLS/LLHLS playback, then request `thumb.png` repeatedly.
3. 404 returns immediately and playback keeps running. `thumb.jpg` still works.

### DIFF
--- a/src/publishers/thumbnail/thumbnail_publisher.cpp
+++ b/src/publishers/thumbnail/thumbnail_publisher.cpp
@@ -355,9 +355,8 @@ std::shared_ptr<ThumbnailInterceptor> ThumbnailPublisher::CreateInterceptor()
 			return http::svr::NextHandler::DoNotCall;	
 		}
 
-		// Return the cached image without waiting, this handler runs on the socket worker thread
-		// TODO(Keukhan): PullStream() and the AdmissionWebhooks call still block that thread. The exchange
-		// should be handed over to a worker thread as the HLS publisher does, then it can wait again.
+		// Do not wait here, this handler runs on the socket worker thread
+		// TODO(Keukhan): PullStream() and AdmissionWebhooks still block it, handle this on a worker thread
 		auto endcoded_video_frame = std::static_pointer_cast<ThumbnailStream>(stream)->GetVideoFrameByCodecId(media_codec_id, 0);
 		if (endcoded_video_frame == nullptr)
 		{
--- a/src/publishers/thumbnail/thumbnail_publisher.cpp
+++ b/src/publishers/thumbnail/thumbnail_publisher.cpp
@@ -357,8 +357,8 @@ std::shared_ptr<ThumbnailInterceptor> ThumbnailPublisher::CreateInterceptor()
 
 		// Do not wait here, this handler runs on the socket worker thread
 		// TODO(Keukhan): PullStream() and AdmissionWebhooks still block it, handle this on a worker thread
-		auto endcoded_video_frame = std::static_pointer_cast<ThumbnailStream>(stream)->GetVideoFrameByCodecId(media_codec_id, 0);
-		if (endcoded_video_frame == nullptr)
+		auto encoded_video_frame = std::static_pointer_cast<ThumbnailStream>(stream)->GetVideoFrameByCodecId(media_codec_id, 0);
+		if (encoded_video_frame == nullptr)
 		{
 			response->AppendString(ov::String::FormatString("There is no thumbnail image"));
 			response->SetStatusCode(http::StatusCode::NotFound);
@@ -367,7 +367,7 @@ std::shared_ptr<ThumbnailInterceptor> ThumbnailPublisher::CreateInterceptor()
 
 		response->SetHeader("Content-Type", MimeTypeFromMediaCodecId(media_codec_id));
 		response->SetStatusCode(http::StatusCode::OK);
-		response->AppendData(endcoded_video_frame);
+		response->AppendData(encoded_video_frame);
 		auto sent_size = response->Response();
 		exchange->Release();
 
--- a/src/publishers/thumbnail/thumbnail_publisher.cpp
+++ b/src/publishers/thumbnail/thumbnail_publisher.cpp
@@ -355,12 +355,14 @@ std::shared_ptr<ThumbnailInterceptor> ThumbnailPublisher::CreateInterceptor()
 			return http::svr::NextHandler::DoNotCall;	
 		}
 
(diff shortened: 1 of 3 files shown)


src/publishers/thumbnail/thumbnail_stream.cpp
```diff
@@ -102,8 +103,20 @@ void ThumbnailStream::SendAudioFrame(const std::shared_ptr<MediaPacket> &media_p
...
+	// If the stream has no track for the codec, return nullptr immediately
+	auto tracks = GetTracks();
+	if (std::any_of(tracks.begin(), tracks.end(), [codec_id](const auto &item) {
+			return (item.second != nullptr) && (item.second->GetCodecId() == codec_id);
+		}) == false)
+	{
+		logtd("There is no %s track in the stream [%s/%s]",
+			  cmn::GetCodecIdString(codec_id), GetApplicationName(), GetName().CStr());
+
+		return nullptr;
```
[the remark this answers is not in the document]
Keukhan: It was left as debug instead of wanring to prevent log spam. this should be resolved by changeing the log level. thanks

src/publishers/thumbnail/thumbnail_stream.cpp
```diff
@@ -102,8 +103,20 @@ void ThumbnailStream::SendAudioFrame(const std::shared_ptr<MediaPacket> &media_p
 
 std::shared_ptr<ov::Data> ThumbnailStream::GetVideoFrameByCodecId(cmn::MediaCodecId codec_id, int64_t timeout_ms)
 {
+	// There is no track for this codec, so the image will never arrive
+	auto tracks = GetTracks();
+	if (std::any_of(tracks.begin(), tracks.end(), [codec_id](const auto &item) {
```
getroot, line 108: Minor: the common case is a cache hit, but this track scan (and the map snapshot copy in `GetTracks()`) now runs before the cache lookup on every request. Checking the cache first and scanning tracks only on a miss would keep the hot path cheaper. Also `std::none_of` would express this without the `== false`. Non-blocking.

src/publishers/thumbnail/thumbnail_stream.cpp
```diff
@@ -102,8 +103,20 @@ void ThumbnailStream::SendAudioFrame(const std::shared_ptr<MediaPacket> &media_p
...
+			return (item.second != nullptr) && (item.second->GetCodecId() == codec_id);
+		}) == false)
+	{
+		logtd("There is no %s track in the stream [%s/%s]",
```
getroot, line 112: With `logtd` an operator in production cannot see why the 404 happens. Debug level is defensible against request spam, just confirming this is intentional.

src/publishers/thumbnail/thumbnail_publisher.cpp
```diff
@@ -355,18 +355,19 @@ std::shared_ptr<ThumbnailInterceptor> ThumbnailPublisher::CreateInterceptor()
 			return http::svr::NextHandler::DoNotCall;	
 		}
 
-		// Wait O seconds for thumbnail image to be received
-		auto endcoded_video_frame = std::static_pointer_cast<ThumbnailStream>(stream)->GetVideoFrameByCodecId(media_codec_id, 5000);
+		// Do not wait here, this handler runs on the socket worker thread
+		// TODO(Keukhan): PullStream() and AdmissionWebhooks still block it, handle this on a worker thread
```
getroot, line 359: Typo carried over from the old code: `endcoded_video_frame` could become `encoded_video_frame` since this line is already touched. Non-blocking.

src/publishers/thumbnail/thumbnail_publisher.cpp
```diff
@@ -355,18 +355,19 @@ std::shared_ptr<ThumbnailInterceptor> ThumbnailPublisher::CreateInterceptor()
...
+		auto endcoded_video_frame = std::static_pointer_cast<ThumbnailStream>(stream)->GetVideoFrameByCodecId(media_codec_id, 0);
 		if (endcoded_video_frame == nullptr)
 		{
 			response->AppendString(ov::String::FormatString("There is no thumbnail image"));
```
getroot, line 363: Both "codec not configured" (permanent) and "first image not encoded yet" (temporary) now return the same 404, so a polling client cannot tell whether a retry makes sense. A 503 or a Retry-After header for the temporary case could distinguish them. Non-blocking, the old behavior was 404 as well.

getroot (approved):
Reviewed and verified the core claims:

- `AppendData()` clones internally (http_response.cpp), so dropping the explicit `Clone()` is safe and it removes a double copy.
- `GetTracks()` returns an atomic_load snapshot, so the new track scan is thread safe, and the null guard matches `GetFirstTrackByType()`.
- With timeout 0 the loop checks the cache exactly once and returns without sleeping, so the worker thread is never held.
- This handler is the only caller of `GetVideoFrameByCodecId()`, so no other path is affected.

All inline comments are non-blocking suggestions. One out-of-scope note: the codec scan loop in `ThumbnailStream::Start()` still dereferences track slots without a null guard, so a follow-up could align it with the new code.

Mergeable as is.

PR #2282, opened 2026-08-03
refactor: harden marker handling in the packagers

getroot:
**Problem**

Each publisher derived the return(IN) marker from an OUT on its own at consumption time, and the marker lifecycle rules were scattered across the consumers, so a segment cut could miss the marker position depending on chunk timing and a pending return point could not be adjusted or cleaned up consistently.

**Solution**

Publishers no longer derive a return marker: the marker pipeline consumes an OUT and its return(IN) as separate events. The marker lifecycle rules are centralized in the marker box, where a return point announced by a planned duration stays adjustable by an explicit IN until it is consumed, a duplicate replaces the pending one idempotently, and leftovers are discarded with a warning. The packager guarantees a segment boundary at the first cuttable frame at or after each marker position.

**Test**

Unit tests (`ome_test_bmff`, 6 scenarios) cover cross-track sequence alignment over repeated marker events in three GOP/segment configurations, marker-position cuts, and the provisional/confirmed IN replacement rules.

### DIFF
--- a/src/base/modules/data_format/cue_event/cue_event.cpp
+++ b/src/base/modules/data_format/cue_event/cue_event.cpp
@@ -8,9 +8,9 @@
 //==============================================================================
 #include "cue_event.h"
 
-std::shared_ptr<CueEvent> CueEvent::Create(CueType cue_type, uint32_t duration_sec, uint32_t elapsed_msec)
+std::shared_ptr<CueEvent> CueEvent::Create(CueType cue_type, uint32_t duration_sec, uint32_t elapsed_msec, bool provisional)
 {
-	return std::make_shared<CueEvent>(cue_type, duration_sec, elapsed_msec);
+	return std::make_shared<CueEvent>(cue_type, duration_sec, elapsed_msec, provisional);
 }
 
 std::shared_ptr<CueEvent> CueEvent::Parse(const std::shared_ptr<const ov::Data> &data)
@@ -22,7 +22,8 @@ std::shared_ptr<CueEvent> CueEvent::Parse(const std::shared_ptr<const ov::Data> 
 
 	ov::ByteStream stream(data);
 
-	if (data->GetLength() != 9)
+	// The provisional flag is appended only when set; older payloads are 9 bytes
+	if (data->GetLength() != 9 && data->GetLength() != 10)
 	{
 		return nullptr;
 	}
@@ -42,7 +43,13 @@ std::shared_ptr<CueEvent> CueEvent::Parse(const std::shared_ptr<const ov::Data> 
 	uint32_t duration_msec = stream.ReadBE32();
 	uint32_t elapsed_msec = stream.ReadBE32();
 
-	return Create(cue_type, duration_msec, elapsed_msec);
+	bool provisional = false;
+	if (data->GetLength() == 10)
+	{
+		provisional = stream.Read8() == 1;
+	}
+
+	return Create(cue_type, duration_msec, elapsed_msec, provisional);
 }
 
 CueEvent::CueType CueEvent::GetCueTypeByName(ov::String type)
@@ -63,20 +70,28 @@ CueEvent::CueType CueEvent::GetCueTypeByName(ov::String type)
 	return CueType::Unknown;
 }
 
-CueEvent::CueEvent(CueType cue_type, uint32_t duration_sec, uint32_t elapsed_msec)
+CueEvent::CueEvent(CueType cue_type, uint32_t duration_sec, uint32_t elapsed_msec, bool provisional)
 {
 	_cue_type = cue_type;
 	_duration_msec = duration_sec;
 	_elapsed_msec = elapsed_msec;
+	_provisional = provisional;
 }
 
 std::shared_ptr<ov::Data> CueEvent::Serialize() const
 {
-	ov::ByteStream stream(5);
+	ov::ByteStream stream(10);
 
 	stream.Write8(static_cast<uint8_t>(_cue_type));
 	stream.WriteBE32(_duration_msec);
 	stream.WriteBE32(_elapsed_msec);
+
+	// Appended only when set, so unrelated events stay wire-identical to
+	// older versions
+	if (_provisional == true)
+	{
+		stream.Write8(1);
+	}
 
 	return stream.GetDataPointer();
 }
@@ -109,4 +124,9 @@ uint32_t CueEvent::GetDurationMsec() const
 uint32_t CueEvent::GetElapsedMsec() const
 {
 	return _elapsed_msec;
+}
+
+bool CueEvent::IsProvisional() const
+{
+	return _provisional;
 }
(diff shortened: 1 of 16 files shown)


src/base/modules/data_format/cue_event/cue_event.cpp
```diff
@@ -63,21 +70,29 @@ CueEvent::CueType CueEvent::GetCueTypeByName(ov::String type)
...
 	_cue_type = cue_type;
 	_duration_msec = duration_sec;
 	_elapsed_msec = elapsed_msec;
+	_provisional = provisional;
 }
 
 std::shared_ptr<ov::Data> CueEvent::Serialize() const
 {
-	ov::ByteStream stream(5);
+	ov::ByteStream stream(10);
```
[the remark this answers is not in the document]
getroot: it's wrong. ByteStream(capacity) : capacity is reserved size

src/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -514,62 +537,48 @@ namespace bmff
...
 				bool last_chunk = can_be_last_chunk && next_frame_is_idr;
 				if (marker_handling == kShouldFlushImmediately)
 				{
 					last_chunk = true;
 				}
 
+				if (last_chunk == true)
+				{
+					// A completed boundary satisfies any pending marker cut
+					_pending_marker_cut_timestamp = -1;
```
Keukhan, line 575: **`_pending_marker_cut_timestamp` is never cleared when the storage force-completes the segment**

Two paths complete a segment, and this clear only covers one.

**Path A** — the packager passes `last_chunk = true`, the storage completes the segment, this block clears the pending cut. Correct.

**Path B** — the packager passes `last_chunk = false`, but `AppendMediaChunk` overwrites its own parameter at `fmp4_storage.cpp:576-583` when the segment exceeds `segment_duration_ms * 2` and completes it anyway. Since `last_chunk` is passed **by value** (`fmp4_storage.h:59`), the packager never learns this, and the pending cut survives.

```
t=3.0s   IN marker at T=3.0s falls in a chunk range on a non-IDR frame
         -> chunk consumes it, _pending_marker_cut_timestamp = 3.0s, last_chunk = false
t=12.0s  segment length 12s > 6s*2 -> storage force-completes it
         -> _pending_marker_cut_timestamp is still 3.0s
t=13.0s  first IDR -> return_cut = (13.0 >= 3.0) = true
         -> the *next* segment is cut after 1s   <-- extra boundary
```

Two consequences:

1. **An extra boundary on video only** — audio can't reach this path, since every audio frame counts as an IDR and its segments always complete at the target. That is exactly the cross-rendition sequence drift this PR is meant to eliminate.
2. **`PopExpiredMarkers` is skipped**, since it lives in the same block. A stranded marker keeps being counted by `GetMarkerCountBefore()`, inflating `GetEstimatedSequenceNumber()` for every marker inserted afterwards until the next normal completion cleans it up.

The "no IDR for 2x target" precondition is less exotic than it looks: `kShouldDeferSegmentFlush` deliberately postpones completion to wait for a marker, and `return_cut` waits for an IDR on video. It is most likely under the exact misconfiguration the new warning in `SendDataFrame` describes — which only warns and keeps running.

Suggestion: make `last_chunk` a `bool &` so the forced completion propagates back. Checking `IsCompleted()` after the call won't work, since `AppendMediaChunk` already called `CreateNextSegment()`.

src/base/modules/marker/marker_box.cpp
```diff
@@ -353,9 +369,18 @@ std::tuple<bool, ov::String> MarkerBox::CanInsertMarker(const std::shared_ptr<Ma
 		// IN -> IN
 		if (last_out_of_network_value == false)
 		{
-			if (_last_inserted_marker->GetTimestamp() > marker->GetTimestamp())
+			bool last_still_pending = _markers_by_timestamp.find(_last_inserted_marker->GetTimestamp()) != _markers_by_timestamp.end();
+
+			if (_last_inserted_marker->GetTimestamp() >= marker->GetTimestamp())
```
Keukhan: **`last_still_pending` is computed but not applied here, so a duplicate IN is accepted after the pending one was already consumed**

`_last_inserted_marker` is only assigned on a successful insert (`marker_box.cpp:439`, `marker_box.cpp:480`); no pop or remove path resets it. So it means "the last marker ever inserted", not "the marker currently pending" — it keeps pointing at a marker already emitted into a playlist. That is what `last_still_pending` exists to distinguish, but it is only consulted in the `else if` branch.

The comment says the same timestamp "replaces it harmlessly". That holds only while the marker is still in `_markers_by_timestamp`; once it is gone, this is a fresh insert, not a replacement:

```
t=0s     OUT@10s (duration 20s) + provisional IN@30s
t=10s    OUT consumed/emitted
t=30s    pIN@30s consumed/emitted   -> box {}, but _last_inserted is still pIN@30s
t=30.2s  the confirming IN@30s arrives late (retry, or sent at the actual return point)
         last_still_pending = false (computed, unused);  30s >= 30s -> accepted
         both erases in InsertMarker are no-ops
         -> box {IN@30s}   <-- a second IN with no OUT in between
```

The parent is inherited correctly, so this stray marker renders as a fully valid tag — nothing downstream rejects it. Outcome depends on the write head:

- **Already past 30s** -> no cut, `PopExpiredMarkers` drops it with a warning. Harmless.
- **Still just short of 30s** (up to one chunk on audio, one GOP on video) -> treated as a real marker: an extra segment boundary and a second `EXT-X-CUE-IN` / `SCTE35-IN` with no OUT in between.

The worst case is when those two outcomes split across tracks — audio already past it, video not — giving video +1 boundary and audio none, i.e. permanent sequence drift between renditions.

Note this was previously rejected: the old condition was `>`, so an identical timestamp fell through to `"IN marker only can be modified with the less timestamp"`. Widening it to `>=` allows behaviour that is only safe while the marker is pending. The tests don't cover it because they insert OUT / provisional IN / explicit IN at the same media time, so the marker is always still pending.

Suggestion:

```cpp
if (last_still_pending == true && _last_inserted_marker->GetTimestamp() >= marker->GetTimestamp())
```

An IN arriving after the previous one was emitted is an IN with no open break. This also makes the erase at `marker_box.cpp:456` meaningful only in the pending state. (The pre-existing `>` case has the same weakness; the same guard covers both.)

Keukhan (commented):
Two findings below, both the same pattern: state that outlives the thing it describes and is then used as the basis for a decision. Both lead to the cross-rendition media-sequence drift this PR is trying to remove.

src/base/modules/marker/marker_box.cpp
getroot: good, applied

src/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
getroot: good, applied

Keukhan (approved):
lgtm

PR #2284, opened 2026-08-05
fix(transcoder): preserve frames and keyframe cadence across track changes

getroot:
## Problem

On a scheduled channel rotating items with different resolutions, video segments come out as 4.033s or 4.067s instead of a constant 4.000s, and the video keyframe positions drift by one frame per item transition.

With `KeyFrameIntervalType` set to `time`, every CUE/SCTE-35 marker is rejected with "Duration of the marker must be greater than 50000ms", and the event insertion lead grows to tens of seconds.

## Solution

- The FPS filter parks the last frame until its successor arrives. When an input format change replaces the filter, that frame was silently dropped and the new instance lost the output slot position, leaving a one-frame hole per transition. The outgoing filter now flushes the parked frame downstream, and the new instance inherits the slot position so a frame lost around the swap is refilled.
- A reinitialized codec session opens with an immediate keyframe, which permanently shifts the keyframe cadence. On a track change the encoder is now armed to force one keyframe where the previous cadence would have produced one: FRAME mode mirrors the codec's own frame counting, TIME mode projects from the last output keyframe timestamp.
- The stream-level special case that tore down and recreated a specific hardware encoder on every track change is removed. It discarded the frames still queued inside the encoder each time; an encoder that must reinitialize its codec session on an input change detects it per frame on its own thread (`NeedReinitForFrame`), which flushes the codec downstream and continues with the same queue.
- `MediaTrack::GetKeyframeIntervalDurationMs()` treated a TIME-mode interval, which is already milliseconds, as a frame count. It is now interval-type aware, fixing the marker gap validation, the event insertion lead, and the alignment warning at once. The duplicated conversion inside the marker box was replaced with this helper.
- Removed `FMP4MarkerSyncTest.ForcedCompletionSettlesPendingMarkerCut`: after a forced completion the adaptive pacing target drops below zero and the next keyframe cuts a segment either way, so the scenario it asserts is not observable from the outside.

## Test

- Unit: 11 tests added (`FilterFpsTest` x5, `KeyframeGridRestoreTest` x5, `TimeModeKeyframeIntervalGapValidation`). The gap validation test was confirmed failing before the fix and passing after. Full suite passes except the pre-existing `FilterLavfiResampler.MismatchedFrameDoesNotDisableFilter` failure, which is unrelated to this change.
- E2E: a scheduled channel rotating two 10s VOD items (1080p and 360p sources) encoded with x264 1080p30, keyint 30, LLHLS 4s segments, run for 95 seconds. Pass condition: every video segment is exactly 4.000s, the output keyframe interval map is perfectly regular, and exactly one cadence restore fires per item transition. Before the fix, segments alternate 4.033/4.067 and one frame is lost per transition.

### DIFF
--- a/src/base/info/media_track.cpp
+++ b/src/base/info/media_track.cpp
@@ -834,6 +834,13 @@ double MediaTrack::GetKeyFrameInterval() const
 
 double MediaTrack::GetKeyframeIntervalDurationMs() const
 {
+	// A TIME interval is already milliseconds, and it only ever comes from the
+	// configuration; the resolved getter falls back to a measured frame count
+	if (GetKeyFrameIntervalTypeByConfig() == cmn::KeyFrameIntervalType::TIME)
+	{
+		return GetKeyFrameIntervalByConfig();
+	}
+
 	double keyframe_interval = std::ceil(GetKeyFrameInterval());
 	double framerate = std::ceil(GetFrameRate());
 
--- a/src/base/modules/marker/marker_box.cpp
+++ b/src/base/modules/marker/marker_box.cpp
@@ -709,13 +709,9 @@ double MarkerBox::GetActualTargetSegmentDurationMs() const
 
 	// Video can only cut at keyframes, so its real cadence is the target rounded
 	// up to the keyframe interval; the interval may be unknown (0) at this point
-	if (segment_info->media_type == cmn::MediaType::Video && segment_info->framerate > 0)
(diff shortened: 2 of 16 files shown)


Keukhan (approved):
There's nothing I'm particularly worried about. I do have some concern about the process of sending all the frames flushed from the FPS filter over to the Scaler filter for its flush, but I think we'll need to test that once this feature is developed. When merging into Enterprise, please make sure the changes are applied to the NVIDIA and NETINT codecs as well.

LGTM

PR #2285, opened 2026-08-05
fix(alert): send an alert when decoder recreation fails 

Keukhan:
## Summary

`RecreateDecoderForCodecChange()` failed to send an alert when the replacement
decoder could not be created — only logged, unlike every other decoder-creation
failure path. This adds the missing `EGRESS_STREAM_CREATION_FAILED_DECODER` alert
so the failure is reported consistently.

## Changes

- `src/transcoder/transcoder_stream.cpp`: on `CreateDecoder()` failure in
  `RecreateDecoderForCodecChange()`, call `TranscoderAlerts::UpdateErrorWithoutCount()`
  with `CREATION_ERROR_DECODER`, mirroring `CreateDecoders()`.

## Testing

1. Configure a Scheduler Provider with a schedule made of files using different
   codecs.
2. Force a decoder init failure with the Fault Injection API:

   ```
   POST /v2/internals/tests:setFaultInject
   {
     "transcoder": [
       {
         "targetModule": "default",
         "targetDeviceId": 0,
         "targetComponent": "decoder",
         "faultType": "INITFAILED",
         "faultRate": 100
       }
     ]
   }
   ```

3. When the schedule advances to an item whose input track has a different codec,
   `RecreateDecoderForCodecChange()` runs, and `CreateDecoder()` fails due to the
   injected fault.
4. Verify:
   - the log shows `Failed to recreate decoder for the changed codec`
   - the Alert webhook receives `EGRESS_STREAM_CREATION_FAILED_DECODER`
5. Clear the fault injection with `{"transcoder": []}` and reproduce the same
   codec change: the decoder is recreated successfully (`Decoder has been
   recreated for the changed codec`) and no alert is sent.

### DIFF
--- a/src/transcoder/transcoder_stream.cpp
+++ b/src/transcoder/transcoder_stream.cpp
@@ -1682,6 +1682,17 @@ void TranscoderStream::RecreateDecoderForCodecChange(MediaTrackId track_id, cons
 	{
 		logte("%s Failed to recreate decoder for the changed codec. Id(%d), InputTrack(%d), Codec(%s)",
 			  _log_prefix.CStr(), decoder_id.value(), track_id, cmn::GetCodecIdString(input_track->GetCodecId()));
+
+#if NOTIFICATION_ENABLED
+		TranscoderAlerts::UpdateErrorWithoutCount(
+			TranscoderAlerts::ErrorType::CREATION_ERROR_DECODER,
+			nullptr,
+			_input_stream,
+			input_track,
+			nullptr,
+			nullptr);
+#endif
+
 		return;
 	}
 
--- a/src/transcoder/transcoder_stream.cpp
+++ b/src/transcoder/transcoder_stream.cpp
@@ -1680,8 +1680,9 @@ void TranscoderStream::RecreateDecoderForCodecChange(MediaTrackId track_id, cons
 
 	if (CreateDecoder(decoder_id.value(), _input_stream, input_track) == false)
 	{
-		logte("%s Failed to recreate decoder for the changed codec. Id(%d), InputTrack(%d), Codec(%s)",
-			  _log_prefix.CStr(), decoder_id.value(), track_id, cmn::GetCodecIdString(input_track->GetCodecId()));
+		logte("%s Failed to recreate decoder for the changed codec. Id(%d)<Codec(%s), Module(%s), Device(%u)>, InputTrack(%d)",
+			  _log_prefix.CStr(), decoder_id.value(), cmn::GetCodecIdString(input_track->GetCodecId()),
+			  cmn::GetCodecModuleIdString(input_track->GetCodecModuleId()), input_track->GetCodecDeviceId(), track_id);
 
 #if NOTIFICATION_ENABLED
 		TranscoderAlerts::UpdateErrorWithoutCount(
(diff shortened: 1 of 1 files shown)


src/transcoder/transcoder_stream.cpp
```diff
@@ -1682,6 +1682,17 @@ void TranscoderStream::RecreateDecoderForCodecChange(MediaTrackId track_id, cons
 	{
 		logte("%s Failed to recreate decoder for the changed codec. Id(%d), InputTrack(%d), Codec(%s)",
 			  _log_prefix.CStr(), decoder_id.value(), track_id, cmn::GetCodecIdString(input_track->GetCodecId()));
+
+#if NOTIFICATION_ENABLED
+		TranscoderAlerts::UpdateErrorWithoutCount(
```
getroot, line 1688: Since `stream_metric` is nullptr here, `Alert::IsAlertNeeded()` keys the dedup on the input stream URI alone, and a second identical `EGRESS_STREAM_CREATION_FAILED_DECODER` for the same stream is suppressed. Repeated codec-change failures on one stream therefore report only the first one.

That matches `CreateDecoders()`, so nothing is wrong here, but it does mean step 5 of the test plan cannot tell recovery apart from suppression. Running the successful codec change first and injecting the fault afterwards makes the first alert observable for sure.

src/transcoder/transcoder_stream.cpp
```diff
@@ -1682,6 +1682,17 @@ void TranscoderStream::RecreateDecoderForCodecChange(MediaTrackId track_id, cons
...
+			TranscoderAlerts::ErrorType::CREATION_ERROR_DECODER,
+			nullptr,
+			_input_stream,
+			input_track,
+			nullptr,
+			nullptr);
+#endif
+
 		return;
```
getroot, line 1697: Returning here leaves the track with no decoder at all, so every following packet hits `logte("Could not found decoder")` in `DecodePacket()`. That is one line per packet, tens of lines per second at the input frame rate.

One alert against an unbounded log stream is an odd asymmetry for a change whose goal is observability. Out of scope for this PR, but rate limiting that log or tearing the track down looks worth a follow-up.

src/transcoder/transcoder_stream.cpp
```diff
@@ -1682,6 +1682,17 @@ void TranscoderStream::RecreateDecoderForCodecChange(MediaTrackId track_id, cons
 	{
 		logte("%s Failed to recreate decoder for the changed codec. Id(%d), InputTrack(%d), Codec(%s)",
```
getroot: The failure log in `CreateDecoders()` carries `Codec, Module, Device`, while this one carries only the codec. The alert payload includes codec_module, but the log does not, so the log alone does not say which module or device failed. Since this PR is about making the two paths consistent, matching the log arguments would help too.

getroot (commented):
The alert call is correct and consistent with `CreateDecoders()`: `TranscoderStream` inherits `TranscoderAlerts` so the qualified call resolves to a member call, `input_track` is already updated to the new codec so the codec_module lookup reflects what actually failed, and the alert path only enqueues, so it does not block the packet thread this function runs on. Three notes inline.

src/transcoder/transcoder_stream.cpp
Keukhan: Not suppressed. `SendStreamMessage()` calls `IsAlertNeeded()` but never `PutVerifiedMessages()` — that is only called by `MetricWorkerThread` for INTERNAL_QUEUE and INGRESS (alert.cpp:236, 265). The `EGRESS:<uri>` key is never stored, so `IsAlertNeeded()` is always true and every failure alerts. Test plan step 5 stands as written.

src/transcoder/transcoder_stream.cpp
Keukhan: Confirmed. It also does not self-heal — `_last_input_tracks[track_id]` is updated at line 1623 before the recreate attempt, so it will not retry until the next codec change. Rather than patching this one spot, log flooding should be addressed across the transcoder in a separate PR.

src/transcoder/transcoder_stream.cpp
Keukhan: Fixed in b8c334ca — both logs now carry `<Codec, Module, Device>` like `CreateDecoders()`, which also matches what the alert payload reports.

getroot (approved):
LGTM

PR #2287, opened 2026-08-06
fix(transcoder): recover delayed frames and keep slot continuity at track changes

getroot:
### Problem

At a resolution change the decoder was reinitialized without draining, losing the frames still delayed inside the codec session. Recovering them exposes a second defect: a scheduled item can start less than one frame after the previous one ends, and the fps filter re-anchored backward onto an already emitted slot, duplicating output frames and shifting the keyframe cadence and segment lengths. A randomly assigned filter source id could also collide with the "no frame seen yet" sentinel the encoders use for reinit detection.

### Solution

The decoder drains its delayed frames downstream before reinitializing. The fps filter keeps its inherited slot position over a small backward overlap, letting the queue rules discard the stale slot, and re-anchors only on a real discontinuity. A filter is never assigned source id 0, and the input discontinuity diagnostic moved to trace because rate conversion triggers it structurally.

### Test

- Unit: ome_test_transcoder 39/39, including a new FilterFps test for the backward overlap discard.
- e2e: a scheduled channel rotating two VODs (640x360/1080p, 10 s items, x264), 120 s steady state with 14 transitions. Pass conditions: every LLHLS segment exactly 4.000 s, zero out of order DTS warnings, and the frame-level trace showing no lost slot at any transition (previously every transition lost one decoder-delayed frame), with the recovered frames' slot overlaps (4 occurrences) discarded without duplicate emission.

### DIFF
--- a/src/transcoder/codec/decoder/decoder_avcodec_video.cpp
+++ b/src/transcoder/codec/decoder/decoder_avcodec_video.cpp
@@ -73,6 +73,28 @@ bool AVCodecVideoDecoder::ReinitCodecIfNeed()
 			  _codec.GetWidth(), _codec.GetHeight(),
 			  _analyzer.GetWidth(), _analyzer.GetHeight());
 
+		// Drain the frames still delayed inside the codec session and hand them
+		// downstream before teardown, so the resolution change loses no frame.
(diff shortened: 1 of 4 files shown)


dimiden (approved):
LGTM

PR #2288, opened 2026-08-07
fix: include the downstream handoff wait in the SkipFrames load metric

Keukhan:
## Summary

`SkipFrames` sized itself from the time spent inside the filter graph alone. But the filter and the encoder are decoupled by a queue, so a saturated encoder grew that queue without ever touching the filter's own timing - the metric stayed flat at ~250us while input FPS collapsed from 30 to 10, and `SkipFrames` never left 0. The overload it exists to relieve went unhandled.

The filter thread does stall on the encoder. The encoder input queue has a threshold of 2 with exceed-wait enabled, so the handoff inside `OnComplete()` blocks until the encoder drains. That wait simply fell outside the measurement bracket, which closes as soon as `ReceiveFrame()` returns a frame.

## Changes

The load metric becomes a **per-frame budget** - the total time the filter thread is occupied per output frame:

```
frame_budget = filtering time + handoff wait
max_fps      = 1000000 / frame_budget
```

- The thread cannot outrun the sum of the two, so the budget is the real   throughput ceiling. A slow rescaler and a backed-up encoder now throttle   `SkipFrames` through one metric; previously only the former could.
- The skip formula, the 90% safety margin, the 1s evaluation interval and the 5s   recovery hold are unchanged - only the input to the formula got wider. 
- The two halves are averaged separately with the same EMA alpha so the log  attributes the bottleneck, but only the total drives the decision.
- Time measured on a round that produces no frame stays pending and is carried to  whichever frame completes next; on the error paths it is dropped rather than  pinned on an unrelated later frame.

## How to test

Inject lag into the encoder and watch the filter react. Each triggered frame sleeps 300ms inside the encoder, so a 20% rate on a 30fps stream pushes the encoder down to roughly 10fps.

**1.** Enable automatic `SkipFrames` on the video encode:

```xml
<Encodes>
    <Video>
        <SkipFrames>0</SkipFrames>  <!-- 0 = auto, -1 = disabled, 1~120 = fixed -->
        <Framerate>30</Framerate>
        ...
    </Video>
</Encodes>
```

**2.** Start a 30fps stream, then inject the fault
(`targetModule` is the encoder module: `default`, `nv`, `xma`, `nilogan`):

```bash
curl -u "$TOKEN:" -X POST \
  "http://<host>:8081/v2/internals/tests:setFaultInject" \
  -H 'Content-Type: application/json' \
  -d '{
        "transcoder": [
          {
            "targetComponent": "encoder",
            "targetModule": "nv",
            "targetDeviceId": 0,
            "faultType": "LAGGING",
            "faultRate": 20
          }
        ]
      }'
```

**3.** Watch the filter log. Before this change the budget stayed flat and
`SkipFrames` never moved:

```
SkipFrames evaluation. Input FPS: 10.66, Weighted Avg Time: 243.64us, MaxIeal: 3694.05
SkipFrames evaluation. Input FPS: 13.06, Weighted Avg Time: 254.15us, MaxIeal: 3541.27
```

Now the handoff wait dominates the budget and the skip count climbs:

```
Changed SkipFrames() 0 -> 1 (Bottleneck). Output FPS: ..., Frame Budget: 60000.00us (Inproc: 250.00us + Handoff: 59750.00us = Possible FPS: 15.00)
```

**4.** Clear the fault and confirm recovery - `SkipFrames` should decay back to 0
one step per 5s recovery hold:

### DIFF
--- a/src/transcoder/filter/filter_video_base.cpp
+++ b/src/transcoder/filter/filter_video_base.cpp
@@ -78,7 +78,7 @@ FilterResult FilterVideoBase::PopCompletedFrameInternal()
 {
 	if (GetState() == FilterBase::State::ERROR)
 	{
-		_pending_processing_time_us = 0;
+		ClearPendingTime();
 
 		return FilterResult::Error("The filter is in the error state");
 	}
@@ -90,15 +90,14 @@ FilterResult FilterVideoBase::PopCompletedFrameInternal()
 		auto completed_frame = ReceiveFrame();
 		if (completed_frame != nullptr)
 		{
-			UpdateProcessingTimePerFrame(start_time);
+			CommitPendingTime(start_time);
 
 			return FilterResult::Ready(std::move(completed_frame));
 		}
 
 		if (GetState() == FilterBase::State::ERROR)
 		{
-			// The measured time belongs to work that will never complete.
-			_pending_processing_time_us = 0;
+			ClearPendingTime();
 
 			return FilterResult::Error("The backend rescaler has failed");
 		}
@@ -107,8 +106,7 @@ FilterResult FilterVideoBase::PopCompletedFrameInternal()
 		auto frame = _fps_filter.Pop();
 		if (frame == nullptr)
 		{
-			// No completed frame available yet. Update pending processing time
-			_pending_processing_time_us += ElapsedTimeInUs(start_time);
+			AddProcessingTime(start_time);
 
 #if _SKIP_FRAMES_ENABLED
 			// Update skip frames based on the current processing time and framerate.
@@ -181,19 +179,36 @@ void FilterVideoBase::InheritContinuity(const FilterBase *previous)
 	_fps_filter.SetContinuationPts(previous_video->_fps_filter.GetNextPts());
 }
 
-int64_t FilterVideoBase::ElapsedTimeInUs(const std::chrono::steady_clock::time_point &start_time) const
+void FilterVideoBase::AddHandoffTime(int64_t elapsed_us)
 {
-	return std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::steady_clock::now() - start_time).count();
+	// Called on the filter thread, so no synchronization is needed.
+	_pending_handoff_time_us += elapsed_us;
 }
 
-void FilterVideoBase::UpdateProcessingTimePerFrame(const std::chrono::steady_clock::time_point &start_time)
+void FilterVideoBase::CommitPendingTime(const std::chrono::steady_clock::time_point &start_time)
 {
 	static constexpr double kProcessingTimeEmaAlpha = 0.1;
 
-	auto elapsed_time_us = _pending_processing_time_us + ElapsedTimeInUs(start_time);
+	auto processing_time_us = _pending_processing_time_us + ov::Clock::GetElapsedMicroSecondsFromNow(start_time);
+	auto handoff_time_us	= _pending_handoff_time_us;
 
-	_pending_processing_time_us				= 0;
-	_weighted_avg_frame_processing_time_us	= (_weighted_avg_frame_processing_time_us * (1.0 - kProcessingTimeEmaAlpha)) + (elapsed_time_us * kProcessingTimeEmaAlpha);
+	ClearPendingTime();
+
+	_weighted_avg_frame_processing_time_us	= (_weighted_avg_frame_processing_time_us * (1.0 - kProcessingTimeEmaAlpha)) + (processing_time_us * kProcessingTimeEmaAlpha);
+	_weighted_avg_frame_handoff_time_us		= (_weighted_avg_frame_handoff_time_us * (1.0 - kProcessingTimeEmaAlpha)) + (handoff_time_us * kProcessingTimeEmaAlpha);
+}
+
+void FilterVideoBase::AddProcessingTime(const std::chrono::steady_clock::time_point &start_time)
+{
+	// No frame came out this round, so the time waits for one that does.
+	_pending_processing_time_us += ov::Clock::GetElapsedMicroSecondsFromNow(start_time);
+}
+
+void FilterVideoBase::ClearPendingTime()
+{
+	// Time spent on work that failed must not be pinned on a later frame.
+	_pending_processing_time_us = 0;
+	_pending_handoff_time_us	= 0;
 }
 
 #if _SKIP_FRAMES_ENABLED
@@ -240,31 +255,21 @@ void FilterVideoBase::UpdateSkipFrames()
 	}
 	_skip_frames_last_check_time = curr_time;
 
-	// Remain for debugging and future improvement for queue-based skip frame adjustment
-	// -----------------------------------------------------------------------------
-	// double actual_input_fps			   = _fps_filter.GetInputFramesPerSecond();
-	// double expected_input_fps		   = _fps_filter.GetInputFrameRate();
-
-	// double expected_output_fps		   = _fps_filter.GetExpectedOutputFramesPerSecond();
-
-	// int64_t queue_waiting_deviation_us = _input_buffer.GetWaitingTimeInUs();
-	// double expected_frame_interval_us  = (expected_input_fps > 0.0) ? (1000000.0 / expected_input_fps) : 0.0;
-	// bool is_queue_overload			   = (expected_frame_interval_us > 0.0) &&
-	// 						 (queue_waiting_deviation_us > expected_frame_interval_us * _SKIP_FRAMES_QUEUE_BACKLOG_RATIO);
-	// bool is_queue_stable = (expected_frame_interval_us > 0.0) &&
-	// 					   (queue_waiting_deviation_us < expected_frame_interval_us * _SKIP_FRAMES_QUEUE_RECOVERY_RATIO);
-
 	double fixed_output_fps			   = _fps_filter.GetOutputFrameRate();
 	double expected_output_fps		   = _fps_filter.GetExpectedOutputFramesPerSecond();
 	double actual_output_fps		   = _fps_filter.GetOutputFramesPerSecond();
 
-	if (_weighted_avg_frame_processing_time_us <= 0.0 || fixed_output_fps <= 0.0)
+	// A slow rescaler and a backed-up encoder each throttle this thread, and it
+	// cannot outrun their sum, so the skip decision uses the total.
+	double frame_budget_us = _weighted_avg_frame_processing_time_us + _weighted_avg_frame_handoff_time_us;
+
+	if (frame_budget_us <= 0.0 || fixed_output_fps <= 0.0)
 	{
 		return;
 	}
 
 	// Calculate the maximum possible frames per second.
-	double max_frames_per_second = (1000000.0 / _weighted_avg_frame_processing_time_us);
+	double max_frames_per_second = (1000000.0 / frame_budget_us);
 	// To ensure stability, set a margin and use OO% of the calculated maximum FPS.
 	double ideal_frames_per_second = max_frames_per_second * _SKIP_FRAMES_ENSURE_FPS_MARGIN_RATIO;
 
@@ -285,12 +290,15 @@ void FilterVideoBase::UpdateSkipFrames()
 		next_skip_frames = FilterFps::SkipFramesMin;
 	}
 
-	ov::String common_log = ov::String::FormatString("Possible FPS: %.2f/%.2f(ideal), Output FPS: %.2f/%.2f/%.2f", max_frames_per_second, ideal_frames_per_second, fixed_output_fps, expected_output_fps, actual_output_fps);
+	ov::String common_log = ov::String::FormatString("Output FPS: %.2f (Conf: %.2f, Expect: %.2f), Frame Budget: %.2fus (Inproc: %.2fus + Handoff: %.2fus = Possible FPS: %.2f)",
+													 actual_output_fps, fixed_output_fps, expected_output_fps,
+													 frame_budget_us, _weighted_avg_frame_processing_time_us, _weighted_avg_frame_handoff_time_us,
+													ideal_frames_per_second);
 
 	// Increase skip frames immediately when bottleneck occurs.
 	if (_skip_frames < next_skip_frames)
 	{
-		logtw("[%s] Changed SkipFrames %d -> %d (Bottleneck). %s", GetLogPrefix().CStr(), _skip_frames, next_skip_frames, common_log.CStr());
+		logtw("[%s] Changed SkipFrames() %d -> %d (Bottleneck). %s", GetLogPrefix().CStr(), _skip_frames, next_skip_frames, common_log.CStr());
 
 		_skip_frames = next_skip_frames;
 		_fps_filter.SetSkipFrames(_skip_frames);
--- a/src/transcoder/filter/filter_video_base.cpp
+++ b/src/transcoder/filter/filter_video_base.cpp
@@ -66,10 +66,16 @@ FilterResult FilterVideoBase::ProcessFrameInternal(const std::shared_ptr<MediaFr
 		return FilterResult::Error("Received frame is null");
 	}
 
+	auto start_time = std::chrono::steady_clock::now();
+
 	if (_fps_filter.Push(media_frame) == false)
 	{
 		logtw("[%s] Dropped a frame because the FPS filter is full.", GetLogPrefix().CStr());
 	}
+
+	// The push runs on the filter thread like everything else, so it counts toward how
+	// busy the thread was this window.
+	_window_busy_time_us += ov::Clock::GetElapsedMicroSecondsFromNow(start_time);
 
 	return FilterResult::NoOutput();
 }
@@ -166,7 +172,6 @@ void FilterVideoBase::InheritContinuity(const FilterBase *previous)
 	{
 		return;
 	}
-
 	// The slot position only transfers within the same output cadence
 	if (_fps_filter.GetOutputFrameRate() != previous_video->_fps_filter.GetOutputFrameRate())
 	{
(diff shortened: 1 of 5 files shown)


src/transcoder/filter/filter_video_base.cpp
```diff
@@ -240,31 +257,21 @@ void FilterVideoBase::UpdateSkipFrames()
...
-	// 					   (queue_waiting_deviation_us < expected_frame_interval_us * _SKIP_FRAMES_QUEUE_RECOVERY_RATIO);
-
 	double fixed_output_fps			   = _fps_filter.GetOutputFrameRate();
 	double expected_output_fps		   = _fps_filter.GetExpectedOutputFramesPerSecond();
 	double actual_output_fps		   = _fps_filter.GetOutputFramesPerSecond();
 
-	if (_weighted_avg_frame_processing_time_us <= 0.0 || fixed_output_fps <= 0.0)
+	// A slow rescaler and a backed-up encoder each throttle this thread, and it
+	// cannot outrun their sum, so the skip decision uses the total.
+	double frame_budget_us = _weighted_avg_frame_processing_time_us + _weighted_avg_frame_handoff_time_us;
```
Keukhan, line 339: **[1/6] The handoff wait is a backpressure signal, not a capacity signal - folding it into this budget makes a healthy encoder trip SkipFrames.**

`MAX_QUEUE_SIZE` is 2 (`transcoder_encoder.cpp:24`) and `TranscodeEncoder::SendBuffer()` enqueues with an exceed-wait, so this thread blocks in `OnComplete()` whenever the encoder already holds 2 frames. That happens on any encoder running near - but *within* - capacity. It means the queue is shallow, not that the encoder cannot keep up.

Failure scenario, 30 fps target:

1. `_weighted_avg_frame_handoff_time_us` converges toward one frame interval.
2. As soon as `processing + handoff > 0.9 * 33 ms = 30 ms`, `next_skip_frames = ceil(30 / ideal - 1) >= 1` -> output halves to 15 fps.
3. The encoder goes idle, the handoff EMA collapses, and the recovery branch restores `skip = 0` after the 5 s hold.
4. The encoder saturates again -> back to step 2. Output oscillates between 30 and 15 fps indefinitely.

Before this change only the filter's own CPU work (a few ms) fed the metric, so `skip` stayed at 0 unless the filter itself was the bottleneck.

Suggestions - any one, or a combination:

- Subtract the expected steady-state block (one frame interval) and charge only the excess, so normal backpressure contributes ~0.
- Drive the decision from the queue's own backlog instead, which is what the `_input_buffer.GetWaitingTimeInUs()` block removed in this commit was reaching for.
- Require N consecutive evaluation intervals above the threshold before raising `skip` (hysteresis), so a single saturated second cannot halve the output.

src/transcoder/filter/filter_video_base.cpp
```diff
@@ -181,19 +179,38 @@ void FilterVideoBase::InheritContinuity(const FilterBase *previous)
 	_fps_filter.SetContinuationPts(previous_video->_fps_filter.GetNextPts());
 }
 
-int64_t FilterVideoBase::ElapsedTimeInUs(const std::chrono::steady_clock::time_point &start_time) const
+void FilterVideoBase::AddHandoffTime(int64_t elapsed_us)
 {
-	return std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::steady_clock::now() - start_time).count();
+	// Called on the filter thread, so no synchronization is needed.
+	_pending_handoff_time_us += elapsed_us;
```
Keukhan, line 236: **[2/6] No clamp on the external wait: one transient stall latches a long output degradation.**

`elapsed_us` is whatever the downstream took, unbounded. `UpdateSkipFrames()` raises `skip` immediately but decays it only 20% per 5 s, and that asymmetry turns a hiccup into a sustained drop.

A single 1000 ms block in `OnComplete()` - `SendBuffer()`'s 1000 ms exceed-wait timing out, an encoder reinit, GPU contention, `TranscodeFaultInjector`'s 300 ms `Lagging` usleep - contributes `0.1 * 1e6 = 100,000 us` to the EMA on the very next commit:

- `frame_budget_us` becomes ~105,000 -> `ideal_frames_per_second` ~8.6
- at a 30 fps target, `next_skip_frames = ceil(30 / 8.6 - 1) = 3` -> output drops to 7.5 fps instantly
- getting back to 0 then takes several 5 s recovery holds, ~15-20 s of degraded output

Nothing bounds `elapsed_us`, and the pre-encode audio path (`TranscodeFilter::SendBuffer()`, `transcoder_filter.cpp:367`) uses an Infinite timeout, so it can block far longer still.

Suggestion: clamp each sample (e.g. 2-3 frame intervals), or use a median/percentile so a single outlier cannot move the metric - and rate-limit the increase the way recovery is already rate-limited.

src/transcoder/filter/filter_video_base.cpp
```diff
@@ -90,15 +90,14 @@ FilterResult FilterVideoBase::PopCompletedFrameInternal()
 		auto completed_frame = ReceiveFrame();
 		if (completed_frame != nullptr)
 		{
-			UpdateProcessingTimePerFrame(start_time);
+			CommitPendingTime(start_time);
```
Keukhan, line 99: **[3/6] `CommitPendingTime()` runs before `OnComplete()` for the same frame, so every handoff sample lands one frame late - and is discarded outright on error or filter swap.**

Order per drain iteration (`transcoder_filter.cpp:293-302`):

```
Pop -> DataReady -> CommitPendingTime()   // clears _pending_handoff_time_us
     -> OnComplete() blocks for h -> AddHandoffTime(h)
     -> Pop -> NoOutput -> UpdateSkipFrames()
```

So `UpdateSkipFrames()` always evaluates with `h` still pending and excluded from `_weighted_avg_frame_handoff_time_us`. At the onset of a stall the first and largest sample is invisible for an extra frame.

Worse, if the filter enters `ERROR` (lines 81 / 100) or is replaced by a reconfiguration before the next frame completes, `ClearPendingTime()` throws that sample away entirely - the exact stall this PR sets out to observe is never recorded.

Suggestion: the handoff measurement is already complete when `AddHandoffTime()` is called, so there is no reason to buffer it as pending. Updating the EMA directly in `AddHandoffTime()` and dropping `_pending_handoff_time_us` fixes both the one-frame lag and the loss on error/swap. A handoff happens exactly once per completed frame, so the per-frame meaning of the EMA is preserved.

src/transcoder/transcoder_filter.cpp
```diff
@@ -293,8 +293,14 @@ void TranscodeFilter::ThreadLoop()
 			auto recv = base->PopCompletedFrameInternal();
 			if (recv.result == TranscodeResult::DataReady)
 			{
+				// OnComplete() blocks here while the encoder queue is full, so the
+				// filter needs that wait to react to a backed-up encoder.
+				auto handoff_start = std::chrono::steady_clock::now();
+
 				OnComplete(recv.result, std::move(recv.frame));
 
+				base->AddHandoffTime(ov::Clock::GetElapsedMicroSecondsFromNow(handoff_start));
```
Keukhan, line 302: **[4/6] The two reconfiguration flush loops call `OnComplete()` without this timer, so the largest handoffs in the pipeline are never measured.**

On a format change or a `_setup_pending` swap, `ThreadLoop()` dumps every frame parked in the outgoing filter via `OnComplete(TranscodeResult::DataReady, ...)` in a loop - lines 233-236 and 259-262 - with no `AddHandoffTime()` around it. Against the 2-slot encoder queue that burst blocks for roughly one frame interval per frame past the second, potentially hundreds of milliseconds.

The goal stated in the comment just above ("the filter needs that wait to react to a backed-up encoder") is therefore unmet precisely on the path where the encoder is most likely to be backed up, and the wait is invisible to whichever filter instance is queried next.

Suggestion: wrap `OnComplete()` + `AddHandoffTime()` in a small helper and use it at all three call sites. Charging the pre-swap flush needs finding 5 as well, otherwise the sample dies with the outgoing instance.

src/transcoder/filter/filter_video_base.cpp
```diff
@@ -181,19 +179,38 @@ void FilterVideoBase::InheritContinuity(const FilterBase *previous)
 	_fps_filter.SetContinuationPts(previous_video->_fps_filter.GetNextPts());
```
Keukhan, line 184: **[5/6] The load state is dropped on every filter swap, and `InheritContinuity()` - the existing hook for exactly this - does not carry it.**

`TranscodeFilter::Initialize()` always constructs a fresh `FilterVideoBase`, so `_weighted_avg_frame_handoff_time_us`, `_weighted_avg_frame_processing_time_us` and `_skip_frames` all reset to their defaults.

After a resolution or pixel-format change while the encoder is backed up, the new filter starts at `skip = 0` and floods the still-full encoder queue at full rate, then has to re-derive the skip level from scratch over the next several evaluation intervals - with the 5 s / 20% recovery ratchet in between.

Suggestion: carry the two EMAs, `_skip_frames` and `_skip_frames_last_changed_time` here alongside the slot position. The hook and its same-cadence guard already exist; only the values are missing.

src/transcoder/filter/filter_video_base.cpp
```diff
@@ -107,8 +106,7 @@ FilterResult FilterVideoBase::PopCompletedFrameInternal()
 		auto frame = _fps_filter.Pop();
 		if (frame == nullptr)
 		{
-			// No completed frame available yet. Update pending processing time
-			_pending_processing_time_us += ElapsedTimeInUs(start_time);
+			AddProcessingTime(start_time);
```
Keukhan, line 115: **[6/6] Pre-existing, but this PR makes it load-bearing: the processing EMA grows with the current skip level, so the controller's input is a function of its own output.**

`FilterFps::Pop()` returns `nullptr` for skipped slots, so at `skip = 29` roughly 29 of every 30 `PopCompletedFrameInternal()` calls take this branch and accumulate their `ReceiveFrame()` + `Pop()` cost into `_pending_processing_time_us`; all of it is then committed onto the single frame that does complete.

`_weighted_avg_frame_processing_time_us` therefore rises monotonically with the skip level (~15% inflation at `skip = 29` for a typical rescale) -> `frame_budget_us` rises -> `next_skip_frames` rises. Combined with the new handoff term this is a positive-feedback path that works against the recovery branch.

Suggestion: normalize the budget per *input* frame handled and compare against the input rate, or divide the committed time by the number of rounds it accumulated over, so the metric no longer scales with the skip level it produced.

Keukhan (commented):
Self-review of the SkipFrames load metric change. The problem statement is right - the filter and the encoder are decoupled by a queue, so a saturated encoder never touched the filter's own timing and the metric stayed flat. The concern is with *how* the blocking handoff is folded into the metric.

Six findings inline, roughly in severity order:

| # | Location | Issue |
|---|---|---|
| 1 | `filter_video_base.cpp:266` | A healthy-but-saturated encoder trips SkipFrames -> sustained 30/15 fps limit cycle |
| 2 | `filter_video_base.cpp:185` | Unbounded wait sample: one 1 s stall -> 7.5 fps for ~15-20 s |
| 3 | `filter_video_base.cpp:93` | Handoff sample lands one frame late, and is lost on error / filter swap |
| 4 | `transcoder_filter.cpp:302` | Reconfiguration flush loops are not measured at all |
| 5 | `filter_video_base.cpp:179` | Load state (EMAs, `_skip_frames`) is not inherited across a swap |
| 6 | `filter_video_base.cpp:109` | Processing EMA scales with the skip level it produced (pre-existing, now load-bearing) |

2, 3, 4 and 5 are each a local fix. **1 is the blocker**: without separating normal backpressure from a real bottleneck, the remaining fixes still leave fps oscillating on a healthy pipeline.

The log and declaration inconsistencies found alongside these - the stray `()` in the bottleneck line that broke greps matching the three sibling SkipFrames lines, the `Inproc + Handoff = Possible FPS` equation that does not hold, `max_frames_per_second` no longer being surfaced, the interleaved access specifiers, and `GetElapsedMiliSecondsFromNow()` returning `uint64_t` next to the new `int64_t` sibling - are already addressed in 0674c85c.

src/transcoder/filter/filter_video_base.cpp
Keukhan: Addressed in cf2fa9e1.

The budget still measures `processing + handoff`, but raising the skip level now needs two independent conditions:

- `utilization >= 0.95` - the thread had no idle time left in the evaluation window. Idle time is spent in `_input_buffer.Dequeue()`, outside the filter, so accumulating only busy time (`_window_busy_time_us`) yields the ratio directly. A burst of input or a jittery encode blocks the handoff but still leaves the thread idle afterwards, so neither reads as a bottleneck any more.
- `actual_output_fps < expected_output_fps * 0.9` - it is missing the rate the current skip level asks for. `GetExpectedOutputFramesPerSecond()` was already being computed and only logged; folding it into the decision also makes the level converge, because each raise lowers the expected rate until the two meet.

The FPS filter push is counted as well, so utilization is not understated.

Stepping down got a matching gate at `utilization < 0.80`. Without it the boundary case still oscillates: an encoder with 14 fps of capacity against a 30 fps target sits at ~93% utilization at `skip = 1`, which the raise gate reads as "not saturated", so it would walk back to 0 and immediately be a bottleneck again. 0.80-0.95 is a hold band now.

Both thresholds are tuning values - the shared log line reports `Busy: %` so they can be checked against a real workload.

src/transcoder/filter/filter_video_base.cpp
Keukhan: Addressed in cf2fa9e1, by bounding the response rather than by replacing the EMA.

- A single handoff sample contributes at most 4 output frame intervals (133 ms at 30 fps), logged at debug when it clamps.
- The level rises by at most one step per evaluation interval.
- It rises only after two consecutive windows repeat the diagnosis. That is what actually kills the scenario you described: with alpha 0.1 a 100,000 us contribution decays to ~4% within one second (`0.9^30`), so a lone stall is already gone by the second window.

The clamp deliberately runs after `_window_busy_time_us` and before `_pending_handoff_time_us`, since utilization has to reflect the wall clock the thread really lost.

Not switching to a median/percentile: it needs a rolling buffer, and the clamp plus the hysteresis covers the same failure. The tradeoff is that sustained overload past 4 frame intervals per handoff under-corrects - capacity below ~1/4 of the target reads as exactly 1/4. That is a misconfigured track rather than a tuning case, and the encoder queue stays full either way.

src/transcoder/filter/filter_video_base.cpp
Keukhan: Not addressed - deliberately left out of scope for this PR.

The ordering stands: `UpdateSkipFrames()` still evaluates with the newest handoff sample pending, and `ClearPendingTime()` still discards it on error or filter swap. Restructuring the commit path was judged more churn than the symptom is worth, and the symptom is bounded - a raise can lag by one evaluation interval (1 s).

The utilization gate is unaffected. `_window_busy_time_us` accumulates inside `AddHandoffTime()` as the sample is measured, not through the pending path, so the gate never sees the lag.

src/transcoder/transcoder_filter.cpp
Keukhan: Not addressed - deliberately left out of scope for this PR.

The two flush loops still call `OnComplete()` untimed, so a reconfiguration burst never reaches `AddHandoffTime()` and that window's utilization reads lower than it really was, which can leave the gate closed.

Bounded in practice: reconfigurations are rare, and a raise needs two consecutive windows to agree, so one under-measured window does not change the outcome on its own.

src/transcoder/filter/filter_video_base.cpp
Keukhan: Addressed in cf2fa9e1.

`InheritContinuity()` now carries both averages, `_skip_frames`, the bottleneck counter and `_skip_frames_last_changed_time` alongside the slot position. `Initialize()` runs before `InheritContinuity()`, so the config-derived values are already in place, and the automatic level only transfers when both sides are in automatic mode (`_skip_frames_conf == 0`).

All of it sits under the existing same-cadence guard, since the skip level counts frames within a cadence and means nothing across a change of one.

src/transcoder/filter/filter_video_base.cpp
Keukhan: Not addressed - pre-existing, and left as its own change.

`AddProcessingTime()` still charges every no-output round to the frame that eventually completes, so the processing average still scales with the skip level that produced it.

The utilization gate contains the feedback path without fixing it: an inflated budget can still raise `next_skip_frames`, but it cannot act unless the thread is genuinely out of idle time and missing its current target, and both of those are measured independently of the skip level. Normalizing the budget per input frame is the actual fix.

Keukhan:
`cf2fa9e1` addresses findings 1, 2 and 5. Replies are on each thread; the short version:

| # | Status | What changed |
|---|---|---|
| 1 | fixed | Raising the skip level now needs the thread to have no idle time left in the window (`utilization >= 0.95`) **and** the output to fall short of the rate the current level asks for. Stepping down needs `utilization < 0.80`, making 0.80-0.95 a hold band. |
| 2 | fixed | A single handoff contributes at most 4 frame intervals, the level rises one step per interval, and only after two consecutive windows agree. |
| 3 | open | Handoff sample still lands one interval late and is dropped on error/swap. Costs up to 1 s of reaction time; the utilization gate does not go through the pending path, so it is unaffected. |
| 4 | open | Reconfiguration flush bursts stay unmeasured, so that window's utilization reads low. Rare, and a raise needs two windows to agree. |
| 5 | fixed | `InheritContinuity()` carries both averages, the skip level, the bottleneck counter and the last-changed time. |
| 6 | open | Pre-existing. The processing average still scales with the skip level; the gate contains the feedback path but does not remove it. |

3, 4 and 6 were left deliberately - 3 and 4 as churn out of proportion to the symptom, 6 as a pre-existing defect worth its own change.

Not yet verified against a live stream. `_SKIP_FRAMES_SATURATION_RATIO` (0.95), `_SKIP_FRAMES_RECOVERY_BUSY_RATIO` (0.80) and the clamp multiplier (4) are tuning values picked from the arithmetic, not from measurement - the shared log line now reports `Busy: %` so they can be checked on a real workload.

PR #2289, opened 2026-08-07
refactor(transcoder): restructure SkipFrames control and improve its accuracy

Keukhan:
## Summary

SkipFrames did not react when the encoder was the bottleneck: it only measured the filter's own work, so the level stayed at 0 while frames piled up in the decoder queue.
It now measures the wait on the encoder too, and checks itself against the input rate instead of the output rate it aims at.
The decision logic moves into its own file.

## Changes

**Load is measured as a per-frame budget.**

```
budget  = time filtering + time waiting on the handoff
max fps = 1000000 / budget
```

The filter thread blocks on the handoff when the encoder queue is full, so a slow encoder and a slow rescaler now throttle SkipFrames through the same number.

**The level is checked against the input rate.**
Skipping lowers the output target, so checking against that let the filter keep meeting a bar it had just lowered while the queue grew.

**Busy alone does not raise the level.**
The encoder queue is only two frames deep, so the handoff blocks even when the chain keeps up.
The level rises only when the thread is busy *and* losing input, and only after a second window agrees.

**Recovery is a probe.**
After 5s at one level it steps down 20%.
If that does not hold, the next window puts it back.

## Testing

Slow the encoder down and watch the filter react.
Each triggered frame sleeps 300ms, so 20% on a 30fps stream drops the encoder to about 10fps.

**1.** Turn on automatic SkipFrames:

```xml
<Encodes>
    <Video>
        <SkipFrames>0</SkipFrames>  <!-- 0 = auto, -1 = off, 1~120 = fixed -->
        <Framerate>30</Framerate>
    </Video>
</Encodes>
```

**2.** Start a 30fps stream, then inject the fault:

```bash
curl -u "$TOKEN:" -X POST \
  "http://<host>:8081/v2/internals/tests:setFaultInject" \
  -H 'Content-Type: application/json' \
  -d '{"transcoder":[{"targetComponent":"encoder","targetModule":"nv",
                      "targetDeviceId":0,"faultType":"LAGGING","faultRate":20}]}'
```

**3.** Watch the filter log. It should climb until the thread keeps up:

```
SkipFrames 0 -> 1, pipeline cannot keep up. Input: 10.7/30.0fps, Busy: 99.5%, Budget: 60.00ms (proc 0.25 + handoff 59.75), Capacity: 15.0fps, ...
SkipFrames 1 -> 2, ...
SkipFrames 2, steady. Input: 29.8/30.0fps, Busy: 90.1%, ...
```

Two things to check: `Input` returns to ~30/30fps, and the `dec_H264_t0` queue stops reporting `Exceeded`.

**4.** Clear the fault. SkipFrames should walk back to 0, one step per 5s:

```bash
curl -u "$TOKEN:" -X POST \
  "http://<host>:8081/v2/internals/tests:setFaultInject" \
  -H 'Content-Type: application/json' -d '{"transcoder":[]}'
```

### DIFF
--- a/src/transcoder/filter/filter_skipframes_controller.cpp
+++ b/src/transcoder/filter/filter_skipframes_controller.cpp
@@ -176,7 +176,9 @@ std::optional<SkipFramesController::Result> SkipFramesController::Evaluate(int64
 	{
 		next_skip_frames = static_cast<int32_t>(std::floor(observation.max_output_fps - 1));
 	}
-	else if (next_skip_frames < FilterFps::SkipFramesMin)
+
+	// Prevent the level from going below the minimum, which is also the disabled value.
+	if (next_skip_frames < FilterFps::SkipFramesMin)
 	{
 		next_skip_frames = FilterFps::SkipFramesMin;
 	}
--- a/src/transcoder/filter/filter_skipframes_controller.cpp
+++ b/src/transcoder/filter/filter_skipframes_controller.cpp
@@ -133,22 +133,22 @@ std::optional<SkipFramesController::Result> SkipFramesController::Evaluate(int64
 
 	// A slow rescaler and a backed-up encoder both throttle the thread, and it cannot outrun
 	// their sum, so the decision uses the total.
-	double frame_budget_us = _weighted_avg_frame_processing_time_us + _weighted_avg_frame_handoff_time_us;
+	double frame_cost_us = _weighted_avg_frame_processing_time_us + _weighted_avg_frame_handoff_time_us;
 
-	if (frame_budget_us <= 0.0 || observation.cadence_fps <= 0.0)
+	if (frame_cost_us <= 0.0 || observation.max_output_fps <= 0.0)
 	{
 		return std::nullopt;
 	}
 
-	// The most frames per second the budget allows, less a safety margin.
-	double peak_fps		   = (1000000.0 / frame_budget_us);
+	// The most frames per second the cost allows, less a safety margin.
+	double peak_fps		   = (1000000.0 / frame_cost_us);
 	double sustainable_fps = peak_fps * kSafetyMarginRatio;
 
-	// How far the cadence has to be divided down to land on the sustainable rate.
-	auto next_skip_frames = static_cast<int32_t>(std::ceil(observation.cadence_fps / sustainable_fps - 1.0));
-	if (next_skip_frames > observation.cadence_fps - 1)
+	// How far the output ceiling has to be divided down to land on the sustainable rate.
+	auto next_skip_frames = static_cast<int32_t>(std::ceil(observation.max_output_fps / sustainable_fps - 1.0));
+	if (next_skip_frames > observation.max_output_fps - 1)
 	{
-		next_skip_frames = static_cast<int32_t>(std::floor(observation.cadence_fps - 1));
+		next_skip_frames = static_cast<int32_t>(std::floor(observation.max_output_fps - 1));
 	}
 	else if (next_skip_frames < FilterFps::SkipFramesMin)
 	{
@@ -157,19 +157,19 @@ std::optional<SkipFramesController::Result> SkipFramesController::Evaluate(int64
 
 	// Busy is not the same as overloaded. The encoder queue is only two frames deep, so the
 	// handoff blocks even on a chain that keeps up. Idle time is what tells the two apart.
-	bool is_saturated = (utilization >= kSaturationRatio);
+	bool is_fully_busy = (utilization >= kSaturationRatio);
 
-	// A saturated thread also has to be losing ground. Zero means the FPS filter has not
+	// A fully busy thread also has to be losing input. Zero means the FPS filter has not
 	// measured a second yet, not that nothing was consumed.
-	bool is_falling_behind = (observation.expected_input_fps > 0.0) &&
-							 (observation.actual_input_fps > 0.0) &&
-							 (observation.actual_input_fps < observation.expected_input_fps * kInputKeepUpRatio);
+	bool is_losing_input = (observation.expected_input_fps > 0.0) &&
+						   (observation.actual_input_fps > 0.0) &&
+						   (observation.actual_input_fps < observation.expected_input_fps * kInputKeepUpRatio);
 
-	if (is_saturated == true && is_falling_behind == true)
+	if (is_fully_busy == true && is_losing_input == true)
 	{
 		_bottleneck_count++;
 
-		// The budget decays between stalls, so without this the recovery branch below could
+		// The cost decays between stalls, so without this the recovery branch below could
 		// walk the level down while the bottleneck is still there.
 		next_skip_frames = std::max(next_skip_frames, _skip_frames);
 	}
@@ -181,9 +181,13 @@ std::optional<SkipFramesController::Result> SkipFramesController::Evaluate(int64
 		// busy; gating on utilization alone would pin a level reached during a spike
 		// forever. The hold interval makes this a probe - if the level below cannot hold,
 		// the next window says so and puts it back.
-		bool has_headroom = (utilization < kRecoveryMaxBusyRatio) || (is_falling_behind == false);
+		bool may_step_down = (utilization < kRecoveryMaxBusyRatio) || (is_losing_input == false);
 
-		next_skip_frames = has_headroom ? FilterFps::SkipFramesMin : _skip_frames;
+		// The step down lands where the cost allows, not at nothing.
+		if (may_step_down == false)
+		{
+			next_skip_frames = _skip_frames;
+		}
 	}
 
 	if (next_skip_frames > _skip_frames)
@@ -203,16 +207,23 @@ std::optional<SkipFramesController::Result> SkipFramesController::Evaluate(int64
 	// Every value labelled, every pair actual/target, times in ms - readable on its own.
 	Result result;
 	result.skip_frames = _skip_frames;
-	result.metrics	   = ov::String::FormatString("Input: %.1f/%.1ffps, Busy: %.1f%%, Budget: %.2fms (proc %.2f + handoff %.2f), Capacity: %.1ffps, Output: %.1f/%.1ffps (cadence %.1f)",
+	result.metrics	   = ov::String::FormatString("Input: %.1f/%.1ffps, Busy: %.1f%%, Cost: %.2fms (proc %.2f + handoff %.2f), Capacity: %.1ffps, Output: %.1f/%.1ffps (max %.1f)",
 												  observation.actual_input_fps, observation.expected_input_fps,
 												  utilization * 100.0,
-												  frame_budget_us / 1000.0, _weighted_avg_frame_processing_time_us / 1000.0, _weighted_avg_frame_handoff_time_us / 1000.0,
+												  frame_cost_us / 1000.0, _weighted_avg_frame_processing_time_us / 1000.0, _weighted_avg_frame_handoff_time_us / 1000.0,
 												  sustainable_fps,
-												  observation.actual_output_fps, observation.expected_output_fps, observation.cadence_fps);
+												  observation.actual_output_fps, observation.expected_output_fps, observation.max_output_fps);
 
 	// Increase skip frames immediately when bottleneck occurs.
 	if (_skip_frames < next_skip_frames)
 	{
+		// The step down could not hold. Wait longer before trying the next one.
+		if (_probe_origin_level > 0)
+		{
+			_recovery_hold_ms	= std::min(_recovery_hold_ms * 2, kRecoveryHoldMaxMs);
+			_probe_origin_level = 0;
+		}
+
 		result.decision	   = Decision::Bottleneck;
 		result.skip_frames = next_skip_frames;
 
@@ -222,8 +233,15 @@ std::optional<SkipFramesController::Result> SkipFramesController::Evaluate(int64
 	// Decrease skip frames slowly when the system is recovering.
 	else if (_skip_frames > next_skip_frames)
 	{
-		if (elapsed_since_change_ms > kRecoveryHoldIntervalMs)
+		if (elapsed_since_change_ms > _recovery_hold_ms)
 		{
+			// The last step down stood. Back to the base interval.
+			if (_probe_origin_level > 0)
+			{
+				_recovery_hold_ms = kRecoveryHoldIntervalMs;
+			}
+			_probe_origin_level = _skip_frames;
+
 			// Come down 20% at a time, never in one jump.
 			int32_t rate_limited_next = _skip_frames - std::max(1, _skip_frames / 5);
 			next_skip_frames		  = std::max(rate_limited_next, next_skip_frames);
@@ -254,12 +272,16 @@ void SkipFramesController::InheritFrom(const SkipFramesController &previous)
 	_weighted_avg_frame_processing_time_us = previous._weighted_avg_frame_processing_time_us;
 	_weighted_avg_frame_handoff_time_us	   = previous._weighted_avg_frame_handoff_time_us;
 
-	// The level counts frames within a cadence, so it only transfers between two automatic
+	// The level is a divisor of the output rate, so it only transfers between two automatic
 	// controllers. A configured level was already applied by Configure().
 	if (IsAutomatic() == true && previous.IsAutomatic() == true)
 	{
 		_skip_frames	  = previous._skip_frames;
 		_bottleneck_count = previous._bottleneck_count;
+
+		// What the probes learned describes the pipeline, not the filter measuring it.
+		_recovery_hold_ms	= previous._recovery_hold_ms;
+		_probe_origin_level = previous._probe_origin_level;
 
 		// Both carry, or neither does: Evaluate() re-seeds the pair whenever either is zero,
 		// which would restart the recovery hold at every replacement.
--- a/src/transcoder/filter/filter_skipframes_controller.cpp
+++ b/src/transcoder/filter/filter_skipframes_controller.cpp
@@ -35,6 +35,9 @@ namespace
 	// steps the level may rise in one window
 	constexpr int32_t MAX_INCREASE_PER_WINDOW	= 1;
 
+	// how long a step down must stand before it counts
+	constexpr int64_t PROBE_SETTLE_INTERVAL_MS = EVALUATION_INTERVAL_MS * (BOTTLENECK_CONFIRM_COUNT + 1);
+
 	// Weight given to the newest sample in the per-frame time averages.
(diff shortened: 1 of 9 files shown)


src/transcoder/transcoder_filter.cpp
```diff
@@ -573,6 +574,12 @@ void TranscodeFilter::OnComplete(TranscodeResult result, std::shared_ptr<MediaFr
 	{
 		_complete_handler(result, _id, frame);
 	}
```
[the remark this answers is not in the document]
Keukhan: Not important enough to change, so leaving this as is.

The fault-injection delay is inside the timer on purpose. That block only runs when fault injection is switched on, which never happens in production, and the `Filter`/`Lagging` fault is injected inside `OnComplete()` itself - moving the timer past it would make that fault a no-op for SkipFrames and remove the only way to test this path.

Setting the codec module and device ID is two field assignments. The thresholds this feeds are tens of milliseconds, so it does not register.

`_complete_handler` is set in `Configure()` before the thread starts and both `Create()` overloads require it, so it is never null in practice. Even if it were, the measured time would be near zero.

src/transcoder/filter/filter_skipframes_controller.h
```diff
@@ -0,0 +1,125 @@
...
+//
+//   * Judge against the INPUT rate, never the output rate the level aims at. Skipping
+//     lowers that output target, so the filter would keep meeting a bar it just lowered
+//     while the upstream queue grew.
+//   * Busy is not overloaded. The encoder queue is two frames deep, so the handoff blocks
+//     even on a chain that keeps up. Idle time in the window is what tells them apart.
+//
+// Computation only: no clock, no logging, no FPS filter. The caller supplies the time and
+// applies the result, which is what makes the tuning ratios testable.
+class SkipFramesController
```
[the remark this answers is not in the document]
Keukhan: Added in 1bfa6e4 - `filter_skipframes_controller_test.cpp`, 13 cases covering the evaluation interval, bottleneck confirmation, one-step-per-window rise, recovery hold, 20% decay and inheritance.

src/transcoder/filter/filter_skipframes_controller.cpp
```diff
@@ -0,0 +1,281 @@
...
+#include "filter_fps.h"
+
+#define _SKIP_FRAMES_EVALUATION_INTERVAL_MS 	1000	// how often the level is reconsidered
+#define _SKIP_FRAMES_RECOVERY_HOLD_INTERVAL_MS 	5000	// how long a level must stand before it may drop
+#define _SKIP_FRAMES_SAFETY_MARGIN_RATIO 		0.9f	// aim at 90% of what the budget allows
+#define _SKIP_FRAMES_SATURATION_RATIO 			0.95f	// busy this much of a window counts as saturated
+#define _SKIP_FRAMES_RECOVERY_MAX_BUSY_RATIO 	0.80f	// below this the thread has room to try one level down
+#define _SKIP_FRAMES_INPUT_KEEP_UP_RATIO 		0.95f	// share of the input rate the thread must consume
+#define _SKIP_FRAMES_BOTTLENECK_CONFIRM_COUNT 	2		// windows that must agree before the level rises
+#define _SKIP_FRAMES_MAX_INCREASE_PER_WINDOW 	1		// steps the level may rise in one window
```
dimiden: These constant names don't seem to follow our existing conventions.
What do you think about moving them into an anonymous namespace as `constexpr` constants to prevent potential naming collisions?

src/transcoder/filter/filter_skipframes_controller.cpp
```diff
@@ -0,0 +1,281 @@
...
+	{
+		_bottleneck_count = 0;
+
+		// A thread keeping up with its input has earned a step down even while it looks
+		// busy; gating on utilization alone would pin a level reached during a spike
+		// forever. The hold interval makes this a probe - if the level below cannot hold,
+		// the next window says so and puts it back.
+		bool has_headroom = (utilization < _SKIP_FRAMES_RECOVERY_MAX_BUSY_RATIO) || (is_falling_behind == false);
+
+		next_skip_frames = has_headroom ? FilterFps::SkipFramesMin : _skip_frames;
```
dimiden: `has_headroom` is an `||` - it goes `true` as soon as the level keeps up with the input, and the
target then becomes `SkipFramesMin` regardless of utilization or budget. A nonzero level is unholdable: succeeding is what releases it.

So, if you set `SkipFrames` as `0`, publish as 30fps, and any sustained overload (e.g. a rescale
costing 238ms/frame), the level never settles. Climbs to 6 by t=7s, Recovery 6->5 at t=12s,
Bottleneck 5->6 at t=14s, then every 7s indefinitely - ~2s per cycle one level low.

Re-targeting the budget here doesn't fix it: the handoff term dies with the backpressure
(153.6 -> 74.8ms in two windows). Needs hysteresis on the last known-bad level.

src/transcoder/filter/filter_skipframes_controller.cpp
```diff
@@ -0,0 +1,281 @@
...
+	}
+	else
+	{
+		_bottleneck_count = 0;
+
+		// A thread keeping up with its input has earned a step down even while it looks
+		// busy; gating on utilization alone would pin a level reached during a spike
+		// forever. The hold interval makes this a probe - if the level below cannot hold,
+		// the next window says so and puts it back.
+		bool has_headroom = (utilization < _SKIP_FRAMES_RECOVERY_MAX_BUSY_RATIO) || (is_falling_behind == false);
```
dimiden: Raising requires `actual_input_fps > 0` (line 178); lowering doesn't, so a not-yet-measured 0.0
reads as "keeping up".

So, if you set `SkipFrames` as `0`, let the level climb to 8, and then trigger a filter swap
(format change or timestamp jump) more than 5s after the last level change, the first post-swap
window steps down at 99.5% utilization - the new `FilterFps` reports 0.0 until its own first Pop
+ 1s, while `InheritFrom` carries the level and the already-expired hold. Measured 8 -> 7,
restored two windows later, ~2s of over-driving the 2-deep encoder queue.

Deterministic, not a race: the window boundary sits at inherited `_last_check_time_ms` + 1000ms.

src/transcoder/filter/filter_skipframes_controller.cpp
```diff
@@ -0,0 +1,281 @@
...
+	// Keep skip frames unchanged when the system is stable.
+	else
+	{
+		result.decision = Decision::Unchanged;
+	}
+
+	return result;
+}
+
+void SkipFramesController::InheritFrom(const SkipFramesController &previous)
```
dimiden, line 314: `InheritFrom` carries `_last_check_time_ms` but not `_window_busy_time_us`, so the first
post-swap window divides partial busy time by a full window length.

So, if you set `SkipFrames` as `0`, sit at level 8 with the input clearly behind at 12/30fps, and
a swap lands partway into the inherited window, the level drops on a perfectly valid sample.
Measured with the remainder of the window fully busy: +100ms reads 89.5% and holds, +200ms reads
79.5% and steps down, +700ms reads 29.6%. Any swap past ~20% into the window.

Fixing the zero-sample guard alone leaves this open. Inherit both, or reset
`_last_check_time_ms` to the swap instant.

dimiden (changes requested):
I've left a few comments, please take a look! 👍

src/transcoder/filter/filter_skipframes_controller.cpp
Keukhan: Applied your suggestion. thanks.  👍

src/transcoder/filter/filter_skipframes_controller.cpp
Keukhan: Fixed

src/transcoder/filter/filter_skipframes_controller.cpp
Keukhan: Improved. the recovery no longer repeats.

src/transcoder/filter/filter_skipframes_controller.cpp
Keukhan: Fixed.  skipFrames is no longer evaluated while the input rate is unmeasured.

src/transcoder/filter/filter_video_base.cpp
```diff
@@ -21,8 +21,8 @@ bool FilterVideoBase::InitializeFpsFilter()
 	_fps_filter.SetInputFrameRate(_input_track->GetFrameRate());
 
 	// Configure skip frames
-	int32_t skip_frames_config = _output_track->GetSkipFramesByConfig();
 #if _SKIP_FRAMES_ENABLED
+	int32_t skip_frames_config = _output_track->GetSkipFramesByConfig();
 	int32_t skip_frames = (skip_frames_config >= FilterFps::SkipFramesMin) ? skip_frames_config : FilterFps::SkipFramesDisabled;
 	_fps_filter.SetSkipFrames(skip_frames);
```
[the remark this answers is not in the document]
Keukhan: Changed to shared video base filter.

src/transcoder/filter/filter_skipframes_controller.h
```diff
@@ -0,0 +1,157 @@
...
+	static constexpr double kRecoveryMaxBusyRatio	 = 0.80;
+	// share of the input rate the thread must consume
+	static constexpr double kInputKeepUpRatio		 = 0.95;
+	// windows that must agree before the level rises
+	static constexpr int32_t kBottleneckConfirmCount = 2;
+	// steps the level may rise in one window
+	static constexpr int32_t kMaxIncreasePerWindow	 = 1;
+
+	// Weight given to the newest sample in the per-frame time averages.
+	static constexpr double kFrameTimeEmaAlpha = 0.1;
```
dimiden: These ten are used in exactly two files: their own declaration here and
`filter_skipframes_controller.cpp` (12 sites). Nothing outside references them - not even the
test, which re-states the values as its own literals (`kWindowMs = 1001` for the 1000ms interval,
a bare `6000` for the 5000ms hold, `0.99` against `kSaturationRatio`).

So the public placement buys nothing today and costs a rebuild of every TU that includes this
header whenever a tuning ratio moves. An anonymous namespace in the .cpp is the right home.

Also drop the `k` prefix - OME names constants in SCREAMING_SNAKE_CASE. There are other `kXxx`
sites in the tree, but those are slated to be unified, so new code shouldn't add to them.

If the intent was to let the test assert against these rather than duplicate them, that's a fair
reason to keep them in the header - but then the test should actually use them.

Keukhan: Changed to SCREAMING_SNAKE_CASE and anonymous namespace. thanks.

src/transcoder/filter/filter_skipframes_controller.cpp
```diff
@@ -0,0 +1,321 @@
...
+	else if (_skip_frames > next_skip_frames)
+	{
+		if (elapsed_since_change_ms > _recovery_hold_ms)
+		{
+			// The last step down stood. Back to the base interval.
+			if (_probe_origin_level > 0)
+			{
+				_recovery_hold_ms = RECOVERY_HOLD_INTERVAL_MS;
+			}
+			_probe_origin_level = _skip_frames;
```
dimiden: `_probe_origin_level` is only cleared by a rise, so a probe that succeeded stays marked outstanding.

Level 0 can't produce a Recovery, so the flag from the last 1 -> 0 step is never cleared and the
next unrelated overload charges it as a failed probe. On a track bouncing 0 <-> 1 the single
recovery step per episode never reaches the reset either, so the hold ratchets and pins at the cap -
measured first-probe waits 5005, 10010, 20020, 30030, 30030 ms. That is the ratchet
`ReturnsToTheBaseHoldOnceAProbeStands` says cannot happen.

The reset is also too eager: on a multi-step descent the second step puts the hold back to base
even though the descent is about to fail, so the backoff never engages (7.5fps encoder:
4 -> 3 -> 2 -> 3 -> 4, 24.9s period, 38% below target).

Clear it when the probe is judged to have stood, not only when it fails.

src/transcoder/filter/filter_skipframes_controller.cpp
```diff
@@ -0,0 +1,321 @@
...
+#include "filter_fps.h"
+
+namespace
+{
+	// how often the level is reconsidered
+	constexpr int64_t EVALUATION_INTERVAL_MS	= 1000;
+	// how long a level must stand before it may drop
+	constexpr int64_t RECOVERY_HOLD_INTERVAL_MS	= 5000;
+	// ceiling the hold backs off to while probes keep failing
+	constexpr int64_t RECOVERY_HOLD_MAX_MS		= 30000;
```
dimiden: The probe never stops, only gets rarer. Against a permanently encoder-bound chain the level cycles
for as long as the overload lasts - 6.5fps encoder, 1800s run: 58 Bottleneck / 54 Recovery, period
34.4s, 4.3s per cycle one level low, no drift. Worth deciding whether it should give up after N
consecutive failures.

(Still a large improvement: the rescaler-bound oscillation is gone outright - Recovery=0 across a
40..500ms cost sweep, vs 14-25 per point on the previous head.)

src/transcoder/filter/filter_skipframes_controller.cpp
```diff
@@ -0,0 +1,321 @@
...
+	// How far the output ceiling has to be divided down to land on the sustainable rate.
+	auto next_skip_frames = static_cast<int32_t>(std::ceil(observation.max_output_fps / sustainable_fps - 1.0));
+	if (next_skip_frames > observation.max_output_fps - 1)
+	{
+		next_skip_frames = static_cast<int32_t>(std::floor(observation.max_output_fps - 1));
+	}
+	else if (next_skip_frames < FilterFps::SkipFramesMin)
+	{
+		next_skip_frames = FilterFps::SkipFramesMin;
+	}
```
dimiden, line 187: The `SkipFramesMin` clamp is an `else if`, so the ceiling branch above skips it: for
`max_output_fps` < 1, `floor(max_output_fps - 1)` is negative and the level becomes -1, which is
`SkipFramesDisabled`. Reachable with an explicit output `<Framerate>` and an input measured under
1fps. Make the floor clamp unconditional.

src/transcoder/filter/filter_skipframes_controller.cpp
Keukhan: The problem was improved by increasing the hold time for recovery (30 -> 60). Fundamentally, recovery is unavoidable.

Thanks.

src/transcoder/filter/filter_skipframes_controller.cpp
Keukhan: Fixed. good suggestion.

src/transcoder/filter/filter_skipframes_controller.cpp
Keukhan: Good suggestion. 👍 
a step down is now checked after a set time instead of waiting for the next step down, and the hold only goes back to 5s if the level is at or below the one that failed before.

src/transcoder/filter/filter_skipframes_controller.cpp
```diff
@@ -0,0 +1,339 @@
...
+	{
+		// Only a step at or below the one that failed proves the pipeline gained room.
+		if (_skip_frames <= _known_bad_skip_frames)
+		{
+			_known_bad_skip_frames = -1;
+			_recovery_hold_ms	   = RECOVERY_HOLD_INTERVAL_MS;
+		}
+
+		_probe_deadline_ms = 0;
+	}
```
dimiden, line 264: The settle judgement runs before the Bottleneck branch and clears `_probe_deadline_ms`
unconditionally, so a probe that fails on the third window is never recorded as failed - the
Bottleneck branch below sees 0 and neither sets `_known_bad_skip_frames` nor doubles the hold.

Three windows is exactly how long re-detection takes when the cost EMA has to ramp back: one
window for the target to exceed the level again, plus BOTTLENECK_CONFIRM_COUNT. Windows land at
+1001/+2002/+3003 and 3003 > 3000, so `PROBE_SETTLE_INTERVAL_MS` and the re-detection latency
collide.

Closed loop, 30fps in / 5ms rescale / 6.5fps encoder, 400 windows, correct level 4:
  8cb4cca6e  17 Bottleneck, 13 Recovery, 43/400 below target (10.8%), hold reaches 30000
  37eb71f6f  53 Bottleneck, 49 Recovery, 151/400 (37.8%), hold stays 5000
  37eb71f6f with (BOTTLENECK_CONFIRM_COUNT + 2)  12/8, 28/400 (7.0%), hold reaches 60000

At a 7.5fps encoder re-detection takes two windows and the backoff works, so this is on or off
depending on the plant. Either judge the probe after the branches, or size the settle interval
above the worst-case re-detection latency rather than at the confirm count.

Keukhan: Fixed. changed the settle interval from 3 windows to the base recovery hold (5s), and it is much more stable now.  :+1:

dimiden: @Keukhan 

`_probe_deadline_ms = 0` at :263 is outside the inner `if`, and this block runs before the
Bottleneck branch at :267 - so a step down that fails one window after the deadline is never
recorded as failed.

4.0fps encoder, 30fps in, sustainable level 7:

```
  t=14.013s  L=6  cost= 59ms  target=1  Recovery    dl=19013
  t=19.018s  L=6  cost=209ms  target=6  Unchanged   dl=0       <- judged as "stood"
  t=20.019s  L=7  cost=218ms  target=7  Bottleneck  dl=0  bad=-1  <- nothing recorded
```

The cost EMA needs 6 windows to push `target` past 6, because level 6 only commits 4 frames per
window. So `_recovery_hold_ms` stays at 5000 and the level cycles 7 <-> 6 every 11s indefinitely,
+6 of those 11 seconds over-driving the encoder. Measured 330/600 windows below the sustainable
level (55%), vs 115/600 (19%) on 8cb4cca6e.

5000 fixed the 6.5fps case (3-window re-detection) but not this one, and no fixed millisecond value
covers both - re-detection latency scales with 1/output-rate. Consider ending the probe on the
decision signal rather than a clock: keep it outstanding until either a Bottleneck fires (failed)
or `target <= _skip_frames` holds for N consecutive windows (stood).

Keukhan: This PR aims to restore the existing simple algorithm, SkIpFrames, which did not function due to regression. In my judgment, the current level of algorithm is sufficient. It doesn't have to be perfect. If performance is delayed, the SF level will increase, and if there is room for performance, it can be recovered. Therefore, there should be no queues and no delays. I will reject this review. Thank you.

dimiden (approved):
LGTM

PR #2294, opened 2026-08-18
build: parallelize the openh264 build

naanlizard:
The below is Claude-written (as is the change).

`InstallPrerequisites.cmake` builds every dependency with the parallel job count except openh264, whose `make` runs serially. This adds `${_J}` there like the other recipes.

Measured on a 16-thread machine: the openh264 step drops from ~35-53 s to ~10-12 s, and a full `--no-cache` Docker build was ~30 s faster.


Keukhan (approved):
Thanks for PR. 👍 
LGTM

PR #2300, opened 2026-08-20
feat: add transport connection info to ingress stream alerts and REST API

dimiden:
## Summary

Adds a `connection` object (transport, protocol, local/remote address and port) to ingress stream alerts (`sourceInfo`) and the REST API stream response (`input`).
For WebRTC it reports the ICE-selected path: with `?transport=all` there was previously no way to tell whether a stream ended up on UDP, direct TCP, or the TURN relay.
RTMP/SRT/MPEG-TS streams report their socket connection as well.

## Changes

- `info::ConnectionInfo` snapshot stored in `StreamStats`; publication is version-gated so a stale selection racing a newer one on another socket thread cannot overwrite it.
- New `IcePortObserver::OnIceCandidatePairSelected()` (default no-op) fires on the first selection and every switch of the active candidate pair; the WebRTC provider classifies the pair as UDP/TCP/TURN and records it (WHIP and WebSocket signalling).
- RTMP/MPEG-TS (SRT) record their connection at stream creation, so it is available from `INGRESS_STREAM_CREATED`; WebRTC from `INGRESS_STREAM_PREPARED` (after ICE completes).
- serdes: implement the `SetConnection()` stub; add `std::optional` setter overloads.
- Unit tests for callback/version ordering, transport classification, and the version gate.

## Testing

Publish via WHIP with `?transport=udp|tcp|relay|all` and via RTMP, then check the `INGRESS_STREAM_PREPARED` alert (`sourceInfo.connection`) and `GET /v1/vhosts/{vhost}/apps/{app}/streams/{stream}` (`input.connection`), e.g.:

```json
{"transport": "TURN", "protocol": "TCP", "localAddress": "10.0.0.5", "localPort": 3478, "remoteAddress": "1.2.3.4", "remotePort": 51234}
```

### DIFF
--- a/src/base/info/stream_stats.cpp
+++ b/src/base/info/stream_stats.cpp
@@ -57,7 +57,7 @@ namespace info
 
 		if (local_port.has_value())
 		{
-			result.AppendFormat("%s:%u", local_address.CStr(), local_port.value());
+			result.AppendFormat("%s:%" PRIu16, local_address.CStr(), local_port.value());
 		}
 		else
 		{
@@ -68,7 +68,7 @@ namespace info
 
 		if (remote_port.has_value())
 		{
-			result.AppendFormat("%s:%u", remote_address.CStr(), remote_port.value());
+			result.AppendFormat("%s:%" PRIu16, remote_address.CStr(), remote_port.value());
 		}
 		else
 		{
--- a/src/base/info/stream_stats.cpp
+++ b/src/base/info/stream_stats.cpp
@@ -10,6 +10,76 @@
 
 namespace info
 {
+	std::shared_ptr<const ConnectionInfo> ConnectionInfo::From(const ov::SocketAddress *local_address, const ov::SocketAddress *remote_address, ov::SocketType socket_type, const char *transport, uint64_t version)
+	{
+		auto connection_info	   = std::make_shared<ConnectionInfo>();
+
+		auto protocol			   = ov::StringFromSocketType(socket_type);
+
+		connection_info->protocol  = protocol;
+		connection_info->transport = (transport != nullptr) ? transport : protocol;
+
+		if ((local_address != nullptr) && local_address->IsValid())
+		{
+			connection_info->local_address = local_address->GetIpAddress();
+			connection_info->local_port	   = local_address->Port();
+		}
+		if ((remote_address != nullptr) && remote_address->IsValid())
+		{
+			connection_info->remote_address = remote_address->GetIpAddress();
+			connection_info->remote_port	= remote_address->Port();
+		}
+		connection_info->version = version;
+
+		return connection_info;
+	}
+
+	std::shared_ptr<const ConnectionInfo> ConnectionInfo::From(const ov::SocketAddressPair &address_pair, ov::SocketType socket_type, const char *transport, uint64_t version)
+	{
+		return From(&(address_pair.GetLocalAddress()), &(address_pair.GetRemoteAddress()), socket_type, transport, version);
+	}
+
+	std::shared_ptr<const ConnectionInfo> ConnectionInfo::From(const std::shared_ptr<ov::Socket> &socket, const char *transport, uint64_t version)
+	{
+		if (socket == nullptr)
+		{
+			return nullptr;
+		}
+
+		return From(socket->GetLocalAddress().get(), socket->GetRemoteAddress().get(), socket->GetType(), transport, version);
+	}
+
+	ov::String ConnectionInfo::ToString() const
+	{
+		auto result = ov::String::FormatString(
+			"%s (%s, local: ",
+			transport.CStr(), protocol.CStr());
+
+		if (local_port.has_value())
+		{
+			result.AppendFormat("%s:%u", local_address.CStr(), local_port.value());
+		}
+		else
+		{
+			result.Append("N/A");
+		}
+
+		result.Append(", remote: ");
+
+		if (remote_port.has_value())
+		{
+			result.AppendFormat("%s:%u", remote_address.CStr(), remote_port.value());
+		}
+		else
+		{
+			result.Append("N/A");
+		}
+
+		result.Append(")");
+
+		return result;
+	}
+
 	StreamStats::StreamStats()
 		: _created_time(std::chrono::system_clock::now())
 	{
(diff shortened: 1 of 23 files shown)


src/modules/json_serdes/common.h
```diff
@@ -69,4 +71,37 @@ namespace serdes
...
+		}
+	}
+
+	inline void SetInt64(Json::Value &parent_object, const char *key, const std::optional<int64_t> &value)
+	{
+		if (value.has_value())
+		{
+			SetInt64(parent_object, key, value.value());
+		}
+	}
```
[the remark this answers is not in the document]
dimiden: No, this won't cause an error due to C++ casting rules

src/base/info/stream_stats.cpp
```diff
@@ -10,6 +10,76 @@
...
+			transport.CStr(), protocol.CStr());
+
+		if (local_port.has_value())
+		{
+			result.AppendFormat("%s:%u", local_address.CStr(), local_port.value());
+		}
+		else
+		{
+			result.Append("N/A");
+		}
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/providers/mpegts/mpegts_stream.cpp
```diff
@@ -43,6 +43,9 @@ namespace pvd
 		SetName(stream_name);
 		_remote = client_socket;
 		SetMediaSource(ov::String::FormatString("%s://%s", ov::StringFromSocketType(client_socket->GetType()), remote_address.ToString(false).CStr()));
+
+		// The shared UDP listener socket knows no remote address; use the packet source instead
+		SetConnectionInfo(info::ConnectionInfo::From(client_socket->GetLocalAddress().get(), &remote_address, client_socket->GetType()));
```
Keukhan: `ClientSocket` calls `RetrieveLocalAddress()` (getsockname), so RTMP, MPEG-TS/TCP and SRT all report a real interface address. MPEG-TS/UDP is the exception: `client_socket` is the shared `DatagramSocket`, whose `_local_address` is the requested *bind* address, and the shipped default is `<IP>*</IP>` — so `input.connection.localAddress` ends up as `0.0.0.0`.

`SocketAddress` already tracks `_is_wildcard_host` but doesn't expose it. Exposing it and skipping `local_address` in `From()` while keeping `local_port` (the useful half) would cover any wildcard-bound listener.

src/providers/webrtc/webrtc_provider.cpp
```diff
@@ -711,6 +711,49 @@ namespace pvd
...
+		if (socket == nullptr)
+		{
+			return;
+		}
+
+		// A non-direct pair goes through the built-in TURN server regardless of the socket protocol
+		auto connection_info = info::ConnectionInfo::From(
+			candidate_pair->GetAddressPair(),
+			socket->GetType(),
+			(candidate_pair->GetTransportType() != IceCandidatePair::TransportType::Direct) ? "TURN" : nullptr,
```
Keukhan: The PR body lists "transport classification" as covered, but this line — the only place the `Direct` → protocol / non-`Direct` → `"TURN"` decision is made — never executes in a test. `stream_stats_test` calls `ConnectionInfo::From()` with the transport string already chosen, and `ice_port_test` builds its pairs via `OnReceivedStunBindingRequest(pair, nullptr)`, so `GetSocket()` is null and this handler would bail out before the classification anyway.

Keukhan (commented):
Two things worth a look.

src/providers/mpegts/mpegts_stream.cpp
Keukhan: Follow-up: the correct value is already computed and then thrown away. `Socket::RecvFrom()` enables `IP_PKTINFO` / `IPV6_RECVPKTINFO` at socket creation and fills the per-datagram destination address into `address_pair->SetLocalAddress()`, but `MpegTsProvider::OnDatagramReceived()` forwards only `address_pair.GetRemoteAddress()`.

So no new `SocketAddress` accessor is needed — plumbing the pair through and using the `From(const ov::SocketAddressPair &, ...)` overload this PR already adds is enough:

```cpp
// OnDatagramReceived (UDP)
OnConnected(remote, address_pair)

// OnConnected(remote) (TCP)
OnConnected(remote, ov::SocketAddressPair(*remote->GetLocalAddress(), *remote->GetRemoteAddress()))

// MpegTsStream ctor
SetConnectionInfo(info::ConnectionInfo::From(address_pair, client_socket->GetType()));
```

SRT hands over a `ClientSocket`, which already has a real local address via `RetrieveLocalAddress()`, so the same shape works there and all three paths become uniform — one signature change on `MpegTsStream::Create()` plus its three callers.

One caveat: when the pktinfo cmsg is absent `QueryLocalAddress()` returns an invalid address, and `From()` then drops `localPort` as well, so falling back to `client_socket->GetLocalAddress()` for the port may be worth it.

Unrelated to this PR, but it affects the value once it is published: in `QueryLocalAddress()` the IPv6 branch does `sock->sin6_port = local_port;` while IPv4 does `HostToNetwork16(local_port)`, so IPv6/UDP would report a byte-swapped port. The line below it also re-reads from `CMSG_FIRSTHDR` instead of the matched cmsg, shadowing the outer `pktinfo`. Probably worth a separate issue.

src/providers/webrtc/webrtc_provider.cpp
Keukhan: Follow-up: this is untestable only because the decision is fused with the `socket->GetType()` deref. The decision itself needs no socket, so pulling it onto the pair opens a seam:

```cpp
// ice_candidate_pair.h
// The transport to report for this pair: `nullptr` means "use the socket protocol",
// "TURN" means OME's built-in TURN server relays it.
const char *GetReportedTransport() const
{
    return (GetTransportType() != TransportType::Direct) ? "TURN" : nullptr;
}
```

Then the socket-less pairs `ice_port_test` already builds cover it with no new test infrastructure:

```cpp
EXPECT_EQ(pair->GetReportedTransport(), nullptr);
pair->SetTurnDataChannel(0x4000);
EXPECT_STREQ(pair->GetReportedTransport(), "TURN");
```

`stream_stats_test`'s `FromClassification` already covers transport-string -> transport/protocol fields, so this closes the remaining `TransportType` -> transport-string hop.

dimiden: Fixed.

src/providers/mpegts/mpegts_stream.cpp
dimiden: Fixed.
And, confirmed both IPv6 findings (missing `HostToNetwork16()` and the shadowed `pktinfo`); tracking them separately.

dimiden:
@Keukhan
I have addressed both of your comments. Thank you!

Keukhan (approved):

PR #2301, opened 2026-08-21
build: stop installing clang when it was not asked for

naanlizard:
The below is Claude-written (as is the change).

### Problem

The installer installs `clang` and `lld` before it looks at which dependency was asked for,
so reinstalling one library also installs a compiler. `Dependencies.cmake` takes that path on
any version mismatch, and a stale jemalloc is enough to trigger it.

`-DOME_USE_CLANG=OFF` does not stop it. The installer runs as a child process, and
`Dependencies.cmake` forwards `OME_HWACCEL_NVIDIA`, `OME_HWACCEL_XMA`, `OME_ENABLE_X264` and
`OME_WHISPER_STATIC` to it, but not `OME_USE_CLANG`, which the child then defaults to ON.

### Reproduction

Ask the installer for one dependency in a throwaway container, and see what else arrives:

```sh
docker run --rm ubuntu:22.04 bash -c '
  export DEBIAN_FRONTEND=noninteractive
  ln -fs /usr/share/zoneinfo/Etc/UTC /etc/localtime
  apt-get update -qq && apt-get install -y -qq tzdata sudo git cmake
  git clone --depth 1 https://github.com/OvenMediaLabs/OvenMediaEngine /src
  cd /src
  echo "--- before: $(command -v clang || echo "clang not present")"
  cmake -DTARGET=jemalloc -P cmake/InstallPrerequisites.cmake
  echo "--- after:  $(command -v clang || echo "clang not present")"'
```

On master, asking for jemalloc also puts a compiler on the machine (output trimmed):

```
--- before: clang not present
-- [OME Prerequisites] Installing clang/lld (OME_USE_CLANG=ON)
-- [OME Prerequisites] Building: apt clang
...
--- after:  /usr/bin/clang
```

With this change:

```
--- before: clang not present
-- [OME Prerequisites] TARGET=jemalloc - building that dependency only, skipping clang
-- [OME Prerequisites] Install targets: jemalloc
...
--- after:  clang not present
```

### Change

Skip the clang block when a single `TARGET` is named, and forward `OME_USE_CLANG` to the
child alongside the other options.

Base packages still install either way, since `autoconf`, `automake` and `libtool` are what
the requested dependency needs in order to build.


dimiden:

PR #2303, opened 2026-08-21
refactor(llhls): move segmentation decisions into boundary policies and add synced segmentation

getroot:
Segmentation decisions were spread across the fMP4 packager and the storage, so adding a segmentation behavior meant editing the packager, and every track decided its own cut positions and segment numbers.

This PR turns those decisions into a module. A boundary policy owns the numbering, the boundary of the accumulating segment, what goes out on each pass, and the marker and discontinuity cuts. The packager and the storage hand over timing facts and execute the answer, so a new behavior is a new policy file plus one line in the factory.

The mode it makes room for is `synced`. In duration mode each track accumulates to the configured duration and cuts at its next keyframe, so the boundary for the same segment index sits at a slightly different position on each track, and a marker cutting near one of those boundaries can land on a different segment number per track. In synced mode one reference track decides the boundaries and publishes them, and every other track cuts there and takes those numbers, markers included. Use it when renditions must carry a marker on the same segment number.

**Changes**

- `SegmentationMode` option: `duration` (default), `synced`.
- `CueOutCutMode` option: `keyframe` (default), `immediate`. A CUE-OUT now cuts at the next keyframe by default, so every segment starts at a keyframe. Streams that need the break to start exactly at the announced position, and that replace the break content, set `immediate`, which is how it behaved before.
- Markers settle by position instead of estimated sequence numbers, and keep the time they were announced with. A late position cuts at the nearest position that can still cut instead of being refused.
- The sender decides where a break returns. A CUE-OUT or SCTE-35 OUT announces its length, and the CUE-IN or splice-in that closes it comes from the sender, so LLHLS and HLS close a break the same way and no return point is invented mid-pipeline. Senders that relied on a return point being generated for them now send it themselves.
- Markers no longer reserve a segment number across tracks, and no track defers its close to match one. Renditions get identical numbering from synced mode instead.
- The HLS (TS) publisher now enforces the minimum break length that was inert until now, so a break shorter than its segment cadence is refused.

**Test**

A refactor, so the point is that the existing output is unchanged.

1. RTMP or SRT in, LLHLS and HLS (TS) out, default config. EXTINF within EXT-X-TARGETDURATION, consecutive segment and part numbers, increasing EXT-X-PROGRAM-DATE-TIME, parts within PART-TARGET, playback without stalls.
2. ABR: rendition switching stays continuous and rendition segment numbers stay aligned.
3. Input configuration change while publishing: every rendition carries the discontinuity at the same position and playback recovers.
4. DVR and dump, and CENC with a key rotation: as before.
5. Ad breaks, CUE and SCTE-35: the break opens where announced, the sender's return point closes it there, and no marker is refused. A second break after a closed one is accepted.
6. `SegmentationMode=synced`: repeat 1 to 3 and 5, and check that one segment number means the same range on every track and that a marker rides the same number on all of them.
7. `CueOutCutMode=immediate`: a CUE-OUT in the middle of a GOP ends the segment at the marker instead of the next keyframe.

Unit tests: 764 pass, covering marker rules and cut positions, discontinuity settlement and its position, duration pacing and its catch-up, the stream clock including a track restart, synced adoption and marker relay, a two track synced pipeline, and part cadence on a fractional timescale.


getroot (commented):
Review of the boundary policy refactor. Fifteen inline notes below, ordered roughly by impact: the first five can change what a player receives, the rest are consistency, coverage and leftover surface. Nothing here blocks the direction of the change.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
```diff
@@ -0,0 +1,705 @@
...
+		// time than it declares. Move the cut back to the last frame where it
+		// had resolved and leave the rest for the next part.
+		bool reorder_resolved = (next_frame.pts_us > buffered.GetHighestPtsUs());
+
+		if (boundary.exact == true && reorder_resolved == false && no_marker == true &&
+			completed_by_split == false && emit_count > 1)
+		{
+			// The lowest presentation timestamp from each position on, to test a
+			// candidate cut against everything it would keep
+			std::vector<int64_t> lowest_kept_pts_us(emit_count, 0);
```
getroot: `lowest_kept_pts_us` is built only over `[0, emit_count)`, but the samples past `emit_count` also stay in the buffer, so their pts never enter the minimum. When `emit_count` was already trimmed above, at the backlog path that slices the emission to one chunk, the guard can pass while a kept sample has a lower pts than an emitted one. With buffered dts order pts {0, 3, 4, 1, 2} and `emit_count == 3`, `resolved_count` lands on 2, so the emitted part covers up to pts 3 while pts 1 and 2 remain buffered. That is the overlap this block exists to prevent. The suffix minimum needs to span every sample that will stay behind, not only the ones inside the current emit window.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
```diff
@@ -0,0 +1,489 @@
...
+		plan.end_us = NextBoundaryUs(*segment_start_us) - kBoundaryBiasUs;
+
+		return plan;
+	}
+
+	bool ReferenceBoundaryPolicy::CanEmitChunk(int64_t chunk_end_us) const
+	{
+		// The reference is never held back; its append progress is the high-water
+		// mark every synced track gates its own chunks on
+		_boundary_feed->UpdateHighWaterMark(chunk_end_us);
```
getroot: `CanEmitChunk` publishes the planned chunk end as the high water mark before `GetChunkPlan` trims the emission at an immediate marker cut, so the mark can sit past the boundary the reference then publishes. With `CueOutCutMode=immediate`, a CUE-OUT at 3010 ms and 400 ms parts, the mark becomes 3200 ms while the same pass trims the reference to dts < 3010 ms and publishes the boundary there. A synced track gated only on `chunk_end_us <= high_water_mark` can already have emitted to 3200 ms, past its own `plan.end_us`, so it closes late: the segment number matches but the range does not. Raising the mark after the trim, or clamping it to the pending cut, keeps the exact guarantee intact.

src/publishers/llhls/llhls_stream.cpp
```diff
@@ -1393,78 +1405,41 @@ bool LLHlsStream::InsertMarkerToAllPackagers(uint32_t data_track_id, cmn::Bitstr
...
-				{
-					// We checked it can be inserted, so it should not fail
-					logtc("Failed to insert marker (timestamp: %" PRId64 ", tag: %s)", marker->GetTimestamp(), marker->GetTag().CStr());
-					return false;
-				}
-			}
+		if (boundary_policy->InsertMarker(marker) == false)
+		{
+			// We checked it can be inserted, so it should not fail
+			logtc("Failed to insert marker (timestamp: %" PRId64 " ms, tag: %s)", marker->GetTimestampMs(), marker->GetTag().CStr());
```
getroot: This returns in the middle of the per track loop, so the tracks already visited keep the marker. `CanInsertMarker` passing for every track no longer implies `InsertMarker` will: the new `EffectiveCutUs` refusal depends on `_newest_sample_end_us`, which the media thread can advance between the check and the insert. The visible result is a CUE or DATERANGE tag on the video chunklist and none on the audio one for the rest of the session. Either roll back the inserts already applied, or take the decision and the insert under one lock so the set is all or nothing.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
```diff
@@ -0,0 +1,705 @@
...
+				logti("%s - Marker (%s, %" PRId64 " ms) rides segment ending at %.3f ms; the media never reached its position",
+					  _log_context.CStr(), it->second->GetTag().CStr(), it->second->GetTimestampMs(), end_us / 1000.0);
+				settled.markers.push_back(it->second);
+			}
+			else
+			{
+				// Further behind than the insertion clamp would ever accept: the
+				// advertised time no longer describes where it would land, and a
+				// leftover would block every later marker
+				logtw("%s - Discarded the marker (%s, %" PRId64 " ms) because its position is more than a segment behind",
```
getroot, line 420: Discarding a stale OUT leaves a pending IN that was inserted after it, so the playlist can carry a return with no break to return from. The `_last_inserted_marker` reset below does not cover this case because the IN, not the OUT, is the last inserted marker. After a keyframe wait hole longer than one segment, the OUT announced at 10000 ms is dropped while the IN at 40000 ms still cuts and renders, giving a bare CUE-IN. Dropping an OUT should also drop the INs that depend on it.

src/base/provider/stream.cpp
```diff
@@ -142,6 +144,45 @@ namespace pvd
 		return _max_generated_timestamp_ms;
 	}
 
+	void Stream::UpdateLastTimestampStat(const std::shared_ptr<const MediaTrack> &track, const std::shared_ptr<const MediaPacket> &packet)
```
getroot, line 145: This only ever raises `_last_media_timestamp_ms`, where the code it replaces assigned unconditionally, so a rollback freezes the stream clock permanently. The comment assumes per track timestamps only move forward, which a 33 bit PTS wrap breaks, as does keeping the same `pvd::Stream` across a reconnect with a fresh timestamp base. Every later packet then compares below the stored maximum, `_elapsed_from_last_media_timestamp` is never restarted, and `GetCurrentTimestampMs()` returns the pre wrap maximum plus wall clock. Cues timestamped from that land far ahead of the real media position and the packagers refuse them as unreachable. A large backward jump should reset the clock rather than be ignored.

src/base/provider/stream.cpp
```diff
@@ -208,8 +234,82 @@ namespace pvd
...
+		{
+			return false;
+		}
+
+		SendProvisionalReturnPoint(timestamp_in_ms, format, frame, urgent, internal);
+
+		return true;
+	}
+
+	void Stream::SendProvisionalReturnPoint(int64_t out_timestamp_ms, const cmn::BitstreamFormat &format, const std::shared_ptr<ov::Data> &out_frame, bool urgent, bool internal)
```
getroot: The provisional return point is placed at the announced OUT time plus the duration, but the packagers realize the OUT at their own cut, so the break renders shorter than the DURATION it advertises. The mpegts path this replaces derived `cue_in_timestamp` from `main_segment_end_timestamp`, that is from the realized segment end. With a 6 s target duration and an OUT announced at 10000 ms but realized at 14000 ms, DURATION=30.000 is advertised while the IN covers 40000 ms, a 26 s break. An ad decision server sizing its pod from DURATION overruns the return point. Anchoring the provisional IN to the realized cut, or moving it once the cut lands, keeps advertised and rendered lengths equal.

src/modules/containers/bmff/fmp4_packager/duration_boundary_policy.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+		_total_expected_duration_us += _segment_duration_us;
+		_total_segment_duration_us += completed.duration_us;
+
+		if (added_a_boundary == true)
+		{
+			_total_expected_duration_us -= _segment_duration_us;
+		}
+
+		_target_segment_duration_us = _total_expected_duration_us - _total_segment_duration_us + _segment_duration_us;
```
getroot, line 63: The catch up ledger has no upper clamp, so one short segment pushes the next target above the configured duration while `EXT-X-TARGETDURATION` stays at the fixed `round(SegmentDuration)` value emitted by `llhls_chunklist.cpp`. `DurationBoundaryPolicyTest.ShortSegmentStretchesNextTarget` asserts that stretch to 4100 ms for a 4 s configuration, and with a 2 s GOP the realized cut can land at 6000 ms, giving EXTINF 6.000 against TARGETDURATION 4. `mpegts_packager.cpp` gained `std::min(_next_target_duration_ms, target_duration_ms)` in this same change for exactly this reason, so the fMP4 policy wants the same ceiling.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
```diff
@@ -0,0 +1,705 @@
...
+		}
+
+		// The chunk has to end here, and the reordering has not resolved at this
+		// frame: some frame still buffered displays after one that is about to
+		// go out, so the part would cover a different stretch of presentation
+		// time than it declares. Move the cut back to the last frame where it
+		// had resolved and leave the rest for the next part.
+		bool reorder_resolved = (next_frame.pts_us > buffered.GetHighestPtsUs());
+
+		if (boundary.exact == true && reorder_resolved == false && no_marker == true &&
```
getroot: Gating the reorder guard on `boundary.exact` limits it to synced mode, but nothing about the defect is synced specific. In duration mode with B frames a part takes the whole buffer, so a part ending on a P frame with pts 3 is followed by a part starting at pts 1 and 2, the same overlap the comment above describes. If the guard is correct it should apply to every mode; if it is deliberately kept off the default path, the reason belongs in the comment.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
```diff
@@ -0,0 +1,489 @@
...
+	//--------------------------------------------------------------------------
+	// SyncedBoundaryPolicy
+	//--------------------------------------------------------------------------
+
+	SyncedBoundaryPolicy::SyncedBoundaryPolicy(const std::shared_ptr<const ReferenceBoundaryPolicy> &reference_policy,
+											   const Config &config)
+		: SegmentBoundaryPolicy(config),
+		  _boundary_feed(reference_policy != nullptr ? reference_policy->GetBoundaryFeed() : nullptr)
+	{
+		OV_ASSERT2(_boundary_feed != nullptr);
```
getroot, line 265: `OV_ASSERT2` is compiled out in release builds, so a null reference policy leaves `_boundary_feed` null and the first `IsInFallback()` or `CanEmitChunk()` call faults on the first arriving sample. `FMP4Storage`'s constructor has the same shape, asserting and then calling straight through the pointer. Neither is reachable from `AddPackager` today, but both are public constructors on a module this change presents as pluggable, so failing hard at construction reads better than a crash on the media path.

src/base/provider/stream.cpp
```diff
@@ -208,8 +234,82 @@ namespace pvd
...
+
+		if (format == cmn::BitstreamFormat::CUE)
+		{
+			auto return_event = CueEvent::Create(CueEvent::CueType::IN, 0, 0, true);
+			return_frame = (return_event != nullptr) ? return_event->Serialize() : nullptr;
+		}
+		else
+		{
+			auto out_event = Scte35Event::Parse(out_frame);
+			auto return_event = Scte35Event::Create(mpegts::SpliceCommandType::SPLICE_INSERT, out_event->GetID(), false,
```
getroot: `out_event` is dereferenced without a null check while the first `Scte35Event::Parse` in this function does check for it. The parse is also redundant: the event parsed in the branch above is the same object and already carries the ID, so hoisting it out of that branch removes both the second parse and the missing check.

src/modules/containers/bmff/fmp4_packager/synced_pipeline_test.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+	return observations;
+}
+
+TEST(SyncedPipelineTest, MarkersRideTheRealizedBoundary)
+{
+	auto rig = MakeRig();
+
+	// Both cues sit mid-GOP (keyframes are at multiples of 1000); with the
+	// immediate cut mode the reference cuts right at them
```
getroot: The comment says the reference cuts right at the markers with the immediate cut mode, but `MakeRig()` leaves `cue_out_cut_mode` at the default Keyframe, so this runs in keyframe mode. `EXPECT_LT(realized_ms, requested_ms + 1000.0)` also passes in both modes for a CUE-OUT at 13210 ms with 1000 ms GOPs, so the assertion cannot tell them apart. The Immediate path, which trims samples inside a segment, currently has no pipeline level coverage.

src/modules/containers/bmff/fmp4_packager/fmp4_marker_sync_test.cpp
```diff
@@ -7,522 +7,509 @@
...
-	public:
-		void OnFMp4StorageInitialized(const int32_t &track_id) override {}
-		void OnMediaSegmentCreated(const int32_t &track_id, const uint32_t &segment_number) override {}
-		void OnMediaChunkUpdated(const int32_t &track_id, const uint32_t &segment_number, const uint32_t &chunk_number, bool last_chunk) override {}
-		void OnMediaSegmentDeleted(const int32_t &track_id, const uint32_t &segment_number) override {}
-		void OnMediaSegmentCompleted(const int32_t &track_id, const uint32_t &segment_number) override {}
-	};
-
-	struct TrackPipeline
+	bmff::SegmentBoundaryPolicy::Config MakeConfig(LLHlsCueOutCutMode cue_out_cut_mode = LLHlsCueOutCutMode::Immediate)
```
getroot: The suite this replaces drove a video and an audio pipeline together over 12 cues and asserted cross track behaviour. Every `MarkerPolicyTest` here uses `MakeVideoPipeline` alone, and only `synced_pipeline_test.cpp` runs two tracks, so duration mode, the shipping default, no longer has a multi track marker test. `MakeConfig` also defaults `cue_out_cut_mode` to Immediate while production defaults to Keyframe, so the default combination is untested on both counts.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.h
```diff
@@ -0,0 +1,291 @@
...
+		// the next segment. false: it ends at the first cuttable frame at or
+		// after end_us.
+		bool exact = false;
+
+		// The name of the segment being accumulated
+		int64_t segment_number = -1;
+
+		// Published timestamps (tfdt) of this track are shifted back by this
+		// much, anchoring the published timeline on the shared boundaries
+		int64_t publish_offset_us = 0;
```
getroot: `publish_offset_us` has no reader. It is set to 0 in `duration_boundary_policy.cpp` and asserted 0 in a test, while the comment here says published tfdt values are shifted by it, which nothing does. `OnMediaChunk` below is dispatched from `fmp4_storage.cpp` for every chunk of every track with no override anywhere. Both widen the interface this change is meant to keep narrow, so dropping them until a policy actually needs them keeps the contract honest.

src/base/modules/marker/marker.h
```diff
@@ -0,0 +1,58 @@
...
+	std::shared_ptr<CueEvent> GetCueEvent() const;
+	std::shared_ptr<Scte35Event> GetScte35Event() const;
+	ov::String ToHlsTag(int64_t timestamp_offset = 0) const;
+
+	void SetParent(const std::shared_ptr<Marker> &parent);
+	std::shared_ptr<Marker> GetParent() const;
+
+	// Move the marker's advertised time onto another position (e.g. a realized
+	// cut), for renderers that advertise realized positions
+	void SetRealizedTimestamp(int64_t timestamp, int64_t timestamp_ms);
```
getroot: `SetRealizedTimestamp` and the `Scte35Event::SetTimestampMsec` it needs have no callers, and the change settles on keeping the announced time, which `InsertMarker` logs. In synced mode one `Marker` instance is shared across every track, as the pipeline test asserts, so a future caller would rewrite the tag on all renditions at once. Better removed than left as a live setter on shared state.

src/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -638,23 +463,8 @@ namespace bmff
 		total_sample_duration = samples != nullptr ? samples->GetTotalDuration() : 0;
```
getroot: `total_sample_duration` and `total_sample_duration_ms` are assigned here and never read before the return, leftovers from the removed segmentation info block. Related: `samples` is set to `_no_samples` earlier in this function whenever the buffer is empty, so the `samples != nullptr` guards after that point can never be false and read as if the empty case were still live.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
```diff
@@ -0,0 +1,773 @@
...
+					_last_inserted_cut_us = -1;
+				}
+			}
+
+			it = _pending_markers.erase(it);
+		}
+
+		// A break that never rendered has no return to render either: a lone IN
+		// would announce a return from a break the playlist never opened
+		for (auto it = _pending_markers.begin(); (discarded_breaks.empty() == false) && (it != _pending_markers.end());)
```
Keukhan, line 441: **A lone IN can still be rendered.** This loop only scans what is *left* in `_pending_markers`, but the first loop above may have already moved the return point into `settled.markers`.

Segment 6s, break 12s. Track change, then no keyframe for over a segment, so the media resumes at 112000 and the segment spans [112000, 118000]:
- OUT at 105000: `112000 - 105000 > 6000` -> discarded, pushed to `discarded_breaks`
- IN at 117000: `117000 >= 112000` -> settled

Both are erased in the first loop, so this loop runs over an empty map. The segment carries only the IN -> `EXT-X-CUE-IN` with no OUT before it. `cut_this_segment` is wrong too, so the pacing discounts a segment no marker cut.

Filtering `settled.markers` against `discarded_breaks` as well would cover it.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+		return plan;
+	}
+
+	bool SyncedBoundaryPolicy::CanEmitChunk(int64_t chunk_end_us) const
+	{
+		// Until the lattice exists, nothing may emit: the first segment's number
+		// must come from the lattice, or every later segment inherits the
+		// initial guess
+		if (IsAnchored() == false && _boundary_feed->HasBoundaries() == false)
```
Keukhan, line 376: **Synced mode blocks every non-reference track until the reference publishes its first boundary.**

`HasBoundaries()` only becomes true once the reference *completes a segment* (`PublishBoundary` is called from `DoOnSegmentCompleted` only). Until then this returns `IsInFallback()`, which is false because the high-water mark is already rising from the reference's chunks, so the lag never reaches `2 * segment_duration`.

So up to one `SegmentDuration` passes with no synced track emitting anything - a full one when the media clock starts near zero, since the reference aims at multiples of the duration. Audio publishes no parts, buffers a whole segment, then bursts.

The block exists only so the first segment gets its number from the lattice. Publishing an anchor when the reference *opens* its first segment (number plus start position) would let synced tracks anchor immediately and still be gated by the high-water-mark check below.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
```diff
@@ -0,0 +1,491 @@
...
+		return _boundary_feed;
+	}
+
+	int64_t ReferenceBoundaryPolicy::NextBoundaryUs(int64_t position_us) const
+	{
+		// Snap jitter tolerance: a segment that starts exactly on a boundary
+		// (within half a frame) belongs to that boundary, not before it
+		const int64_t tolerance_us = (_video_frame_rate > 0) ? std::llround(500000.0 / _video_frame_rate) : 500;
+
+		int64_t multiple = std::max(static_cast<int64_t>(1), position_us / _segment_duration_us);
```
Keukhan, line 163: **Unguarded division and an unbounded loop.** `SegmentDuration` is `Register<Optional>` with no range check, so 0 is reachable. Then `position_us / _segment_duration_us` divides by zero, and `while (true)` never exits because `boundary_us` stays 0.

Master already divides by this value in the server-time numbering path, but there it is once per storage; here it is every sample of the reference track, and the loop hangs instead of crashing. Validating `SegmentDuration > 0` at config load would cover both.

Keukhan (commented):
Read through the policy refactor. The decomposition is clean and the base/hook split reads well. Four things I think are worth resolving before merge:

- a lone `EXT-X-CUE-IN` when a break is discarded and its return point is settled in the same pass (correctness)
- synced mode emitting nothing until the reference publishes its first boundary (join latency)
- `NextBoundaryUs()` dividing by an unvalidated `SegmentDuration` (robustness)
- the newly enforced minimum break length rejecting 15s breaks on the TS publisher (behaviour change)

src/modules/containers/mpegts/marker_box.cpp
```diff
@@ -0,0 +1,270 @@
...
+		const double media_position_ms = _media_position_ms.load(std::memory_order_relaxed);
+		if (media_position_ms >= 0 && _cut_cadence_ms > 0 &&
+			(media_position_ms - static_cast<double>(marker->GetTimestampMs())) > _cut_cadence_ms)
+		{
+			return {false, ov::String::FormatString("The position of the marker has already passed by more than the cut cadence : %s (%" PRId64 " ms)", marker->GetTag().CStr(), marker->GetTimestampMs())};
+		}
+
+		// A break must be long enough to return from; without a configured cadence
+		// the duration and spacing checks stay permissive
+		auto required_gap = _cut_cadence_ms * 1.5;
```
Keukhan: **Enforcing this rejects 15s ad breaks at the default HLS settings.**

On master `mpegts::Packager` never overrode `GetSegmentationInfo()`, so `GetActualTargetSegmentDurationMs()` returned -1 and `required_gap` was negative - the check always passed. This is the "inert until now" from the description.

Now, with HLS `SegmentDuration` at its default 10 and a 2s GOP, `_cut_cadence_ms` is 10000 and `required_gap` is 15000. A 15s CUE-OUT hits `15000 <= 15000` and is refused - and 15s is one of the most common break lengths. LLHLS is unaffected (it already had a real cadence there), so the same break can appear in the LL-HLS playlist and be dropped from the TS one.

Worth reconsidering the 1.5 factor for TS, or at least using `<` instead of `<=` so a break exactly one cadence-and-a-half long still packages.

src/modules/containers/bmff/fmp4_packager/duration_boundary_policy.cpp
getroot: The stretch is deliberate and carries over #2255: server-time based numbering treats every segment number as one wall-clock slot, so after a short close the next segment stretches to finish its slot. Clamping it would leave the numbering permanently behind the wall clock, and a server started later (a failback after a failover) derives its numbers fresh from the wall clock and would never pair again. This now also covers marker cuts, which are no longer discounted in that mode for the same reason. The overshoot is bounded to the one segment right after the short close, and the default mode never stretches. Left unclamped.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
getroot: Fixed. The settlement now drops a return point whose break was discarded in the same pass (the map order guarantees the OUT is visited first), which also keeps cut_this_segment honest. Covered by AReturnPointNeverOutlivesItsDiscardedBreak.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
getroot: Fixed as suggested. The reference publishes an opening anchor with its first stored chunk, so the synced tracks anchor immediately while the high-water mark still caps their emission. Media before the anchor joins the first slot instead of opening a short segment of its own.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
getroot: Fixed with validation at config load: SegmentDuration must be greater than 0. One correction: the integer division fails with SIGFPE rather than hanging, but the conclusion is the same.

src/modules/containers/mpegts/marker_box.cpp
getroot: Relaxed to one cadence on both the TS and LLHLS paths. The invariant only needs the break to span one cadence so the return cannot land on the very cut that opened it; a 15 s break at the 10 s default packages now.

Keukhan (commented):
Retracted - please ignore.

Keukhan (approved):
lgtm

src/base/provider/stream_test.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+TEST(ProviderStreamBreakTest, CueBreakSendsItsReturnPoint)
+{
+	EventCaptureStream stream;
+
+	auto out = CueEvent::Create(CueEvent::CueType::OUT, 12000);
+	ASSERT_TRUE(stream.SendDataFrame(5000, cmn::BitstreamFormat::CUE, cmn::PacketType::EVENT, out->Serialize(), true));
+
+	// The break and its return point
+	ASSERT_EQ(stream._sent.size(), 2u);
```
getroot: These break tests still expect the provider to pair a return point, but 8cdfa6c9d removed `SendProvisionalReturnPoint`. `SendDataFrame` now emits a single packet, so `_sent.size()` is 1 and both `CueBreakSendsItsReturnPoint` and `Scte35BreakFollowsItsAutoReturnFlag` fail.

Beyond the tests, the automatic return point is gone from the branch entirely: master synthesized it in `mpegts_packager.cpp` (`Marker::CreateMarker(CUE, ..., CueEvent::Create(CueType::IN, 0))`) after a CUE-OUT, and nothing here replaces it. Send a CUE-OUT with `duration` through the event API and `#EXT-X-CUE-OUT:DURATION=30.000` is emitted with no `#EXT-X-CUE-IN` ever following, so the break stays open and the next OUT is refused by the OUT-then-OUT rule.

Either the sender is now required to post its own return point (then these tests and the removed synthesis both need to go, and the API needs to emit the pair), or the synthesis needs a home in this PR.

getroot (commented):
A pass over the boundary policy split. The blocking one is the removed automatic return point: the provider break tests still assert the old behaviour and fail, and no code path emits a CUE-IN for a break that declares its own duration. After that, three correctness items stand out: the discontinuity flag ignores whether the break position was reached, the provider clock can never go back down, and the marker validate/commit sequence drops its lock in between. The rest are robustness and cleanup notes.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
```diff
@@ -0,0 +1,794 @@
...
+					  _log_context.CStr(), closed_at_us / 1000.0, target_segment_duration_us / 1000.0,
+					  buffered_duration_us / 1000.0, last_segment_duration_us / 1000.0,
+					  next_frame.dts_us / 1000.0, segment_start_us / 1000.0);
+			}
+		}
+
+		// A pending break rides whichever close comes first, whatever caused it.
+		// The track that broke closes once, so a rendition that closed twice
+		// would carry the break on a different segment number than the rest.
+		decision.discontinuity = (_pending_discontinuity_cut_us >= 0) && decision.completes_segment;
```
getroot: This sets the flag from the mere existence of a pending break, not from having reached it. `break_reached` at line 521 correctly requires `next_frame.dts_us >= _pending_discontinuity_cut_us`, and the empty-buffer path at line 526 honours it, but this line only checks `_pending_discontinuity_cut_us >= 0`.

Audio changes track at 10.000s and calls `RequestCutForDiscontinuity(10000)` on the video packager while video is still at 8.500s (a normal A/V lead). Video's next cadence completion at 9.000s sets `decision.discontinuity`, `Settle` clears the pending cut, and video stamps `EXT-X-DISCONTINUITY` a full segment before the content actually changes while audio stamps its own at 10.000s. The renditions then carry different `EXT-X-DISCONTINUITY-SEQUENCE` values, which is what this code exists to prevent.

Multiplying by `break_reached` here keeps the "rides whichever close comes first" behaviour while still requiring the position to have been reached.

src/base/provider/stream.cpp
```diff
@@ -142,6 +142,45 @@ namespace pvd
...
+		{
+			return;
+		}
+
+		ov::LockGuard lock(_timestamp_mutex);
+
+		// The clock is the newest position over every media track: a lagging
+		// track cannot pull it backward, and per-track timestamps only move
+		// forward, so a running maximum is that position
+		if (media_timestamp_ms > _last_media_timestamp_ms)
```
getroot: The clock is now a running maximum with no path that ever lowers or clears it, and the comment states "per-track timestamps only move forward" as an invariant. That holds between tracks of one continuous source, not across a source whose timestamps restart lower: a re-published RTMP stream reusing the Stream object, a scheduled-channel item whose base is re-derived, or a track added later with an independent epoch.

Once that happens `_last_media_timestamp_ms` is stuck at the old high-water mark, so `GetCurrentTimestampMs()` returns `old_max + wall-clock elapsed`. Every `SendDataFrame(-1)` event and every marker is then stamped far ahead of the media for the life of the stream, and `CanInsertMarker` refuses or mis-places all of them. Master assigned the value from a single master-clock track, so it followed the source down.

A reset on stream restart, or treating a large backward jump as a re-anchor rather than ignoring it, would close this.

src/publishers/llhls/llhls_stream.cpp
```diff
@@ -1938,12 +1876,68 @@ bool LLHlsStream::AddPackager(const std::shared_ptr<const MediaTrack> &media_tra
...
+		}
+		else
+		{
+			boundary_policy = std::make_shared<bmff::SyncedBoundaryPolicy>(_reference_boundary_policy, policy_config);
+		}
+	}
+	else
+	{
+		// Server-time numbering paces every segment as one wall-clock slot
+		boundary_policy = std::make_shared<bmff::DurationBoundaryPolicy>(policy_config, server_time_based_segment_numbering);
```
getroot, line 1942: `server_time_based_segment_numbering` reaches `DurationBoundaryPolicy` only. `ReferenceBoundaryPolicy` (line 1912) and `SyncedBoundaryPolicy` are constructed without it, so with `<ServerTimeBasedSegmentNumbering>true</ServerTimeBasedSegmentNumbering>` and `<SegmentationMode>synced</SegmentationMode>` the initial number is still seeded from the epoch but the wall-clock slot pacing is not there: no marker discount, no reset on a discontinuity, no catch-up to finish each slot.

A second server started an hour later derives a fresh epoch-based number and the two never pair again, which is the failure the `wall_clock_slot_pacing` comment in `duration_boundary_policy.h` says the flag exists to prevent. The combination is currently accepted with no warning.

src/modules/containers/bmff/fmp4_packager/fmp4_packager.cpp
```diff
@@ -337,164 +293,77 @@ namespace bmff
...
+			int64_t segment_start_timestamp = next_frame->GetDts();
+			if (last_segment != nullptr && last_segment->IsCompleted() == false && last_segment->GetPartialCount() > 0)
 			{
-				// Last partial segment
-				can_be_last_chunk = true;
+				segment_start_timestamp = last_segment->GetStartTimestamp();
 			}
-
-			if (return_cut == true)
+			else if (samples->GetTotalDuration() > 0)
```
getroot: The comment says "the samples opening the next segment", but the predicate reads the summed packet duration rather than the count. A track whose packets carry `GetDuration() == 0` (a passthrough source, or the first sample before the duration is filled in) has samples buffered with `GetTotalDuration() == 0`, so `segment_start_timestamp` falls back to `next_frame->GetDts()`.

`GetChunkPlan` then computes `target_segment_duration_us = boundary.end_us - segment_start_us` against the wrong origin and `end_if_closed_here_us` reads as 0, so the segment runs long until the `_segment_duration_us * 2` last-resort close fires. `samples->GetTotalCount() > 0` is the test that matches the intent.

src/modules/containers/bmff/fmp4_packager/duration_boundary_policy.cpp
```diff
@@ -0,0 +1,96 @@
...
```
getroot, line 63: This ledger has no clamp, while the equivalent one on the mpegts side gets `_next_target_duration_ms = std::min(_next_target_duration_ms, target_duration_ms)` at `mpegts_packager.cpp:666`, and the storage's old `GetDurationMs() > segment_duration_ms * 2` force-complete net was removed from `fmp4_storage.cpp` in this PR.

With a 6s target, a keyframe famine producing one 20s segment leaves `_target_segment_duration_us = -8s`. `GetSegmentBoundary` then returns `boundary.end_us < segment_start_us`, so in `GetChunkPlan` the target is negative and every arriving keyframe satisfies `end_if_closed_here_us >= target - kPlanSlackUs`. Every following segment closes at its first keyframe until the ledger repays the 8s, producing a run of one-GOP segments in the playlist.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
```diff
@@ -0,0 +1,794 @@
...
+	std::optional<SegmentBoundaryPolicy::PreparedMarker> SegmentBoundaryPolicy::PrepareMarker(const std::shared_ptr<Marker> &marker) const
+	{
+		auto [can_insert, message] = CanInsertMarker(marker);
+		if (can_insert == false)
+		{
+			logte("%s - Failed to insert the marker : %s", _log_context.CStr(), message.CStr());
+			return std::nullopt;
+		}
+
+		std::shared_lock<std::shared_mutex> lock(_markers_guard);
```
getroot: The validate-then-commit sequence holds no lock across itself: `CanInsertMarker` takes a shared lock and releases it (line 103), `PrepareMarker` re-takes a shared lock here and releases it, then `CommitMarker` takes the exclusive lock (line 266) and re-validates nothing before writing `_pending_markers`, `_last_inserted_marker` and `_last_inserted_cut_us`.

Two ad-break OUTs posted concurrently through the API both read `_last_inserted_marker == nullptr`, both pass the "First marker must be OUT marker" check, and both commit. The result is an OUT-then-OUT chain the validator is written to make impossible, so the following IN is refused and the break never closes.

Validating and committing under one exclusive lock closes this, with `PrepareMarker` kept for the multi-track two-phase case where the decision genuinely has to be made before any track applies it.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.h
```diff
@@ -0,0 +1,312 @@
...
+		// the marker insertion path becomes when it is driven from an API.
+		std::atomic<int64_t> _newest_sample_end_us{-1};
+
+		// The first sample position this track saw; written once, on the same
+		// pass and for the same reason as the newest one
+		std::atomic<int64_t> _first_sample_end_us{-1};
+
+		// Where a propagated timeline break waits to cut, and where the last one
+		// landed (the dedup floor); media-thread only
+		int64_t _pending_discontinuity_cut_us = -1;
```
getroot: These two are plain `int64_t` documented as media-thread only, but `LLHlsStream::OnTrackChanged` calls `RequestCutForDiscontinuity` on the other tracks' packagers from the notification thread, which reaches `RequestDiscontinuity` and assigns `_pending_discontinuity_cut_us` (segment_boundary_policy.cpp:317) and reads `_last_discontinuity_us` (line 302). The owning packager thread concurrently reads the pending cut at line 521 and writes it in `Settle` (line 476).

That is an unsynchronized race on non-atomic members, and the media-thread-only note above them is not true. Every other cross-thread field in this class (`_newest_sample_end_us`, `_pending_marker_count`) was made atomic for exactly this reason. The visible symptom is a missed or duplicated discontinuity cut on the rendition that did not change.

src/modules/containers/bmff/fmp4_packager/fmp4_storage.cpp
```diff
@@ -27,16 +29,11 @@ namespace bmff
...
-		_target_segment_duration_ms = static_cast<int64_t>(_config.segment_duration_ms);
 		_stream_tag = stream_tag;
 
-		_initial_segment_number = 0;
-		if (_config.server_time_based_segment_numbering == true)
-		{
-			// last segment number = current epoch time / segment duration
-			_initial_segment_number = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count() / _target_segment_duration_ms;
-		}
+		OV_ASSERT2(boundary_policy != nullptr);
```
getroot, line 34: `OV_ASSERT2` compiles away in release builds and the next line dereferences the pointer immediately; `CreateNextSegment`, `AppendMediaChunk` and `CompleteLastSegment` all do the same with no check. Meanwhile `FMP4Packager::AppendSample` guards the identical pointer with `if (_boundary_policy != nullptr)` and `RequestCutForDiscontinuity` returns early on null.

The two halves of the same subsystem disagree on whether null is a legal state. Worth picking one: either make it a hard precondition and drop the packager's checks, or null-check here too.

src/modules/containers/mpegts/marker_box.cpp
```diff
@@ -0,0 +1,271 @@
...
+				++it;
+			}
+		}
+
+		return markers;
+	}
+
+	void MarkerBox::SetCutCadenceMs(double segment_duration_ms, double keyframe_interval_ms)
+	{
+		_cut_cadence_ms = segment_duration_ms;
```
getroot: `_cut_cadence_ms` is written here with no synchronization, from the packager constructor and again from the track-setup path at `mpegts_packager.cpp:130` when the main track becomes known, while `CanInsertMarker` reads it at lines 31 and 41 from the API thread holding only the shared lock the writer never takes.

`_media_position_ms` in this class was made `std::atomic<double>` for the same pattern, so this one looks accidental. A marker posted during track setup can validate against a stale cadence and be accepted with a gap below the real minimum.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
```diff
@@ -0,0 +1,794 @@
...
+		{
+			size_t samples_before_marker = 0;
+			int64_t duration_before_marker_us = 0;
+			while (samples_before_marker < sample_list.size() && sample_list[samples_before_marker].timing.dts_us < marker_cut_us)
+			{
+				duration_before_marker_us += sample_list[samples_before_marker].timing.duration_us;
+				samples_before_marker++;
+			}
+
+			if (samples_before_marker > 0)
```
getroot: `marker_cut_now` fires when `next_frame.dts_us >= marker_cut_us || buffer_end_us > marker_cut_us`, but the trim is applied only under `samples_before_marker > 0`. When the segment's start came from the already-stored partial (so `pending_cut_us > segment_start_us` still holds) and every buffered sample sits at or after the CUE-OUT position, `samples_before_marker == 0`, `emit_count` stays at the full buffer, and the segment still completes via `marker_cut_now`.

The whole buffer, content that belongs after the break, is written into the segment that is supposed to end exactly at the announced position. An ad insertor replacing from that boundary then overwrites content it should not, which is the one guarantee `immediate` mode provides. Emitting 0 samples and completing on the stored part is what the comment above describes.

src/base/common_types.h
```diff
@@ -123,6 +123,27 @@ enum class RtcBWEType : uint8_t
...
+	// boundaries it realizes, so the segment grid and numbering agree
+	// across tracks
+	Synced,
+};
+
+enum class LLHlsCueOutCutMode : uint8_t
+{
+	// Cut right at the CUE-OUT even off a non-keyframe; an ad insertor
+	// replaces the break content anyway (server-side ad insertion)
+	Immediate,
```
getroot: `Immediate` is the zero value here, but every default in the code is `Keyframe`: `SegmentBoundaryPolicy::Config::cue_out_cut_mode` and the config member in `ll_hls_publisher.h` both initialize to it. `LLHlsSegmentationMode` right above gets this right, with its default `Duration` first.

A `Config{}` built without setting the field, or any zero-initialized instance, silently means the opposite of the documented default and cuts CUE-OUTs mid-GOP. Swapping the two enumerators makes the zero value agree with the default everywhere.

src/config/items/virtual_hosts/applications/publishers/ll_hls_publisher.h
```diff
@@ -47,6 +55,43 @@ namespace cfg
 						Register<Optional>("PartHoldBack", &_part_hold_back);
 						Register<Optional>("EnablePreloadHint", &_enable_preload_hint);
 						Register<Optional>({"DRM", "drm"}, &_drm);
+
+						Register<Optional>("SegmentationMode", &_segmentation_mode, [=]() -> std::shared_ptr<ConfigError> { return nullptr; }, [=]() -> std::shared_ptr<ConfigError> {
```
getroot: The no-op lambda is only here to reach the fourth parameter. `Register` already declares `OptionalCallback optional_callback = nullptr` (config/engine/item.h), and every existing site in the tree passes `nullptr` there, e.g. `Register<Optional>("MinBitrate", &_min_bitrate_string, nullptr, [=]() { ... })` in `alert/rules/ingress/ingress.h`.

Same at line 79 and in `hls_publisher.h`. Each one costs a heap-allocating `std::function` and an implicit `this` capture for behaviour identical to `nullptr`.

src/modules/containers/bmff/fmp4_packager/fmp4_storage.cpp
```diff
@@ -557,61 +565,56 @@ namespace bmff
...
+			auto completion_result = (discontinuity == true) ? _boundary_policy->OnDiscontinuity(completed_segment)
+															 : _boundary_policy->OnSegmentCompleted(completed_segment);
+
+			// Markers realized on boundaries decided elsewhere, in addition to
+			// the ones the packager consumed from its own sample range
+			segment->AddMarkers(completion_result.markers);
+
 			segment->SetCompleted();
+
 			CreateNextSegment();
```
getroot, line 641: The new `first_chunk_start_timestamp_us` parameter is only reachable for the very first segment. `AppendMediaChunk` passes it on the `segment == nullptr || IsCompleted()` branch (line 576), but this line pre-creates the next segment with the default `std::nullopt` right after each completion, and `CompleteLastSegment` (line 367) does the same, so from the second segment on the parameterized call is never taken and `_next_segment_number` alone does the naming.

The `std::optional` then threads through `SegmentBoundary GetSegmentBoundary(std::optional<int64_t>)` and adds a branch to every policy for a case that fires once per stream. Either pass the start on the pre-create too, or drop the parameter.

src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.h
```diff
@@ -0,0 +1,312 @@
...
+			// segment gained a boundary
+			bool cut_this_segment = false;
+		};
+
+		// Consume the markers this settlement takes over. A position the media
+		// never reached (a hole left by the keyframe wait after a track change)
+		// still rides the next segment, up to a segment length behind; older
+		// than that the advertised time no longer describes where it landed, so
+		// it is discarded.
+		SettledMarkers PopMarkersForSettlement(int64_t start_us, int64_t end_us);
```
getroot, line 277: This six-line block, and several others in the new policy files (lines 23-47, 96-110, 157-161, 254-259, and in `synced_boundary_policy.cpp`), read as design notes rather than code comments. The project guide asks for the core point in a line or two, with the derivation left out.

The cost is that the header becomes a document that has to be re-verified against the code on every change.

PR #2304, opened 2026-08-22
fix(config): unify `webRtcAutoAbr` JSON key casing with a deprecated input alias

dimiden:
## Summary
The playlist option `WebRtcAutoAbr` was exposed with two different JSON key casings: the configuration API auto-derived `webRtcAutoAbr` from the XML name, while the handwritten serializers (stream query response, multiplex channel API, OVT signaling) use `webrtcAutoAbr`.
This PR unifies the JSON key on `webrtcAutoAbr`, which follows the all-lowercase `webrtc` convention used by every other WebRTC-related JSON key.
During the deprecation window the old key is accepted on requests and emitted in responses alongside the new key, and it will be removed in a future release.
XML configuration (`<WebRtcAutoAbr>`) is unchanged.

## Changes
- Add a deprecated JSON alias to the config engine: `ItemName` gains a `deprecated_json_name` and a 3-argument constructor, and the unknown-item check accepts the alias
- Input rules: when only the deprecated key is present, `DataSource::GetValue()` falls back to it and logs a deprecation warning with the item path and file name; when both keys are present with different values (an explicit null counts as a value), the request is rejected with a ConfigError so a stale value can never silently win over the client's intent; an equal pair passes silently since responses emit both keys and an echo is expected
- Output rule: `Item::ToJson()` emits the deprecated key with the same value alongside the canonical key during the deprecation window, so response-only readers keep working
- Registration safety: the alias is restricted to scalar leaf values and must not collide with its own canonical name or a JSON key another child already uses; violations throw a ConfigError, which fails server startup deterministically because the startup schema validation rebuilds every item type reachable from cfg::Server
- Engine cleanups from review: commit the rebuild guard only after `MakeList()` succeeds, and extract the `MakeChildPath()`, `IsScalarValueType()`, and `HasJsonMember()` helpers
- Annotate the propagation path of the new throw with MAY_THROWS, in separate commits
- Update the `Options` schema in `api/ome.yml` (`webrtcAutoAbr` plus the deprecated alias with a description of the window behavior) and the transcode webhook docs
- Correct stale comments in `ovt_signaling.h`: the receiver requires the `options` object and assumes `false` when `webrtcAutoAbr` is omitted

## Configuration
`Server.xml` is not affected; `<WebRtcAutoAbr>` works as before.
Only the JSON key of the REST API (and the transcode webhook response) changes.

Requests accept both spellings; a request that sends both keys with different values is rejected with 400:

```json
"options": { "webrtcAutoAbr": true }
```

```json
"options": { "webRtcAutoAbr": true }
```

Responses contain both keys with the same value during the deprecation window:

```json
"options": { "webrtcAutoAbr": true, "webRtcAutoAbr": true, "hlsChunklistPathDepth": -1, "enableTsPackaging": false }
```

## Testing
Tested with a debug build, using only curl and ffmpeg.

1. Start OME with the API server enabled and create a vhost/app via the API.
2. `POST /v1/vhosts/{vh}/apps/{app}/outputProfiles` with a playlist option in each form:
   - `{"webrtcAutoAbr": true}` -> 200, no warning
   - `{"webRtcAutoAbr": false}` -> 200, the value is applied, and the server logs `The JSON key "playlists.options.webRtcAutoAbr" is deprecated. Use "webrtcAutoAbr" instead`
   - `{"webrtcAutoAbr": true, "webRtcAutoAbr": true}` -> 200 with no warning (echo case)
   - `{"webrtcAutoAbr": true, "webRtcAutoAbr": false}` -> 400 (conflicting values)
   - `{"webrtcAutoAbr": null, "webRtcAutoAbr": false}` -> 400 (null counts as a value)
   - a typo key such as `{"webrtcautoabr": true}` -> still rejected with 400 `Unknown item found: playlists.options.webrtcautoabr`
3. `GET` the profile and the app: every response contains both keys with the same value.
4. XML regression: OME starts normally with `<WebRtcAutoAbr>true</WebRtcAutoAbr>` in `Server.xml`.
5. TranscodeWebhook: with an ffmpeg RTMP input and a local webhook server, a response whose `outputProfiles` uses either spelling creates the stream successfully.
6. Regressions on the handwritten paths: the stream query response key, the multiplex channel API (`.mux` file content and channel query response), and OVT pull propagation of the flag are unchanged.

### DIFF
--- a/api/ome.yml
+++ b/api/ome.yml
@@ -639,8 +639,13 @@ components:
     Options:
       type: object
       properties:
+        webrtcAutoAbr:
+          type: boolean
+        # Deprecated alias of webrtcAutoAbr. It is still accepted on requests for backward compatibility
+        # but is no longer included in responses, and will be removed in a future release.
         webRtcAutoAbr:
           type: boolean
+          deprecated: true
         hlsChunklistPathDepth:
           type: integer
           format: int32
--- a/api/ome.yml
+++ b/api/ome.yml
@@ -644,6 +644,7 @@ components:
         webRtcAutoAbr:
           type: boolean
           deprecated: true
+          writeOnly: true
           description: Deprecated alias of webrtcAutoAbr. It is still accepted on requests for backward compatibility (ignored when webrtcAutoAbr is also present) but is no longer included in responses, and will be removed in a future release.
         hlsChunklistPathDepth:
           type: integer
--- a/api/ome.yml
+++ b/api/ome.yml
@@ -644,8 +644,7 @@ components:
         webRtcAutoAbr:
           type: boolean
           deprecated: true
-          writeOnly: true
-          description: Deprecated alias of webrtcAutoAbr. It is still accepted on requests for backward compatibility (ignored when webrtcAutoAbr is also present) but is no longer included in responses, and will be removed in a future release.
+          description: Deprecated alias of webrtcAutoAbr. During the deprecation window it is accepted on requests (a request that sends both keys with different values is rejected) and included in responses alongside webrtcAutoAbr, and will be removed in a future release.
         hlsChunklistPathDepth:
           type: integer
           format: int32
--- a/api/ome.yml
+++ b/api/ome.yml
@@ -644,7 +644,7 @@ components:
         webRtcAutoAbr:
           type: boolean
           deprecated: true
-          description: Deprecated alias of webrtcAutoAbr. It is still accepted on requests for backward compatibility but is no longer included in responses, and will be removed in a future release.
+          description: Deprecated alias of webrtcAutoAbr. It is still accepted on requests for backward compatibility (ignored when webrtcAutoAbr is also present) but is no longer included in responses, and will be removed in a future release.
         hlsChunklistPathDepth:
           type: integer
           format: int32
--- a/api/ome.yml
+++ b/api/ome.yml
@@ -641,11 +641,10 @@ components:
       properties:
         webrtcAutoAbr:
           type: boolean
-        # Deprecated alias of webrtcAutoAbr. It is still accepted on requests for backward compatibility
-        # but is no longer included in responses, and will be removed in a future release.
         webRtcAutoAbr:
           type: boolean
           deprecated: true
+          description: Deprecated alias of webrtcAutoAbr. It is still accepted on requests for backward compatibility but is no longer included in responses, and will be removed in a future release.
         hlsChunklistPathDepth:
           type: integer
           format: int32
--- a/docs/transcoding/transcodewebhook.md
+++ b/docs/transcoding/transcodewebhook.md
@@ -249,7 +249,7 @@ The `outputProfiles` section in the JSON structure mirrors the configuration in 
           "name": "abr",
           "options": {
             "enableTsPackaging": true,
-            "webRtcAutoAbr": true,
+            "webrtcAutoAbr": true,
             "hlsChunklistPathDepth": -1
           },
           "renditions": [
@@ -280,6 +280,8 @@ The `outputProfiles` section in the JSON structure mirrors the configuration in 
   ]
 }
 ```
+
+> **Note**: `webRtcAutoAbr` is a deprecated alias of `webrtcAutoAbr`. It is still accepted for backward compatibility, and will be removed in a future release.
 
 #### tracksets (optional)
 
--- a/docs/transcoding/transcodewebhook.md
+++ b/docs/transcoding/transcodewebhook.md
@@ -281,7 +281,7 @@ The `outputProfiles` section in the JSON structure mirrors the configuration in 
 }
 ```
 
-> **Note**: `webRtcAutoAbr` is a deprecated alias of `webrtcAutoAbr`. It is still accepted for backward compatibility, and will be removed in a future release.
+> **Note**: `webRtcAutoAbr` is a deprecated alias of `webrtcAutoAbr`. It is still accepted for backward compatibility (ignored when `webrtcAutoAbr` is also present), and will be removed in a future release.
 
 #### tracksets (optional)
 
--- a/docs/transcoding/transcodewebhook.md
+++ b/docs/transcoding/transcodewebhook.md
@@ -281,7 +281,7 @@ The `outputProfiles` section in the JSON structure mirrors the configuration in 
 }
 ```
 
-> **Note**: `webRtcAutoAbr` is a deprecated alias of `webrtcAutoAbr`. It is still accepted for backward compatibility (ignored when `webrtcAutoAbr` is also present), and will be removed in a future release.
+> **Note**: `webRtcAutoAbr` is a deprecated alias of `webrtcAutoAbr`. During the deprecation window it is accepted for backward compatibility (a response that contains both keys with different values is rejected), and will be removed in a future release.
 
 #### tracksets (optional)
 
--- a/src/config/engine/data_source.cpp
+++ b/src/config/engine/data_source.cpp
@@ -698,8 +698,34 @@ namespace cfg
 			case DataType::Xml:
 				return GetValueFromXml(value_type, name.GetName(_type), true, resolve_path, original_value);
 
-			case DataType::Json:
-				return GetValueFromJson(value_type, name.GetName(_type), true, resolve_path, omit_json, original_value);
+			case DataType::Json: {
+				const auto &json_name = name.GetName(_type);
+
+				if (
+					// This item has a deprecated alias
+					(name.deprecated_json_name.IsEmpty() == false) &&
+					// `isMember()` below is only valid on an object
+					_json.isObject())
+				{
+					if (_json.isMember(json_name.CStr()))
+					{
+						// The current name is present, so use it as-is.
+						// Warn if the deprecated key is also present, because it is ignored.
+						if (_json.isMember(name.deprecated_json_name.CStr()))
+						{
+							logtw("The deprecated JSON key \"%s\" is ignored because \"%s\" is also present", name.deprecated_json_name.CStr(), json_name.CStr());
+						}
+					}
+					else if (_json.isMember(name.deprecated_json_name.CStr()))
+					{
+						// The current name is absent - fall back to the deprecated JSON name
+						logtw("The JSON key \"%s\" is deprecated. Use \"%s\" instead", name.deprecated_json_name.CStr(), json_name.CStr());
+						return GetValueFromJson(value_type, name.deprecated_json_name, true, resolve_path, omit_json, original_value);
+					}
+				}
+
+				return GetValueFromJson(value_type, json_name, true, resolve_path, omit_json, original_value);
+			}
 		}
 
 		OV_ASSERT2(false);
--- a/src/config/engine/data_source.cpp
+++ b/src/config/engine/data_source.cpp
@@ -704,37 +704,35 @@ namespace cfg
 				if (
 					// This item has a deprecated alias
 					(name.deprecated_json_name.IsEmpty() == false) &&
-					// `isMember()` below is only valid on an object
+					// `isMember()` below asserts on non-object values other than null
 					_json.isObject())
 				{
-					// Full path of the deprecated key for the warnings below, such as "playlists.options.webRtcAutoAbr in Server.json"
-					auto deprecated_key_location = [&]() -> ov::String {
-						ov::String location = ov::String::FormatString(
+					// Full path of the deprecated key for the warnings below, such as "playlists.options.webRtcAutoAbr"
+					auto deprecated_key_path = [&]() -> ov::String {
+						return ov::String::FormatString(
 							"%s%s%s",
 							item_path.CStr(), item_path.IsEmpty() ? "" : ".", name.deprecated_json_name.CStr());
-
+					};
+					// " in <file>" when the JSON comes from a file, empty otherwise
+					auto file_suffix = [&]() -> ov::String {
 						auto file_name = GetFileName();
-						if (file_name.IsEmpty() == false)
-						{
-							location.AppendFormat(" in %s", file_name.CStr());
-						}
-
-						return location;
+						return file_name.IsEmpty() ? ov::String() : ov::String::FormatString(" in %s", file_name.CStr());
 					};
 
 					if (_json.isMember(json_name.CStr()))
 					{
-						// The current name is present, so use it as-is.
+						// The current name is present, so use it as-is. Key presence decides, not the value:
+						// an explicit null under the current name still wins over the deprecated key.
 						// Warn if the deprecated key is also present, because it is ignored.
 						if (_json.isMember(name.deprecated_json_name.CStr()))
 						{
-							logtw("The deprecated JSON key \"%s\" is ignored because \"%s\" is also present", deprecated_key_location().CStr(), json_name.CStr());
+							logtw("The deprecated JSON key \"%s\" is ignored because \"%s\" is also present%s", deprecated_key_path().CStr(), json_name.CStr(), file_suffix().CStr());
 						}
 					}
 					else if (_json.isMember(name.deprecated_json_name.CStr()))
 					{
 						// The current name is absent - fall back to the deprecated JSON name
-						logtw("The JSON key \"%s\" is deprecated. Use \"%s\" instead", deprecated_key_location().CStr(), json_name.CStr());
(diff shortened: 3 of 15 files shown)


src/config/engine/item_name.h
```diff
@@ -32,6 +32,10 @@ namespace cfg
 	public:
 		ItemName(const char *xml_name);
 		ItemName(const char *xml_name, const char *json_name);
+		// deprecated_json_name is an old JSON name kept for backward compatibility.
+		// It is still accepted on input (never emitted on output), and will be removed in a future release.
+		// Supported only for scalar leaf values; see the assertion in `Item::AddChild()`.
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/config/engine/item.cpp
```diff
@@ -237,6 +237,29 @@ namespace cfg
...
+				(type == ValueType::Integer) ||
+				(type == ValueType::Long) ||
+				(type == ValueType::Boolean) ||
+				(type == ValueType::Double) ||
+				(type == ValueType::Text));
+#endif	// DEBUG
+
+			// Register the deprecated JSON name as well so that the unknown item check accepts it
+			_children_for_json.insert_or_assign(name.deprecated_json_name, child);
+		}
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/config/engine/data_source.cpp
```diff
@@ -691,15 +691,54 @@ namespace cfg
...
+					};
+
+					if (_json.isMember(json_name.CStr()))
+					{
+						// The current name is present, so use it as-is. Key presence decides, not the value:
+						// an explicit null under the current name still wins over the deprecated key.
+						// Warn if the deprecated key is also present, because it is ignored.
+						if (_json.isMember(name.deprecated_json_name.CStr()))
+						{
+							logtw("The deprecated JSON key \"%s\"%s is ignored because \"%s\" is also present", deprecated_key_path().CStr(), file_suffix().CStr(), json_name.CStr());
```
getroot: Since every serialized output now carries the canonical key, an echo and modify flow from a pre-rename client naturally produces both keys: GET returns `webrtcAutoAbr`, the old client sets `webRtcAutoAbr` on its copy and sends it back via PATCH, which replaces the profile verbatim, so the stale canonical value wins while the client gets 200. Suggest treating a value conflict as an error: if both keys are present and the values differ, throw a ConfigError so the API returns 400 and a webhook response fails loudly. If the values are equal, keep the current warning.

src/config/engine/data_source.cpp
```diff
@@ -691,15 +691,54 @@ namespace cfg
...
+					};
+					// " in <file>" when the JSON comes from a file, empty otherwise
+					auto file_suffix = [&]() -> ov::String {
+						auto file_name = GetFileName();
+						return file_name.IsEmpty() ? ov::String() : ov::String::FormatString(" in %s", file_name.CStr());
+					};
+
+					if (_json.isMember(json_name.CStr()))
+					{
+						// The current name is present, so use it as-is. Key presence decides, not the value:
```
getroot: Key presence deciding the winner means an explicit null under `webrtcAutoAbr` suppresses a real value under `webRtcAutoAbr`, and the engine then treats the child as unconfigured and applies the default. Everywhere else the engine treats null as absent, so consider falling back to the deprecated key when the canonical value is null.

api/ome.yml
```diff
@@ -639,8 +639,13 @@ components:
     Options:
       type: object
       properties:
+        webrtcAutoAbr:
+          type: boolean
         webRtcAutoAbr:
           type: boolean
+          deprecated: true
+          writeOnly: true
+          description: Deprecated alias of webrtcAutoAbr. It is still accepted on requests for backward compatibility (ignored when webrtcAutoAbr is also present) but is no longer included in responses, and will be removed in a future release.
```
getroot: The input side gets a deprecation window but the output side flips in the same release, and not only for the REST API. Monitoring event payloads and alert webhook payloads serialize output profiles through the same ToJson path, so consumers reading `webRtcAutoAbr` there break silently. Consider emitting both keys during the deprecation window, or at least documenting those surfaces as well.

src/config/engine/data_source.cpp
```diff
@@ -691,15 +691,54 @@ namespace cfg
...
+						// Warn if the deprecated key is also present, because it is ignored.
+						if (_json.isMember(name.deprecated_json_name.CStr()))
+						{
+							logtw("The deprecated JSON key \"%s\"%s is ignored because \"%s\" is also present", deprecated_key_path().CStr(), file_suffix().CStr(), json_name.CStr());
+						}
+					}
+					else if (_json.isMember(name.deprecated_json_name.CStr()))
+					{
+						// The current name is absent - fall back to the deprecated JSON name
+						logtw("The JSON key \"%s\"%s is deprecated. Use \"%s\" instead", deprecated_key_path().CStr(), file_suffix().CStr(), json_name.CStr());
```
getroot, line 739: This warning fires on every parse, and OutputProfiles are parsed per stream creation via the transcode webhook, so an unmigrated control server floods the log proportionally to stream churn. Consider suppressing it to once per key.

src/config/engine/item_name.h
```diff
@@ -32,6 +32,10 @@ namespace cfg
 	public:
 		ItemName(const char *xml_name);
 		ItemName(const char *xml_name, const char *json_name);
+		// deprecated_json_name is an old JSON name kept for backward compatibility.
+		// It is still accepted on input (never emitted on output), and will be removed in a future release.
+		// Supported only for scalar leaf values; `Item::AddChild()` throws a ConfigError otherwise.
+		ItemName(const char *xml_name, const char *json_name, const char *deprecated_json_name);
```
getroot, line 39: Heads up for the next alias: the generic PATCH merge in the API helpers (OverwriteJson) merges the client body over a ToJson snapshot that already contains the canonical key, so a client sending only the deprecated key ends up with both keys and its value loses. `webrtcAutoAbr` never reaches that path today, but any future alias on an app or vhost level scalar will.

src/config/engine/item.cpp
```diff
@@ -224,6 +236,63 @@ namespace cfg
...
+				(type == ValueType::String) ||
+				(type == ValueType::Integer) ||
+				(type == ValueType::Long) ||
+				(type == ValueType::Boolean) ||
+				(type == ValueType::Double) ||
+				(type == ValueType::Text);
+
+			if (supported == false)
+			{
+				// Fail loudly in every build. The startup schema validation (ValidateOmitJsonNameRules)
```
getroot: This guarantee is positional rather than structural: ValidateOmitJsonNameRules only walks types reachable from cfg::Server, so a future standalone-loaded item root would turn this into a lazy runtime throw instead of a startup refusal.

src/config/engine/item.cpp
```diff
@@ -203,7 +203,19 @@ namespace cfg
 
 			_last_target = this;
 
-			MakeList();
+			try
```
getroot: MakeList overrides only call Register and never re-enter RebuildListIfNeeded, so this try/catch reduces to committing the guard after the call: `MakeList(); _last_target = this;`. On throw the guard keeps its old value and the next call reruns from scratch.

src/config/engine/declare_utilities.h
```diff
@@ -10,6 +10,7 @@
 
 // auto &GetInt() { return int_value; }
 #define CFG_DECLARE_REF_GETTER_OF(getter_name, variable_name) \
+	MAY_THROWS(cfg::ConfigError)                              \
```
getroot, line 13: MAY_THROWS expands to nothing, and this sweep touches about 25 declarations across item.h, list_interface.h, and json_serdes/application.h, which widens the rebuild and buries the PR's real changes. Suggest annotating only the new throw sources and moving the rest to a separate documentation pass.

src/config/engine/data_source.cpp
```diff
@@ -691,15 +691,54 @@ namespace cfg
...
+				const auto &json_name = name.GetName(_type);
+
+				if (
+					// This item has a deprecated alias
+					(name.deprecated_json_name.IsEmpty() == false) &&
+					// `isMember()` below asserts on non-object values other than null
+					_json.isObject())
+				{
+					// Full path of the deprecated key for the warnings below, such as "playlists.options.webRtcAutoAbr"
+					auto deprecated_key_path = [&]() -> ov::String {
```
getroot, line 714: This is the third copy of the dotted child path join (child.cpp:115 and item.cpp:438 have the others), and `item_path` threads through GetValue only for this diagnostic. Either extract a shared helper, or emit the warning from Child::SetValue, which already builds the identical child_path right after this call. That also frees GetIncludeFileList from passing a dummy empty string.

src/config/engine/data_source.cpp
```diff
@@ -691,15 +691,54 @@ namespace cfg
...
 		{
 			case DataType::Xml:
 				return GetValueFromXml(value_type, name.GetName(_type), true, resolve_path, original_value);
 
-			case DataType::Json:
-				return GetValueFromJson(value_type, name.GetName(_type), true, resolve_path, omit_json, original_value);
+			case DataType::Json: {
+				const auto &json_name = name.GetName(_type);
+
+				if (
```
getroot: GetJsonValue in this file already encapsulates the isObject guard for isMember. A HasJsonMember sibling next to it would express this block as two calls and keep the invariant in one place instead of a comment.

src/config/engine/item.cpp
```diff
@@ -224,6 +236,63 @@ namespace cfg
...
+		// and _children inconsistent with each other
+		if (name.deprecated_json_name.IsEmpty() == false)
+		{
+			// The deprecated JSON alias is restricted to scalar leaf values:
+			// an Attribute value lives under "$" so the alias lookup never matches it,
+			// a List child propagates the alias name into the element DataSources
+			// and `IsSourceOf()` silently drops every element,
+			// and an Item child is rejected as well because no use case needs that path and it is untested.
+			const auto type = child->GetType();
+			const bool supported =
```
getroot: This is the fourth hand-written enumeration of the scalar ValueType set (item.cpp:562, item.cpp:461, list_interface.cpp:48). An IsScalarValueType() next to StringFromValueType would keep the classification in one file, so a future ValueType addition cannot silently miss this list.

src/config/engine/item.cpp
```diff
@@ -224,6 +236,63 @@ namespace cfg
...
+			{
+				throw CreateConfigError("The deprecated JSON alias \"%s\" of %s conflicts with a JSON key already used by %s", name.deprecated_json_name.CStr(), name.ToString().CStr(), existing->second->GetItemName().ToString().CStr());
+			}
+		}
+
+		{
+			// A canonical name must not displace another child's deprecated alias:
+			// the same JSON key would then feed two children (one directly, one via the alias fallback)
+			// with no diagnostic anywhere
+			auto existing = _children_for_json.find(name.json_name);
```
getroot: This runs for every child on every MakeList rerun even though only one child in the tree has an alias, and it duplicates the hash lookup that insert_or_assign repeats below. Reusing the find iterator for the insert, or gating on an aliased-child flag, makes it free. Minor, the cost is well under 1% of AddChild.

src/config/engine/data_source.cpp
dimiden: Keeping the per-parse warning. During the migration window we prefer every occurrence to stay visible, and the volume is bounded by requests that still use the deprecated key.

src/config/engine/declare_utilities.h
dimiden: The annotations are limited to the propagation path of the throw this PR introduces, so they belong to this change; they are isolated in separate commits to keep the main diff reviewable.

src/config/engine/item_name.h
dimiden: With the value-conflict check adopted above, that future case fails with 400 instead of silently dropping the client value.

src/config/engine/data_source.cpp
dimiden: Fixed.

api/ome.yml
dimiden: Fixed.

src/config/engine/data_source.cpp
dimiden: Fixed.

src/config/engine/data_source.cpp
dimiden: Fixed.

src/config/engine/item.cpp
dimiden: Fixed.

src/config/engine/item.cpp
dimiden: Fixed.

src/config/engine/item.cpp
dimiden: Fixed.

src/config/engine/item.cpp
dimiden: Fixed.

src/config/items/virtual_hosts/applications/output_profiles/playlist/options.h
```diff
@@ -40,7 +40,10 @@ namespace cfg
 				protected:
 					void MakeList() override
 					{
-						Register<Optional>("WebRtcAutoAbr", &_webrtc_auto_abr);
+						// The JSON name "webrtcAutoAbr" is specified explicitly to follow the all-lowercase "webrtc" convention.
+						// Deprecated: during the deprecation window the auto-derived old name "webRtcAutoAbr" is accepted
+						// on input and emitted on output alongside the new name, and will be removed in a future release.
+						Register<Optional>({"WebRtcAutoAbr", "webrtcAutoAbr", "webRtcAutoAbr"}, &_webrtc_auto_abr);
```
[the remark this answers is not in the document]
dimiden: The PR description was stale, fixed.

api/ome.yml
```diff
@@ -639,8 +639,12 @@ components:
     Options:
       type: object
       properties:
+        webrtcAutoAbr:
+          type: boolean
         webRtcAutoAbr:
           type: boolean
+          deprecated: true
+          description: Deprecated alias of webrtcAutoAbr. During the deprecation window it is accepted on requests (a request that sends both keys with different values is rejected) and included in responses alongside webrtcAutoAbr, and will be removed in a future release.
```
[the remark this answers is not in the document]
dimiden: The PR description was stale, fixed.

docs/transcoding/transcodewebhook.md
```diff
@@ -281,6 +281,8 @@ The `outputProfiles` section in the JSON structure mirrors the configuration in
 }
 ```
 
+> **Note**: `webRtcAutoAbr` is a deprecated alias of `webrtcAutoAbr`. During the deprecation window it is accepted for backward compatibility (a response that contains both keys with different values is rejected), and will be removed in a future release.
```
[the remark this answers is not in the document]
dimiden: The PR description was stale, fixed.

src/config/engine/data_source.cpp
```diff
@@ -691,15 +697,52 @@ namespace cfg
...
+						{
+							throw CreateConfigError("The deprecated JSON key \"%s\"%s conflicts with \"%s\": the two values differ", deprecated_key_path().CStr(), file_suffix().CStr(), json_name.CStr());
+						}
+					}
+					else if (has_deprecated)
+					{
+						// The current name is absent - fall back to the deprecated JSON name
+						logtw("The JSON key \"%s\"%s is deprecated. Use \"%s\" instead", deprecated_key_path().CStr(), file_suffix().CStr(), json_name.CStr());
+						return GetValueFromJson(value_type, name.deprecated_json_name, true, resolve_path, omit_json, original_value);
+					}
```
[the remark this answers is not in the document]
dimiden: The PR description was stale, fixed.

src/config/engine/item.cpp
```diff
@@ -655,7 +726,23 @@ namespace cfg
...
+							if (original_value)
+							{
+								SetOriginalJsonValue(value_type, value, deprecated_json_name, child->GetOriginalValue());
+							}
+							else
+							{
+								SetCurrentJsonValue(value_type, value, deprecated_json_name, child->GetMemberPointer());
+							}
+						}
 						break;
```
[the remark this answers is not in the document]
dimiden: The PR description was stale, fixed.

getroot (approved):
LGTM

PR #2305, opened 2026-08-23
feat(event): add the base modules for SEI message insertion

Keukhan:
## Summary

Writing an SEI message into an H.264 access unit had no reusable module. `TranscoderEvents` held
the only implementation and it was dead code: nothing called `PushEvent()` or `RunEvent()`, so an
SEI event from the SendEvent API was serialized onto a data track and dropped.

This adds the modules and removes the dead class. There is no caller yet — the consumer that
decides which picture to write into follows separately.

## Changes

- **`h264_sei_inserter`** (new) — writes `pic_timing` (type 1, SMPTE timecode) and
  `user_data_unregistered` (type 5). Rewrites an existing `pic_timing` rather than adding a second
  one (ITU-T H.264 7.4.1.2.3).
- **`h264_sps_rewriter`** (new) — turns `pic_struct_present_flag` on, without which a decoder reads
  no clock timestamps (D.2.2). One bit inside `vui_parameters()`, so the RBSP keeps its length.
- **`h264_sei`** — gains the `pic_timing` models and codec, loses the OME payload layout (specific
  to type 5, moved to the inserter).
- **`h264_parser`** — `num_units_in_tick`/`time_scale` are `u(32)`, not `ue(v)` (E.1.1); reading
  them as Exp-Golomb desynchronized the rest of the VUI, and also skewed the frame rate reported to
  the RTSP SDP. Adds `hrd_parameters()` parsing.
- **`nal_unit_insertor`** — adds `Replace()` and a target index; rejects an access unit with no
  parsed NAL unit; reads the source's own start code length.
- **`AVCDecoderConfigurationRecord::Equals()`** — compares the record structurally instead of the
  serialized bytes, which `Parse()` and `Serialize()` disagree on.
- **`H264SeiTimecodeZone`** (new, in `h264_sei.h`) — which clock a `pic_timing` timecode starts
  from: `UTC` (the default), `Local`, or a fixed offset (`+09:00`, `+0900`, `+09`, up to 14:00).
  IANA names are rejected: glibc has no `localtime_rz`, so resolving one would mean
  `setenv("TZ")` + `tzset()`, which races every other thread reading the clock.
- **`SEIEvent`** — moves to `base/modules/data_format/sei_event`, and now carries its fields as
  JSON instead of building an SEI NAL itself.
- **Removed** — `transcoder_events.{h,cpp}`.

## Behaviour changes

- A `pic_timing` timecode anchors on UTC. It used to read `localtime_r()` unconditionally, so the
  same stream carried a different time of day on every node it ran on. `<Timezone>Local</Timezone>`
  asks for the old behaviour.
- The SendEvent API rejects a request with no `seiType`; it used to default to
  `UserDataUnregistered`.
- `SEIEvent::Serialize()` puts JSON on the data track, not an SEI NAL. Nothing reads it today, but a
  node relaying that track over OVT to a different version would see the other format.

## Testing

46 unit test cases: the SEI message split/join, the pic_timing codec, the timecode generator, the
timezone and the SPS rewriter under `ome_test_modules`, and `SEIEvent` under `ome_test_base`. The
SPS fixture is a real parameter set taken off a live stream, chosen because it carries an
`emulation_prevention_three_byte`.

```bash
cmake -B build/Debug -G Ninja -DOME_BUILD_TESTS=ON
cmake --build build/Debug

# just what this PR adds
./build/Debug/bin/ome_test_modules \
    --gtest_filter='H264Sei*:H264TimecodeGenerator*:H264SpsRewriter*'
./build/Debug/bin/ome_test_base --gtest_filter='SeiEvent*'

# or through ctest, with the rest of those targets
cd build/Debug && ctest -L modules && ctest -L base
```

The inserter itself is not covered here: it needs a stream, a track and a packet, so its test lives
next to the consumer that drives it.


src/base/modules/data_format/sei_event/sei_event.cpp
```diff
@@ -0,0 +1,150 @@
...
+		logte("Could not parse SEI event (%zu bytes)", data->GetLength());
+		return nullptr;
+	}
+
+	auto sei_event = std::make_shared<SEIEvent>();
+
+	if (json.isMember("seiType") && json["seiType"].isString())
+	{
+		sei_event->_sei_type = json["seiType"].asString().c_str();
+	}
```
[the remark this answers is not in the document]
Keukhan: Fixed in 698c9dfa. `Deserialize()` now requires a non-empty `seiType` string. An unknown *name* is still let through on purpose: the consumer names it in the line it logs when it discards the event, which says more than dropping it here would. Covered by `SeiEvent.DeserializeRejectsAnythingItDidNotWrite` and `SeiEvent.DeserializeKeepsAnUnknownSeiType`.

src/modules/bitstream/h264/h264_sps_rewriter.cpp
```diff
@@ -0,0 +1,76 @@
...
+	H264SPS patched_sps;
+	if ((H264Parser::ParseSPS(patched->GetDataAs<uint8_t>(), patched->GetLength(), patched_sps) == false) ||
+		(patched_sps.IsPicStructPresent() == false))
+	{
+		logte("The patched SPS does not parse back with pic_struct_present_flag set");
+		return nullptr;
+	}
+
+	// An RBSP, not a NAL: NalUnitInsertor escapes what it is given
+	return NalUnitInsertor::RemoveEmulationPreventionBytes(patched);
```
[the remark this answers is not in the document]
Keukhan: Fixed in 698c9dfa. Both `H264SpsRewriter::EnablePicStructPresent()` and the emulation prevention helpers in `nal_unit_insertor.h` keep the NAL header, so the comments no longer call the result an RBSP.

src/modules/bitstream/h264/h264_decoder_configuration_record.cpp
```diff
@@ -234,24 +234,53 @@ bool AVCDecoderConfigurationRecord::Equals(const std::shared_ptr<DecoderConfigur
 		return false;
 	}
 
-	if (ProfileIndication() != other_config->ProfileIndication())
+	if ((Version() != other_config->Version()) ||
+		(ProfileIndication() != other_config->ProfileIndication()) ||
+		(Compatibility() != other_config->Compatibility()) ||
+		(LevelIndication() != other_config->LevelIndication()) ||
+		(LengthMinusOne() != other_config->LengthMinusOne()) ||
+		(ChromaFormat() != other_config->ChromaFormat()) ||
```
dimiden: `_chroma_format`, `_bit_depth_*_minus8` and `_num_of_sps_ext` are only ever set by `Parse()` - `AddSPS()`/`AddPPS()` leave them at 0 - so comparing them here makes a parsed record and a built record always unequal on High profile, spuriously bumping the track version. Either gate these three on `ProfileIndication()` being 100/110/122/144 on both sides, or populate them from the SPS in `AddSPS()`.

src/modules/bitstream/h264/h264_sei.cpp
```diff
@@ -65,100 +66,431 @@ std::shared_ptr<ov::Data> H264SEI::Serialize()
...
+	}
+
+	if (sps_context.pic_struct_present == false)
+	{
+		// Without pic_struct there is no clock timestamp at all.
+		return true;
 	}
 
-	BitReader payload_reader(payload_data);
+	if (reader.ReadBits(4, pic_timing.pic_struct) == false)
```
dimiden, line 235: Not introduced by this PR, but this new call site makes it reachable, so it would be good to fix it here: `BitReader::ReadBitsInternal` (`bit_reader.h:207`) bounds-checks `ceil(bits/8)` against the remaining bytes and ignores `_bit_offset`, so this 4-bit read at offset 6 with one byte left passes the check and then dereferences one byte past the payload. ASAN flags it on the `DelayOnlyPayloadMustBeReadWithTheEncodersContext` test in this PR.
Comparing against the existing `BitsRemained()` fixes it in one line.

src/modules/bitstream/h264/h264_parser.cpp
```diff
@@ -616,12 +622,14 @@ bool H264Parser::ParseVUI(NalUnitBitstreamParser &parser, H264SPS &sps)
 				uint32_t num_units_in_tick, time_scale;
 				uint8_t fixed_frame_rate_flag;
 
-				if (!parser.ReadUEV(num_units_in_tick))
+				// ITU-T H.264 E.1.1: both are u(32), not ue(v). Reading them as Exp-Golomb
+				// desynchronizes the bit position for everything that follows in the VUI.
+				if (!parser.ReadU32(num_units_in_tick))
```
dimiden, line 630: Same pre-existing `BitReader` bug as in `h264_sei.cpp`, and this PR newly exposes it here too: a 32-bit read at a non-byte-aligned position touches 5 bytes while `bit_reader.h:207` only requires 4.
`ParseSPS` runs on every ingest path, so this one is live - worth fixing together. `NalUnitBitstreamParser::NextPosition` (nal_unit_bitstream_parser.cpp:87) also dereferences `*_position` before its own `BytesRemained()` guard, and `*(_position + 1)` at :92 needs `>= 2`.

src/modules/bitstream/h264/h264_sei.cpp
```diff
@@ -65,100 +66,431 @@ std::shared_ptr<ov::Data> H264SEI::Serialize()
...
+
+	const auto *buffer	= rbsp->GetDataAs<uint8_t>();
+	const size_t length = rbsp->GetLength();
+	size_t position		= 0;
+
+	while (position < length)
+	{
+		// rbsp_trailing_bits() is a single 0x80 byte, because sei_payload() always leaves the
+		// bitstream byte aligned.
+		if (((length - position) == 1) && (buffer[position] == 0x80))
```
dimiden, line 106: The trailing-bits check only fires when exactly one byte is left, but Annex B allows `trailing_zero_8bits` after any NAL and `NalUnitFragmentHeader::Parse` folds those into the previous NAL's length - so on MPEG-TS/SRT ingest this either fabricates a `payloadType=128`/`buffering_period` message that `JoinMessages` re-emits, or returns false and makes `DecideSeiPlacement` insert a second pic_timing. Implement `more_rbsp_data()` semantics: trim trailing 0x00 first, then break on the final 0x80.

src/modules/bitstream/h264/h264_sei.h
```diff
@@ -23,11 +20,124 @@
...
+struct H264SeiTimecodeZone
+{
+	// "UTC", "Z" or an empty string, "Local", or a fixed offset written "+09:00", "-0500" or
+	// "+09". False when the text is none of those, and `zone` is then left as it was.
+	static bool Parse(const ov::String &text, H264SeiTimecodeZone &zone);
+
+	// Round trips through Parse()
+	ov::String ToString() const;
+
+	// The zone the server itself runs in, DST included
```
dimiden: `Generate` reads the anchor once and derives everything else from the PTS, so in `Local` mode a DST transition mid-stream is never picked up - the timecode stays an hour off the server's wall clock until a PTS discontinuity re-anchors (verified against America/New_York in both directions).
"DST included" reads as the opposite; suggest "the offset in effect when the anchor is taken - a DST change mid-stream is not followed". The `Local` branch is also the one branch this commit changed and the only one with no test.

src/modules/bitstream/h264/h264_sei_inserter.cpp
```diff
@@ -0,0 +1,910 @@
...
+
+	H264SeiPicTiming pic_timing;
+
+	if (placement.action == SeiPlacement::Action::Replace)
+	{
+		// The message has to be read the way the encoder wrote it. PatchSps() may have turned
+		// pic_struct_present_flag on since then, and reading a message that has no pic_struct as
+		// if it had one walks off the end of the payload.
+		auto encoder_context				 = sps_context;
+		encoder_context.pic_struct_present	 = (_patched_sps == false) && sps_context.pic_struct_present;
```
dimiden, line 743: `_patched_sps` is never cleared, so once any SPS on the track has been patched a later SPS that natively carries `pic_struct_present_flag=1` is still parsed without `pic_struct` - the encoder's value (3-8 for field pairing / frame doubling) is silently replaced by `DetectPictureStructure`'s 0/1/2. Track the current SPS instead: clear it in `DecideSeiPlacement` whenever an SPS parses with `IsPicStructPresent()` true.

src/modules/bitstream/h264/h264_sei.cpp
```diff
@@ -65,100 +66,431 @@ std::shared_ptr<ov::Data> H264SEI::Serialize()
...
+		uint8_t clock_timestamp_flag = 0;
+		if (reader.ReadBit(clock_timestamp_flag) == false)
+		{
+			return false;
+		}
+
+		if (clock_timestamp_flag == 0)
+		{
+			// This slot carries no timestamp. Keep the slot count aligned with NumClockTS by
+			// not appending, callers rely on clock_timestamps.size() < NumClockTS meaning absent.
```
dimiden: Skipping absent slots does not keep alignment - it destroys it: entries land in arrival order while `SerializePicTiming` treats the index as the slot number, so parsing `pic_struct=3` with slot 0 absent and slot 1 present and re-serializing moves the timecode from the bottom field to the top field. Add a per-slot `present` flag (or push a default-constructed entry for absent slots) and fix the comment, which also contradicts `h264_sei.h:67`.

src/modules/bitstream/h264/h264_sei.h
```diff
@@ -23,11 +20,124 @@
...
+	}
+};
+
+struct H264SeiTimecodeZone
+{
+	// "UTC", "Z" or an empty string, "Local", or a fixed offset written "+09:00", "-0500" or
+	// "+09". False when the text is none of those, and `zone` is then left as it was.
+	static bool Parse(const ov::String &text, H264SeiTimecodeZone &zone);
+
+	// Round trips through Parse()
```
dimiden: "Round trips through Parse()" does not hold for values `Parse()` did not produce, and `offset_seconds` is a plain public int32 that `SetTimezone()` accepts unchecked: 30 -> "+00:00", 54000 -> "+15:00" which `Parse` then rejects, 360000 -> "+100:00", and `-offset_seconds` at h264_sei.cpp:492 is signed-overflow UB for INT32_MIN. Either scope the comment to values Parse produced, or normalise in `ToString()` (widen to int64 for the negation and clamp to +/-14:00).

src/base/modules/data_format/sei_event/sei_event.h
```diff
@@ -0,0 +1,133 @@
...
+	static bool IsValid(const pugi::xml_node &xml);
+	static std::shared_ptr<SEIEvent> Parse(const pugi::xml_node &xml);
+
+	SEIEvent()	= default;
+	~SEIEvent() = default;
+
+	// {"seiType", "data", "keyframeOnly"} as JSON, carried on the event packet. EventConsumer is
+	// the only reader: it builds the real SEI message from these fields, so these bytes never
+	// reach an output.
+	std::shared_ptr<ov::Data> Serialize() const;
```
dimiden: There is no `EventConsumer` in the tree, and this comment's field list is now missing `timezone` (added to `Serialize()` in this commit). Reword it, and note that nothing reads what `Serialize()` writes today - `Deserialize` and `GetTimezone()` have no non-test callers.

src/modules/bitstream/h264/h264_sei_inserter.cpp
```diff
@@ -0,0 +1,910 @@
...
+		{
+			logtw("%s SEI insertion is only supported for H264. track(%u)",
+				  LogPrefix(stream).CStr(), packet->GetTrackId());
+		}
+
+		return false;
+	}
+
+	const auto format = packet->GetBitstreamFormat();
+	if ((format != cmn::BitstreamFormat::H264_ANNEXB) && (format != cmn::BitstreamFormat::H264_AVCC))
```
dimiden, line 270: *Not reachable in the current code* - nothing constructs `H264SeiInserter` yet, and MediaRouter converts AVCC to Annex B before anything downstream sees a packet - so this is a suggestion for whoever wires it up rather than a bug to fix now.

`IsInsertable` accepts `H264_AVCC`, but `NalUnitFragmentHeader::Parse` only scans Annex B start codes and `AppendNalPrefix` always writes a 4-byte length field regardless of `lengthSizeMinusOne`, so on real AVCC data the fragment boundaries would be garbage. Suggest narrowing this to `H264_ANNEXB` (or documenting in the header that the inserter must run after normalization) so the accept list does not promise support the code does not have.

src/base/modules/data_format/sei_event/sei_event_test.cpp
```diff
@@ -0,0 +1,258 @@
...
+	EXPECT_FALSE(parsed->GetTimezone().local) << "the timezone has to reach the consumer";
+	EXPECT_EQ(parsed->GetTimezone().offset_seconds, 9 * 3600);
+}
+
+// The default, and an event written by a build that had no timezone field at all
+TEST(SeiEvent, DeserializeFallsBackToUtc)
+{
+	auto source = std::make_shared<SEIEvent>();
+	source->SetSeiType("PictureTiming");
+	EXPECT_FALSE(SEIEvent::Deserialize(source->Serialize())->GetTimezone().local);
```
dimiden: Both this line and :212 chain `->GetTimezone()` off `Deserialize()` with no null check, unlike :215-216 in the same test - a `nullptr` regression segfaults the whole binary instead of failing one case.
Also, the test comment fits only the second case: the first one round-trips through `Serialize()`, which does emit `timezone`.

src/modules/bitstream/h264/h264_sei_inserter.cpp
Keukhan: Agreed, and thanks for saying so explicitly. Not reachable today: MediaRouter normalizes AVCC to Annex B before anything downstream sees a packet, and nothing constructs `H264SeiInserter` yet.

Narrowing the accept list to `H264_ANNEXB` goes in with the commit that wires the inserter up, so it lands next to the caller that would depend on it.

src/modules/bitstream/h264/h264_decoder_configuration_record.cpp
Keukhan: Fixed in 17858e78. Dropped the three fields rather than gating them on the profile: all three are coded in the SPS, which is already compared byte for byte. Covered by `AVCDecoderConfigurationRecord.AParsedHighProfileRecordEqualsOneBuiltFromTheSameSps`.

src/modules/bitstream/h264/h264_sei.cpp
Keukhan: Fixed in 91b2a289. `SplitMessages()` now trims trailing 0x00 back off when that uncovers the 0x80, so the padding neither fabricates a `payloadType=128` message nor fails the split. Covered by `H264Sei.SplitIgnoresTrailingZeroBytes` and `SplitKeepsPayloadZeroBytes`.

src/modules/bitstream/h264/h264_sei.cpp
Keukhan: Fixed in ddc58929. `ReadBitsInternal` now bounds the read with `BitsRemained()`, so the 4 bit read at offset 6 fails instead of reading a byte past the payload. The two predicates are identical whenever `_bit_offset` is 0 (`ceil(bits/8) <= B` iff `bits <= 8B`), so a byte aligned caller sees no change at all - across bits 1..64, offset 0..7 and 0..9 bytes left, the 224 verdicts that differ are all cases the old check accepted and the loop then dereferenced past the buffer. Covered by `BitReader.ReadPastTheEndOfAPartiallyConsumedByteFails`.

src/modules/bitstream/h264/h264_parser.cpp
Keukhan: Fixed in ddc58929, together with the `BitsRemained()` change. `NextPosition` now checks the bounds before it dereferences and needs 2 bytes for the `*(_position + 1)` lookahead. Covered by `NalUnitBitstreamParser.KeepsATrailingEmulationPreventionByte`; the full ctest run stays green across the AAC/VP8/AV1/HEVC parsers that share the reader.

src/modules/bitstream/h264/h264_sei_inserter.cpp
Keukhan: Fixed in ac3d8b79, but in `PatchSps()` rather than `DecideSeiPlacement()` - that one runs after the patch, so it only ever sees an already-flagged SPS; covered by `H264SeiInserter.ForgetsThePatchWhenAnSpsArrivesAlreadyFlagged`.

src/modules/bitstream/h264/h264_sei.cpp
Keukhan: Fixed in d6ee96ec - `H264SeiClockTimestamp` now carries `clock_timestamp_flag` as `present`, so an absent slot keeps its index instead of collapsing, and both comments say so.

src/modules/bitstream/h264/h264_sei.h
Keukhan: Fixed in 1d8faa50 - the comment now says the offset is the one in effect when the anchor is taken, and that a mid-stream DST change is not followed; the `Local` branch is covered now too.

src/modules/bitstream/h264/h264_sei.h
Keukhan: Fixed in 1d8faa50 - `ToString()` clamps to +/-14:00 and negates in int64, so it can no longer emit what `Parse()` rejects, and the comment scopes the exact round trip to the offsets `Parse()` produces.

src/base/modules/data_format/sei_event/sei_event.h
Keukhan: Fixed in 1d8faa50 - the comment no longer names `EventConsumer`, lists `timezone`, and says that nothing reads it today.

src/base/modules/data_format/sei_event/sei_event_test.cpp
Keukhan: Fixed in 1d8faa50 - all three cases now bind the result and `ASSERT_NE` it, and the comment covers all three rather than only the second.

src/modules/bitstream/h264/h264_sei.cpp
```diff
@@ -65,100 +66,431 @@ std::shared_ptr<ov::Data> H264SEI::Serialize()
...
+			continue;
```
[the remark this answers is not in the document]
Keukhan: Already fixed in d6ee96ec.

src/modules/bitstream/h264/h264_sei_inserter.cpp
```diff
@@ -0,0 +1,910 @@
...
+	{
+		if (should_warn() == true)
+		{
```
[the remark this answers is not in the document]
Keukhan: Already covered - narrowing goes in with the commit that wires the inserter up.

src/modules/bitstream/nalu/nal_unit_insertor.cpp
```diff
@@ -1,46 +1,129 @@
...
+	// 00 00 00 01 is a 4 byte start code, 00 00 01 a 3 byte one.
+	int32_t DetectStartCodeLength(const std::shared_ptr<const ov::Data> &data, const size_t first_nal_offset)
+	{
+		if ((first_nal_offset >= 4) && (data->GetDataAs<uint8_t>()[first_nal_offset - 4] == 0x00))
+		{
+			return 4;
+		}
+
+		return (first_nal_offset >= 3) ? 3 : 4;
+	}
```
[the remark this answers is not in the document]
Keukhan: Not reachable: `Parse` tests the 3 byte pattern first, so a 3 byte match never has a 0x00 in front.

src/modules/bitstream/nalu/nal_unit_insertor.cpp
```diff
@@ -139,36 +233,65 @@ std::shared_ptr<ov::Data> NalUnitInsertor::EmulationPreventionBytes(const std::s
...
+		if ((zero_count == 2) && (byte == 0x03))
+		{
+			zero_count = 0;
+			continue;
+		}
+
+		zero_count = (byte == 0x00) ? (zero_count + 1) : 0;
+
+		rbsp->Append(&byte, 1);
+	}
```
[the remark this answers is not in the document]
Keukhan: Not reachable: `EmulationPreventionBytes()` escapes a literal 0x03 too, so an escaped NAL never holds `00 00 03 04`.

src/modules/bitstream/h264/h264_sei_inserter.cpp
```diff
@@ -0,0 +1,919 @@
...
+			{
+				continue;
+			}
+
+			if (static_cast<H264NalUnitType>(buffer[offset] & kH264NalUnitTypeMask) != H264NalUnitType::Sps)
+			{
+				continue;
+			}
+
+			saw_sps = true;
```
dimiden, line 622: `saw_sps` is set before `EnablePicStructPresent`, so an SPS that fails to parse clears `_patched_sps` while `_sps_context.pic_struct_present` stays stale at true. High because of what ships: when (cpb+dpb delay length) % 8 == 1 the misparse succeeds, `pic_struct` reads 8 (frame tripling) from the padding, line 747 no longer corrects it, and a decoder that honours pic_struct triples the display duration for that whole GOP. This path was correct before ac3d8b7. Set `saw_sps` only after a successful `ParseSPS`.

src/modules/bitstream/h264/h264_decoder_configuration_record.cpp
```diff
@@ -234,24 +234,53 @@ bool AVCDecoderConfigurationRecord::Equals(const std::shared_ptr<DecoderConfigur
...
+		(LevelIndication() != other_config->LevelIndication()) ||
+		(LengthMinusOne() != other_config->LengthMinusOne()))
 	{
 		return false;
 	}
 
-	if (LevelIndication() != other_config->LevelIndication())
+	if ((NumOfSPS() != other_config->NumOfSPS()) ||
+		(NumOfPPS() != other_config->NumOfPPS()) ||
+		(NumOfSPSExt() != other_config->NumOfSPSExt()))
```
dimiden: `NumOfSPSExt()` and `Version()` have exactly the asymmetry this comment describes - only `Parse()` fills them, so a parsed and a built record never compare equal. High because the symptom is the one just fixed: a spurious track version bump, an init-segment swap and an HLS/LLHLS discontinuity for viewers on identical parameter sets. Rare in practice (needs an avcC carrying SPS extensions, or a non-conformant configurationVersion), but unlike the SEI modules this is live code on every keyframe. Drop both, or gate them the way you gated chroma_format.

src/modules/bitstream/h264/h264_sei_inserter.cpp
Keukhan: Fixed in bb3e7013 - `saw_sps` is now set only after a successful `ParseSPS`, so an SPS the parser cannot read leaves the patch state alone.

src/modules/bitstream/h264/h264_decoder_configuration_record.cpp
Keukhan: Fixed in a8f2128c - dropped `Version()` and `NumOfSPSExt()`, and `LengthMinusOne()`, which has the same asymmetry. Gating does not work for these: an SPS extension only exists in a High profile avcC, so the gate would keep the comparison exactly where the problem is.

dimiden (approved):
LGTM

PR #2308, opened 2026-08-26
build(deps): fetch x264 from the GitHub mirror instead of official repo

Keukhan:
## Summary

`code.videolan.org` has been unreliable, and a failed x264 download aborts the whole `InstallPrerequisites` run. **As a temporary workaround, x264 is fetched from the `github.com/mirror/x264` mirror repository** (the same commit) until the official host is dependable again.

- `X264_SOURCE_URL` now points at `github.com/mirror/x264/archive/${X264_SOURCE_REF}.tar.gz`.
  `ome_fetch` auto-detects the compression, so `.tar.gz` needs no other change.
- Built from `X264_SOURCE_REF` directly, so the tarball is always the commit
  pinned in `Versions.cmake` instead of the `master` snapshot the GitLab pattern
  resolved to by default.
- A `TODO` keeps the original URL so it can be restored once the mirror is no
  longer needed.

## Test

- `cmake -DOME_ENABLE_X264=ON -P cmake/InstallPrerequisites.cmake` on Ubuntu 22.04: x264 downloads, builds, and installs.


dimiden (approved):
LGTM

PR #2309, opened 2026-08-27
feat(api): add startOffset to sendEvent for scheduling events ahead of the media

getroot:
`sendEvent` places an event at the position the stream has reached when the request arrives, so an event can only be marked at the moment of the call. A sender that knows the moment in advance, or that cannot call at that exact instant, has no way to express it.

This adds `startOffset`, an optional millisecond value that delays the event by that much from that position. It is resolved once for the request, before the event format is handled, so it applies to every format the API accepts. Negative values are refused and the maximum is 300000. A request that arrives before the stream has sent any media now answers 409 rather than 500, because there is no position to place the event at.

**Test**

Publish an RTMP stream and play LL-HLS.

1. Send an event with `startOffset` omitted and another with it set, back to back. Download the segment that carries them and read the `emsg` boxes.
2. Send `-1`, `300001`, `300000`, `"3000"`, and `1.5`.
3. Send an event to a stream that has not started publishing.

Pass conditions: the two `emsg` presentation times differ by exactly the offset, measured against each request's own base position, and the event with no offset lands where it did before. `-1`, `300001`, `"3000"`, and `1.5` are refused with 400, and `300000` is accepted. The stream that has not started answers 409.

Measured on this branch: presentation times 18.571s and 21.580s against base positions 18571 and 18580, an exact 3000 ms gap. Unit tests 840 pass.

### DIFF
--- a/src/api_server/controllers/v1/vhosts/apps/streams/stream_actions_controller.cpp
+++ b/src/api_server/controllers/v1/vhosts/apps/streams/stream_actions_controller.cpp
@@ -26,6 +26,10 @@ namespace api
 {
 	namespace v1
 	{
+		// An event scheduled ahead of the media holds the stream's event channel
+		// until it is reached, so the reservation is bounded
+		static constexpr int64_t kMaxEventStartOffsetMs = 300000;
+
 		void StreamActionsController::PrepareHandlers()
 		{
 			RegisterPost(R"((hlsDumps))", &StreamActionsController::OnPostHLSDumps);
@@ -283,6 +287,7 @@ namespace api
 			//   "eventFormat": "id3v2",
 			//   "eventType": "video",
 			//	 "urgent": false,
+			//	 "startOffset": 3000,
 			//   "events":[
 			//       {
 			//         "frameType": "TXXX",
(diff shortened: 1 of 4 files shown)


src/api_server/controllers/v1/vhosts/apps/streams/stream_actions_controller.cpp
```diff
@@ -329,10 +334,42 @@ namespace api
...
 
+			// startOffset (Optional, milliseconds) - schedules the event that far
+			// ahead of the position the media has reached now
+			int64_t start_offset_ms = 0;
+			if (request_body.isMember("startOffset") == true)
+			{
+				if (request_body["startOffset"].isInt64() == false)
+				{
+					throw http::HttpError(http::StatusCode::BadRequest, "startOffset must be an integer in milliseconds");
+				}
```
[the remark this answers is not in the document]
getroot: NP

src/api_server/controllers/v1/vhosts/apps/streams/stream_actions_controller.cpp
```diff
@@ -329,10 +334,42 @@ namespace api
...
+			// Every format shares one position, so an event carries the same time
+			// wherever it ends up
+			int64_t timestamp = source_stream->GetCurrentTimestampMs();
+			if (timestamp < 0)
+			{
+				throw http::HttpError(http::StatusCode::Conflict,
+									  "Media has not started yet: [%s/%s/%s]",
+									  vhost->GetName().CStr(), app->GetVHostAppName().GetAppName().CStr(), stream->GetName().CStr());
+			}
+			timestamp += start_offset_ms;
```
Keukhan, line 368: The offset is added after `GetCurrentTimestampMs()`, which is the part that keeps successive results increasing. So a request with `startOffset` followed by one without it sends a timestamp lower than the previous event's.

Ignore this if that ordering is intended.

src/api_server/controllers/v1/vhosts/apps/streams/stream_actions_controller.cpp
```diff
@@ -329,10 +334,42 @@ namespace api
...
+
+				if (start_offset_ms > kMaxEventStartOffsetMs)
+				{
+					throw http::HttpError(http::StatusCode::BadRequest, "startOffset must not exceed %" PRId64 " ms", kMaxEventStartOffsetMs);
+				}
+			}
+
+			// Every format shares one position, so an event carries the same time
+			// wherever it ends up
+			int64_t timestamp = source_stream->GetCurrentTimestampMs();
```
Keukhan, line 361: This sits above the `eventFormat` check, so an unsupported `eventFormat` on a stream that has not started now answers 409 instead of the 400 it used to. Moving the block below the format/events validation keeps that 400.

Keukhan (approved):
I’ve left two simple review comments. Overall, though, I don’t see any critical issues, so I’ll approve it in advance.

src/api_server/controllers/v1/vhosts/apps/streams/stream_actions_controller.cpp
getroot: Good catch. Fixed by claiming the final position on the stream and refusing
anything at or before the last claim. Verified with a 409 on the reversed case.

src/api_server/controllers/v1/vhosts/apps/streams/stream_actions_controller.cpp
getroot: The 409 is true here too: the request fails either way while the stream is not publishing. Leaving it as is.

src/api_server/controllers/v1/vhosts/apps/streams/stream_actions_controller.cpp
```diff
@@ -385,6 +422,16 @@ namespace api
...
+			// request is known to be sound, so a refused request leaves no mark
+			if (source_stream->ClaimEventTimestampMs(timestamp) == false)
+			{
+				throw http::HttpError(http::StatusCode::Conflict,
+									  "An event is already placed at or after this position, so events must be sent in order: [%s/%s/%s]",
+									  vhost->GetName().CStr(), app->GetVHostAppName().GetAppName().CStr(), stream->GetName().CStr());
+			}
+
 			if (source_stream->SendDataFrame(timestamp, event_format, event_type, events_data, urgent, true) == false)
 			{
```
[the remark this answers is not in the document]
getroot: NP

PR #2310, opened 2026-08-28
refactor(transcoder): rework codec module selection

Keukhan:
## Summary

Refactors transcoder codec/module candidate selection to consult the registered module table (instead of only checking device existence), aiming to skip unsupported hardware codec paths earlier and reduce late failures during encoder/decoder creation.

## Changes

- Updates TranscodeModules::GetModule() filtering and uses module-table lookups when building encoder/decoder hardware candidates.
- Extends module registrations (MP3 decode under DEFAULT/FFmpeg audio; Whisper under DEFAULT) and centralizes codec-list formatting via cmn::GetCodecIdListString().
- Removes TranscodeGPU::IsSupported() helpers and clamps NVML-reported device count to MAX_DEVICE_COUNT.

## Testing

`ome_test_transcoder --gtest_filter='TranscodeModulesTest.*'` - 5 cases over `TranscodeModules::GetModule()`:

- Rejects a codec the module never registered, where media type, module, device and direction all match - the case the old lookup accepted
- Rejects a codec registered for the other direction
- Finds a registered codec, the codecs registered for the DEFAULT module, and rejects an unregistered device id

The two codec-only expectations fail against the old lookup.

### DIFF
--- a/src/transcoder/transcoder_decoder.cpp
+++ b/src/transcoder/transcoder_decoder.cpp
@@ -93,10 +93,22 @@ std::shared_ptr<std::vector<std::shared_ptr<info::CodecCandidate>>> TranscodeDec
 		{
 			for (int device_id = 0; device_id < TranscodeGPU::GetInstance()->GetDeviceCount(module_id); device_id++)
 			{
-				if ((gpu_id == ALL_GPU_ID || gpu_id == device_id) && TranscodeGPU::GetInstance()->IsSupported(module_id, device_id) == true)
+				if (gpu_id != ALL_GPU_ID && gpu_id != device_id)
 				{
-					candidate_modules->push_back(std::make_shared<info::CodecCandidate>(track->GetCodecId(), module_id, device_id));
+					continue;
 				}
+
+				if (tc::TranscodeModules::GetInstance()->GetModule(/*coder_type=*/false, track->GetCodecId(), module_id, device_id) == nullptr)
+				{
+					logtd("%s:%d cannot decode %s. Skip this module",
+						  cmn::GetCodecModuleIdString(module_id),
+						  device_id,
+						  cmn::GetCodecIdString(track->GetCodecId()));
+
+					continue;
+				}
+
+				candidate_modules->push_back(std::make_shared<info::CodecCandidate>(track->GetCodecId(), module_id, device_id));
 			}
 		}
 
--- a/src/transcoder/transcoder_encoder.cpp
+++ b/src/transcoder/transcoder_encoder.cpp
@@ -114,10 +114,22 @@ std::shared_ptr<std::vector<std::shared_ptr<info::CodecCandidate>>> TranscodeEnc
 		{
 			for (int device_id = 0; device_id < TranscodeGPU::GetInstance()->GetDeviceCount(module_id); device_id++)
 			{
-				if ((gpu_id == ALL_GPU_ID || gpu_id == device_id) && TranscodeGPU::GetInstance()->IsSupported(module_id, device_id) == true)
+				if (gpu_id != ALL_GPU_ID && gpu_id != device_id)
 				{
-					candidate_modules->push_back(std::make_shared<info::CodecCandidate>(track->GetCodecId(), module_id, device_id));
+					continue;
 				}
+
+				if (tc::TranscodeModules::GetInstance()->GetModule(/*coder_type=*/true, track->GetCodecId(), module_id, device_id) == nullptr)
+				{
+					logtd("%s:%d cannot encode %s. Skip this module",
+						  cmn::GetCodecModuleIdString(module_id),
+						  device_id,
+						  cmn::GetCodecIdString(track->GetCodecId()));
+
+					continue;
+				}
+
+				candidate_modules->push_back(std::make_shared<info::CodecCandidate>(track->GetCodecId(), module_id, device_id));
 			}
 		}
 		else
--- a/src/transcoder/transcoder_modules.h
+++ b/src/transcoder/transcoder_modules.h
@@ -36,11 +36,16 @@ namespace tc
 
 			for (const auto &codec : _modules)
 			{
-				if (codec->GetMediaType() == media_type &&
-					codec->GetModuleId() == module_id &&
-					codec->GetDeviceId() == device_id &&
-					(coder_type ? codec->isEncoder() : codec->isDecoder()) &&
-					std::find(codec->GetSupportedCodecs().begin(), codec->GetSupportedCodecs().end(), codec_id) != codec->GetSupportedCodecs().end())
+				if (codec->GetMediaType() != media_type ||
+					codec->GetModuleId() != module_id ||
+					codec->GetDeviceId() != device_id ||
+					(coder_type ? codec->isEncoder() : codec->isDecoder()) == false)
+				{
+					continue;
+				}
+
+				const auto &supported_codecs = codec->GetSupportedCodecs();
+				if (std::find(supported_codecs.begin(), supported_codecs.end(), codec_id) != supported_codecs.end())
 				{
 					return codec;
 				}
--- a/src/transcoder/transcoder_test.cpp
+++ b/src/transcoder/transcoder_test.cpp
@@ -2,16 +2,58 @@
 //
 //  OvenMediaEngine - Unit Tests
 //
-//  tests/transcoder/test_transcoder_placeholder.cpp
-//  Covers: TranscoderContext creation, GPU detection, codec mapping
+//  Covers: codec module table lookups
 //
-//  NOTE: Skeleton. Transcoder tests may require FFmpeg initialization.
-//        Tests that do NOT require actual codec instances can be added here.
+//  NOTE: Transcoder tests may require FFmpeg initialization. Tests that do NOT
+//        require actual codec instances can be added here.
 //
 //==============================================================================
 #include <gtest/gtest.h>
 
-TEST(TranscoderPlaceholder, TodoImplementTests)
+#include "transcoder_modules.h"
+
+// The registrations below are the software ones, which are always present; the
+// hardware modules need a device and are not part of these expectations.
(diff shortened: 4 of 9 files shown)


src/transcoder/transcoder_encoder.cpp
```diff
@@ -114,10 +114,22 @@ std::shared_ptr<std::vector<std::shared_ptr<info::CodecCandidate>>> TranscodeEnc
...
 			{
-				if ((gpu_id == ALL_GPU_ID || gpu_id == device_id) && TranscodeGPU::GetInstance()->IsSupported(module_id, device_id) == true)
+				if (gpu_id != ALL_GPU_ID && gpu_id != device_id)
 				{
-					candidate_modules->push_back(std::make_shared<info::CodecCandidate>(track->GetCodecId(), module_id, device_id));
+					continue;
 				}
+
+				if (tc::TranscodeModules::GetInstance()->GetModule(/*coder_type=*/true, track->GetCodecId(), module_id, device_id) == nullptr)
+				{
```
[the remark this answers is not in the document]
Keukhan: This tree has no hardware codec implementations - the encoder and decoder factories have no NVENC case - so a hardware candidate cannot be created here regardless of the device slot. An uninitialized slot now falls through to the next candidate instead of being filtered one step earlier, which ends at the same place. Not changing this.

src/transcoder/transcoder_decoder.cpp
```diff
@@ -93,10 +93,22 @@ std::shared_ptr<std::vector<std::shared_ptr<info::CodecCandidate>>> TranscodeDec
...
 			{
-				if ((gpu_id == ALL_GPU_ID || gpu_id == device_id) && TranscodeGPU::GetInstance()->IsSupported(module_id, device_id) == true)
+				if (gpu_id != ALL_GPU_ID && gpu_id != device_id)
 				{
-					candidate_modules->push_back(std::make_shared<info::CodecCandidate>(track->GetCodecId(), module_id, device_id));
+					continue;
 				}
+
+				if (tc::TranscodeModules::GetInstance()->GetModule(/*coder_type=*/false, track->GetCodecId(), module_id, device_id) == nullptr)
+				{
```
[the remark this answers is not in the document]
Keukhan: This tree has no hardware codec implementations - the encoder and decoder factories have no NVENC case - so a hardware candidate cannot be created here regardless of the device slot. An uninitialized slot now falls through to the next candidate instead of being filtered one step earlier, which ends at the same place. Not changing this.

getroot (approved):

src/transcoder/transcoder_encoder.cpp
```diff
@@ -114,10 +114,22 @@ std::shared_ptr<std::vector<std::shared_ptr<info::CodecCandidate>>> TranscodeEnc
...
 			for (int device_id = 0; device_id < TranscodeGPU::GetInstance()->GetDeviceCount(module_id); device_id++)
 			{
-				if ((gpu_id == ALL_GPU_ID || gpu_id == device_id) && TranscodeGPU::GetInstance()->IsSupported(module_id, device_id) == true)
+				if (gpu_id != ALL_GPU_ID && gpu_id != device_id)
 				{
-					candidate_modules->push_back(std::make_shared<info::CodecCandidate>(track->GetCodecId(), module_id, device_id));
+					continue;
 				}
+
+				if (tc::TranscodeModules::GetInstance()->GetModule(/*coder_type=*/true, track->GetCodecId(), module_id, device_id) == nullptr)
```
getroot, line 122: `GetModule()` cannot answer the question `IsSupported()` answered. It reports whether a codec was registered for a device slot, not whether that slot actually came up.

The removed `IsSupportedNV()` carried the reason:

```cpp
// Supported iff this slot was successfully initialized. Indices are sparse
// (by NVML device id), so a count comparison is not enough.
return gpu_id >= 0 && gpu_id < MAX_DEVICE_COUNT && _device_context_nv[gpu_id] != nullptr;
```

`GetDeviceIdNV()` still states the same fact at `transcoder_gpu.cpp:458`. The per-device arrays are indexed by NVML device id, so they are sparse, and `_device_count_nv` is a success counter rather than an index bound.

Meanwhile the registration loop in `transcoder_modules.cpp` registers `for i in 0 .. GetDeviceCount() - 1` unconditionally, so a slot that failed to initialize still gets a table entry.

Concrete case, `CUDA_VISIBLE_DEVICES=1` on a two-GPU host. NVML enumerates both devices, CUDA exposes one, so NVML device 0 finds no PCI match and is skipped. `_device_context_nv[0]` stays null while `_device_count_nv` becomes 1.

* Before: the loop visits `device_id` 0, `IsSupportedNV(0)` sees the null context, no candidate is added, and the track falls back to the next module.
* After: the loop visits `device_id` 0, `GetModule(..., 0)` finds the entry registered for slot 0 and adds the candidate. Encoder creation then calls `GetDeviceContextNV(0)`, gets null, and fails.

A clean fallback becomes a hard encoder creation failure. The decoder path has the same gap.

`_supported_devices` already holds exactly the `(module_id, device_id)` pairs that initialized successfully. Gating on that, or skipping dead slots at registration time so the table only ever describes usable devices, would both close this.

src/transcoder/transcoder_decoder.cpp
```diff
@@ -93,10 +93,22 @@ std::shared_ptr<std::vector<std::shared_ptr<info::CodecCandidate>>> TranscodeDec
...
 			for (int device_id = 0; device_id < TranscodeGPU::GetInstance()->GetDeviceCount(module_id); device_id++)
 			{
-				if ((gpu_id == ALL_GPU_ID || gpu_id == device_id) && TranscodeGPU::GetInstance()->IsSupported(module_id, device_id) == true)
+				if (gpu_id != ALL_GPU_ID && gpu_id != device_id)
 				{
-					candidate_modules->push_back(std::make_shared<info::CodecCandidate>(track->GetCodecId(), module_id, device_id));
+					continue;
 				}
+
+				if (tc::TranscodeModules::GetInstance()->GetModule(/*coder_type=*/false, track->GetCodecId(), module_id, device_id) == nullptr)
```
getroot, line 101: Same liveness gap as the encoder path, see the comment on `transcoder_encoder.cpp`.

src/transcoder/transcoder_modules.h
```diff
@@ -36,11 +36,16 @@ namespace tc
...
-					std::find(codec->GetSupportedCodecs().begin(), codec->GetSupportedCodecs().end(), codec_id) != codec->GetSupportedCodecs().end())
+				if (codec->GetMediaType() != media_type ||
+					codec->GetModuleId() != module_id ||
+					codec->GetDeviceId() != device_id ||
+					(coder_type ? codec->isEncoder() : codec->isDecoder()) == false)
+				{
+					continue;
+				}
+
+				const auto &supported_codecs = codec->GetSupportedCodecs();
```
getroot, line 47: Worth calling out: binding once here fixes undefined behavior in the old code.

The previous form was `std::find(codec->GetSupportedCodecs().begin(), codec->GetSupportedCodecs().end(), codec_id)`. `GetSupportedCodecs()` returns by value, so `begin()` and `end()` came from two separate temporary vectors and never formed a valid range.

While you are here, `GetSupportedCodecs()` still returns `std::vector` by value (`base/info/codec.h:118`), so this line copies the vector on every iteration of the scan. Returning `const std::vector<cmn::MediaCodecId> &` would drop the copy and remove the trap for the next caller.

src/transcoder/transcoder_test.cpp
```diff
@@ -2,16 +2,58 @@
...
 //
 //==============================================================================
 #include <gtest/gtest.h>
 
-TEST(TranscoderPlaceholder, TodoImplementTests)
+#include "transcoder_modules.h"
+
+// The registrations below are the software ones, which are always present; the
+// hardware modules need a device and are not part of these expectations.
+class TranscodeModulesTest : public ::testing::Test
```
getroot, line 17: These cover the software registrations only, but the gate this PR actually changed is the hardware path in the encoder and decoder candidate builders. A case pinning the behavior for a device slot that enumerated yet failed to initialize would have caught the issue noted on `transcoder_encoder.cpp`.

getroot (changes requested):
Thanks for the cleanup, the `GetModule()` rewrite fixes a real undefined-behavior bug on its own (noted inline).

One change needs a second look before merge. Replacing `TranscodeGPU::IsSupported()` with a module-table lookup drops the device liveness check, and the module table carries no liveness information to replace it with. Details on `transcoder_encoder.cpp`.

src/transcoder/transcoder_encoder.cpp
Keukhan: Pre-existing: registration adds 0..GetDeviceCount()-1 regardless of which slots came up, so the table already describes dead slots and misses live sparse ones. The fix belongs there - _supported_devices holds exactly the pairs that came up - not in another liveness check next to the lookup. Out of scope for this refactor, tracking it separately.

src/transcoder/transcoder_decoder.cpp
Keukhan: Pre-existing: registration adds 0..GetDeviceCount()-1 regardless of which slots came up, so the table already describes dead slots and misses live sparse ones. The fix belongs there - _supported_devices holds exactly the pairs that came up - not in another liveness check next to the lookup. Out of scope for this refactor, tracking it separately.

src/transcoder/transcoder_modules.h
Keukhan: Done - GetSupportedCodecs() now returns a const reference (c3391186).

src/transcoder/transcoder_test.cpp
Keukhan: The hardware codec implementations live in the enterprise tree, so this tree cannot construct a hardware encoder or decoder at all. The hardware path is not reachable here to pin.

getroot (approved):

PR #2311, opened 2026-08-28
fix(llhls): anchor synced segment boundaries on the stream's first segment

getroot:
In synced segmentation the reference track aims each cut at positions counted from zero. A stream whose timeline does not start there has its frames at their own offset, so every aimed position falls between two of them and the cut lands on the following frame instead. The stream then reports every segment as closing past its plan, for as long as it runs. `TimestampMode` `SystemClock` puts every stream in that state, and any source that opens away from zero reaches it as well.

The first segment start now fixes where those positions begin, so each one is a frame the stream actually has and the cut lands on it. The spacing between them is unchanged, so a segment cut short by a marker still returns to the cadence at the next one.

**Test**

Publish a stream whose timeline does not start at zero, with synced segmentation, and play LL-HLS.

1. Watch the log for `Segment closed at ... past its ... plan`.
2. Read EXTINF and the part durations from the chunklist.

Pass conditions: the warning does not appear, EXTINF equals the configured segment duration, and the parts equal the configured chunk duration.

Measured on this branch with a scheduled channel on `SystemClock`: ten minutes and three item switches with no such warning, EXTINF 2.000 on every segment, and parts of 1.000. The same build before the change logged the warning on every segment.

Unit tests: 840 pass, including the boundary tests that cover a segment cut short returning to the cadence.

### DIFF
--- a/src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
+++ b/src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
@@ -160,18 +160,23 @@ namespace bmff
 		// (within half a frame) belongs to that boundary, not before it
 		const int64_t tolerance_us = (_video_frame_rate > 0) ? std::llround(500000.0 / _video_frame_rate) : 500;
 
-		int64_t multiple = std::max(static_cast<int64_t>(1), position_us / _segment_duration_us);
+		const int64_t origin_us = (_grid_origin_us >= 0) ? _grid_origin_us : 0;
+
+		// The origin itself is a boundary, so the search starts at it; the check
+		// below rejects it when the segment already starts there
+		int64_t multiple = std::max(static_cast<int64_t>(0), (position_us - origin_us) / _segment_duration_us);
 
 		while (true)
 		{
-			int64_t boundary_us = multiple * _segment_duration_us;
+			int64_t boundary_us = origin_us + (multiple * _segment_duration_us);
 
 			// The aimed position must be a real frame position, so the realized
-			// cut can land exactly on it when the keyframe cadence allows
+			// cut can land exactly on it when the keyframe cadence allows. The
+			// cadence is counted from the origin, where this stream's frames are
 			if (_video_frame_rate > 0)
 			{
-				double frame_index = std::round(static_cast<double>(boundary_us) * _video_frame_rate / 1000000.0);
-				boundary_us = std::llround(frame_index * 1000000.0 / _video_frame_rate);
+				double frame_index = std::round(static_cast<double>(boundary_us - origin_us) * _video_frame_rate / 1000000.0);
+				boundary_us = origin_us + std::llround(frame_index * 1000000.0 / _video_frame_rate);
 			}
 
 			if (boundary_us > position_us + tolerance_us)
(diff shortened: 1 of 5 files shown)


src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
```diff
@@ -194,6 +200,13 @@ namespace bmff
 			return plan;
 		}
 
+		// The first start this reference sees fixes where the aimed positions
+		// begin for the stream
+		if (_grid_origin_us < 0)
+		{
+			_grid_origin_us = *segment_start_us;
+		}
```
dimiden: A negative first start never latches here: the stored value still satisfies `_grid_origin_us < 0`, so the grid silently stays zero-anchored until the first non-negative start (then shifts once), and a start below `-tolerance_us` plans `end_us = -1`, colliding with the unknown-start sentinel. Negative first dts is reachable (per-stream rebase latch in `AdjustTimestampByBase`, B-frame leading dts) - latch the first value unconditionally or make the origin `std::optional`.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.h
```diff
@@ -121,6 +121,13 @@ namespace bmff
 		// Whether the opening anchor went out with the first stored chunk; the
 		// reference's media thread is the only writer
 		bool _opening_anchor_published = false;
+
+		// Where the aimed positions start counting. The frames of a stream sit
+		// at their own offset, and counting from zero would aim between two of
+		// them on every segment; taking the first segment start as the origin
+		// puts every aimed position on a frame this stream actually has.
+		// Negative until the first start is known, which counts from zero.
+		int64_t _grid_origin_us = -1;
```
dimiden: This new comment is correct, but three older comments still describe the zero-anchored grid it replaces:
- class doc (`.h:83`): "multiples of the configured duration snapped to its own frame cadence"
- ctor doc (`.h:92`): "(0 keeps plain multiples)"
- `llhls_stream.cpp:1906`: "The reference aims at multiples of the segment duration on its own frame cadence"

Sweep them to the origin-anchored wording.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
dimiden, line 186: Pre-existing nit: the doc on `NextBoundaryUs` (`h:109`) says "The first aimed position at or after the given one", but the loop only returns boundaries strictly past `position_us + tolerance_us` - worth fixing the doc while this function is being touched.

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
getroot: Fixed

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.h
getroot: Applied

src/modules/containers/bmff/fmp4_packager/synced_boundary_policy.cpp
getroot: Applied

dimiden (approved):
LGTM

PR #2316, opened 2026-08-31
perf(scheduled): close the item's format context off the releasing thread

getroot:
A scheduled channel keeps an ffmpeg format context open for every file item in its schedule. When a schedule is replaced, the files of the old schedule are closed on whichever thread releases them last, and that is the playback thread, at the exact moment it starts the next program. On a storage that responds slowly, such as a network mount, closing a single file can take hundreds of milliseconds, so the program starts late by the total of those closes. And when the start is more than one second late, the channel also seeks the first item forward by the lost time to stay on schedule; that seek lands on the next keyframe, so the item can reach the end of its file before its configured duration and end early.

The files are now closed on a dedicated worker of the task pool named `SchedCtxClose`. The thread that releases a file only hands it over and moves on, so a slow storage no longer delays the program start, and the late-start seek is not triggered in the first place. The dedicated worker runs its closes one at a time without occupying the pool's shared workers, so however long a close blocks, work of other modules such as the DRM key prefetch is unaffected. The worker lives only while it has work: it exits once its queue runs dry and is created again by the next close.

**Test**

Play a scheduled channel with a fallback and a scheduled program, update the schedule file while the fallback is on air, and let the program start.

1. Read the log around the program start: `Start ... program` and `GetFirstItem`.
2. Find the `Closing format context` lines and read the thread name in front of them.
3. Let the program run through its items and switch back to the fallback.

Pass conditions: the program starts at its scheduled time with `GetFirstItem` reporting an elapsed time under a second, the `Closing format context` lines carry the `SchedCtxClose` thread name instead of the `Scheduled` thread, and the items play and switch as before.

Measured on this branch with a one second delay added to each close: across 19 schedule replacements the program start was at most 17 ms late, all 81 closes ran on the `SchedCtxClose` worker one at a time, the worker thread disappeared between bursts and returned for the next one, and the process thread count stayed constant throughout. The same build with the closes kept synchronous started 4,009 ms late.

### DIFF
--- a/src/providers/scheduled/schedule.cpp
+++ b/src/providers/scheduled/schedule.cpp
@@ -65,15 +65,15 @@ namespace pvd
 				return;
 			}
 
-			// Closing can block on the storage the file lives on, so it must not
-			// run on whichever thread happens to drop the last reference (the
-			// playback thread swapping schedules, or the API thread updating one)
+			// Closing can block on the storage the file lives on, so it goes to a dedicated
+			// worker instead of the releasing thread (normally the playback thread) or the
+			// shared workers. When the pool cannot take it, the close runs right here.
 			auto close = [ctx, file_path_copy]() mutable {
 				logti("LoadContext: Closing format context : %s", file_path_copy.CStr());
 				::avformat_close_input(&ctx);
 			};
 
-			if (ov::TaskPool::GetInstance()->Post(close) == false)
+			if (ov::TaskPool::GetInstance()->PostDedicated("SchedCtxClose", close) == false)
 			{
 				close();
 			}
--- a/src/providers/scheduled/schedule.cpp
+++ b/src/providers/scheduled/schedule.cpp
@@ -10,6 +10,7 @@
 #include "schedule.h"
 #include "schedule_private.h"
 #include <base/ovlibrary/files.h>
+#include <modules/task_pool/task_pool.h>
 
 namespace pvd
 {
@@ -59,10 +60,22 @@ namespace pvd
 
 		ov::String file_path_copy = _file_path;
 		_format_context.reset(ctx, [file_path_copy](AVFormatContext *ctx) {
-			if (ctx)
+			if (ctx == nullptr)
 			{
+				return;
+			}
+
+			// Closing can block on the storage the file lives on, so it must not
+			// run on whichever thread happens to drop the last reference (the
+			// playback thread swapping schedules, or the API thread updating one)
+			auto close = [ctx, file_path_copy]() mutable {
 				logti("LoadContext: Closing format context : %s", file_path_copy.CStr());
 				::avformat_close_input(&ctx);
+			};
+
+			if (ov::TaskPool::GetInstance()->Post(close) == false)
+			{
+				close();
 			}
 		});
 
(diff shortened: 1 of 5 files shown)


src/providers/scheduled/schedule.cpp
```diff
@@ -59,10 +60,22 @@ namespace pvd
...
+
+			// Closing can block on the storage the file lives on, so it must not
+			// run on whichever thread happens to drop the last reference (the
+			// playback thread swapping schedules, or the API thread updating one)
+			auto close = [ctx, file_path_copy]() mutable {
 				logti("LoadContext: Closing format context : %s", file_path_copy.CStr());
 				::avformat_close_input(&ctx);
+			};
+
+			if (ov::TaskPool::GetInstance()->Post(close) == false)
```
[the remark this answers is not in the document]
getroot: NP

src/providers/scheduled/schedule.cpp
```diff
@@ -59,10 +60,22 @@ namespace pvd
...
+			{
+				close();
 			}
```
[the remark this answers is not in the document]
getroot: NP

src/providers/scheduled/schedule.cpp
dimiden: This shared pool is also the one the LL-HLS DRM key prefetch runs on.
So, if the storage hangs, these closes park all 4 workers and a queued prefetch behind them never runs (`_key_request_in_flight` stays set, so no retry either) - key rotation is then deferred until the mount recovers. Consider a dedicated pool instance for the closes, the option `task_pool.h` itself suggests for work that shouldn't sit behind other modules.

src/providers/scheduled/schedule.cpp
```diff
@@ -59,10 +60,22 @@ namespace pvd
...
 		_format_context.reset(ctx, [file_path_copy](AVFormatContext *ctx) {
-			if (ctx)
+			if (ctx == nullptr)
 			{
+				return;
+			}
+
+			// Closing can block on the storage the file lives on, so it must not
+			// run on whichever thread happens to drop the last reference (the
+			// playback thread swapping schedules, or the API thread updating one)
```
dimiden: The fallback below runs the close on exactly this thread when `Post` fails - "must not" overstates it; "should not normally" would match the code.

dimiden (approved):
LGTM, the `schedule.cpp:76` comment is non-blocking.

To be clear, this is a very unlikely edge case in practice (it needs a hard-hung mount and DRM key rotation on the same server), so take it as a non-blocking suggestion rather than a change request - worth keeping in mind as `TaskPool` gains more users, but I'm fine merging as is.

src/providers/scheduled/schedule.cpp
getroot: Reworked: the closes now go to a dedicated worker of the pool (PostDedicated("SchedCtxClose", ...)), a single thread of their own, so however long they block, the shared workers stay free for the DRM key prefetch. The worker lives only while it has work and is recreated by the next close.

src/providers/scheduled/schedule.cpp
getroot: Right, reworded.

PR #2318, opened 2026-09-01
refactor(mediarouter): record per-track readiness on shared track stats

getroot:
Track readiness (valid and quality-measured) is evaluated in multiple places, and the result is visible only inside the mediarouter's own copy of the stream info because parsed track properties are published into that copy only.

IsStreamReady() now evaluates every track without early return and records per-track readiness on the shared TrackStats, and CheckUnpreparedTrackTimeout() reuses the recorded fact instead of duplicating the judgment. Stream preparation behavior is unchanged.

## Test
- Full unit test suite passed (846/846).
- WHIP ingest with VP8+OPUS and H264+OPUS: both streams prepared normally within about 1 second, no unprepared-track warnings for healthy input, and WebRTC/LLHLS playback was unaffected.

### DIFF
--- a/src/mediarouter/mediarouter_stream.cpp
+++ b/src/mediarouter/mediarouter_stream.cpp
@@ -131,22 +131,35 @@ bool MediaRouteStream::IsStreamReady()
 
 	auto tracks = _stream->GetTracks();
 
+	bool all_tracks_ready = true;
+
+	// Evaluate every track (no early return) so each ready track is recorded even
+	// while another track still blocks the stream
 	for (const auto &track_it : tracks)
 	{
 		auto track = track_it.second;
 
 		// The map entry is the latest published version of this track; while a
 		// track stays undescribed no version has been published yet
-		if (track->IsValid() == false)
+		if ((track->IsValid() == false) ||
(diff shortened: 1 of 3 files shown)


src/mediarouter/mediarouter_stream.cpp
```diff
@@ -131,24 +131,37 @@ bool MediaRouteStream::IsStreamReady()
...
 		{
-			return false;
+			all_tracks_ready = false;
+			continue;
 		}
 
-		// If the track is an outbound track, it is necessary to check the quality.
-		if (_stream->HasTrackQualityMeasured(track->GetId()) == false)
+		// Record readiness on the shared stats; modules holding setup-time track
+		// versions (e.g. the provider) read this fact instead of judging validity
```
dimiden, line 152: nit: No code holding a setup-time track version reads `TrackStats::IsReady()`; its only consumer is `CheckUnpreparedTrackTimeout()` below, which uses the router's own up-to-date copy. Reword this to describe the actual reader, or land the provider-side consumer this describes.

dimiden (approved):
LGTM

PR #2319, opened 2026-09-01
fix(llhls): remove the segment overrun warning

getroot:
A video segment can only close at a keyframe, so it closes at the first keyframe at or after the pacing plan and comes out longer whenever the plan falls between two of them. That is the cadence working, not a fault, and the warning reported it on every segment for the life of the stream: a stream whose first segment ran long once carries that difference in the pacing ledger and never stops logging.

The warning is removed. A segment that finds no cuttable position for twice the configured duration is still force-closed and reported at error level, and a keyframe interval that does not divide the segment duration is still reported once when the stream is prepared, so nothing an operator can act on is lost. The keyframe interval is also judged after its conversion to microseconds, so a value that rounds away to nothing cannot divide the duration.

Test: publish a stream whose first segment runs longer than the configured duration; a WHIP source reproduces it, since its first frame arrives about two seconds before the rest and stretches the first GOP. Pass conditions: no per-segment warning while segment durations stay at the configured value, and a segment that finds no cuttable position for twice the configured duration still reports the existing error.

### DIFF
--- a/src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
+++ b/src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
@@ -38,9 +38,9 @@ namespace bmff
 		_cut_cadence_us = _segment_duration_us;
 		if (config.keyframe_interval_ms > 0)
 		{
-			int64_t keyframe_interval_us = static_cast<int64_t>(config.keyframe_interval_ms * 1000.0 + 0.5);
-			int64_t intervals = (_segment_duration_us + keyframe_interval_us - 1) / keyframe_interval_us;
-			_cut_cadence_us = std::max(_cut_cadence_us, intervals * keyframe_interval_us);
+			_keyframe_interval_us = static_cast<int64_t>(config.keyframe_interval_ms * 1000.0 + 0.5);
+			int64_t intervals = (_segment_duration_us + _keyframe_interval_us - 1) / _keyframe_interval_us;
+			_cut_cadence_us = std::max(_cut_cadence_us, intervals * _keyframe_interval_us);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/modules/containers/bmff/fmp4_packager/segment_boundary_policy.cpp
```diff
@@ -36,11 +36,14 @@ namespace bmff
...
-		if (config.keyframe_interval_ms > 0)
+		// Judged after the conversion: an interval that rounds away to nothing
+		// cannot divide the duration
+		int64_t keyframe_interval_us = static_cast<int64_t>(config.keyframe_interval_ms * 1000.0 + 0.5);
+		if (keyframe_interval_us > 0)
 		{
-			int64_t keyframe_interval_us = static_cast<int64_t>(config.keyframe_interval_ms * 1000.0 + 0.5);
-			int64_t intervals = (_segment_duration_us + keyframe_interval_us - 1) / keyframe_interval_us;
-			_cut_cadence_us = std::max(_cut_cadence_us, intervals * keyframe_interval_us);
+			_keyframe_interval_us = keyframe_interval_us;
```
Keukhan: On a bypass video track this change has no effect: `_keyframe_interval_us` stays zero, `reach_us` falls back to `next_frame.duration_us`, and the warning still fires on every segment.

The interval is read once, in `AddPackager`, before any media has arrived (llhls_stream.cpp:2107). At that point it can only come from an encoding profile, and a bypass track has none (transcoder_stream_internal.cpp:168-191). The measured value is still zero because no keyframes have been counted yet, and a `FRAME`-type interval additionally needs a frame rate, which a WHIP ingest never provides.

So the allowance takes effect when `KeyFrameIntervalType=TIME` is configured — likely how the repro was verified — but not on a bypass rendition of the same stream.

This is also why the new header comment reads wrong: zero here usually means "keyframe interval unknown", not "every frame can cut", and the two want opposite allowances.

The simplest way to make the premise hold everywhere is to skip the warning when a track cuts only at keyframes and its interval is unknown, rather than falling back to one frame. Giving those tracks a real allowance needs the interval observed from the samples instead of configured — the policy already sees `independent` on every packet.

getroot: I realized the warning had no real meaning, so it is removed. Segment overruns caused by keyframe placement are the cadence working, not a fault. The twice-the-duration force close and the prepare-time keyframe interval check still cover what matters.

Keukhan (approved):
lgtm

PR #2321, opened 2026-09-03
fix(bitstream): guard the SPS frame rate division and reject out-of-range ue(v) codes

dimiden:
## Summary

An H.264 SPS whose VUI `timing_info` carries `num_units_in_tick = 0x80000000` terminates the process with SIGFPE.
The frame rate computation guards `num_units_in_tick` against 0, but it divides by `2 * num_units_in_tick`, which wraps to a zero divisor in 32-bit unsigned arithmetic.
The division runs on the MediaRouter inbound worker, reached from `MediaRouterNormalize::ProcessH264AnnexBStream()` and `ProcessH264AVCCStream()` through `AVCDecoderConfigurationRecord::AddSPS()`, so a single SPS from any H.264 ingest takes the whole server down.
The defect dates back to f0665c532 and first shipped in v0.10.8; the `num_units_in_tick != 0` guard added later does not cover the overflow.

Review turned up more defects in the same parsing layer, and this PR takes them as well so the part is stabilized in one go.

### The reported crash

- `h264_parser.cpp`: widen `num_units_in_tick` to `uint64_t` before doubling it, so the divisor can no longer wrap to zero. `num_units_in_tick == 0` keeps yielding an fps of 0 as before.

### Shared bitstream reader

- `nal_unit_bitstream_parser.cpp`: reject an Exp-Golomb code with more than 31 leading zero bits, and compute the shift on an unsigned `1`. codeNum tops out at `2^32 - 2` with 31 leading zeros, which is the range every `ue(v)` element is defined over (Rec. ITU-T H.264 clause 9.1 sets no global bound, but `pan_scan_rect_id` and the other widest elements are capped there). The range check also removes the `int` to `uint8_t` narrowing that made a 256-zero code report success with a bogus value while leaving the bit position desynchronized.
- The length check runs inside the leading-zero loop, so a hostile NAL is refused after 32 bits instead of after the whole run. An 8 Mbit zero run cost 24 ms at `-O2` before and returns immediately now.
- `Skip()` consumes its count in 64 bit steps. It forwarded the count to `ReadBits()`, whose width parameter is a `uint8_t` capped at the width of its output, so `Skip(88)` always failed and an HEVC `profile_tier_level()` with `sub_layer_profile_present_flag[i] == 1` never parsed, while `Skip(256)` reported success after consuming nothing.
- Dropped the dead `else` that reassigned `value` to the 0 it already holds.

### Frame rate derivation

- `h264_parser.cpp`: derive the frame rate whenever `timing_info` is present. Rec. ITU-T H.264 E.2.1 defines the clock tick from `num_units_in_tick` and `time_scale` alone, and `fixed_frame_rate_flag` only constrains how evenly the output times are spaced. The flag is clear by default for x264 and openh264, so gating on it left `_fps` at 0 for most streams.
- `h265_parser.h`: return 0 from `H265SPS::GetFps()` when the tick count is 0, which is the case whenever the VUI carries no `timing_info`, and cast the numerator so the division does not truncate before the conversion to `float`. `60000 / 2002` reported 29.00 instead of 29.97.

### Tests

- New `h264_parser_test.cpp` covers the timing_info fields: a sane 30 fps SPS, `num_units_in_tick = 0x80000000` against several `time_scale` values, a zero tick count, and a derivable rate with `fixed_frame_rate_flag` clear.
- `nal_unit_bitstream_parser_test.cpp` covers the `ue(v)` boundary at `0xFFFFFFFE`, the rejected 32-zero length, long zero runs, and `Skip()` counts of 1 through 1024 asserting `BitsConsumed()`.
- `h265_parser_test.cpp` parses an SPS instead of default-constructing one for the fps guard, and adds a sub-layer `profile_tier_level` case that aborts on the pre-fix `Skip()` in a DEBUG build. `BuildSps()` takes an optional VUI so timing_info cases can be built.
- The `ue(v)` writers moved into `nalu/nal_unit_test_helpers.h`, and escaping now goes through `NalUnitInsertor::EmulationPreventionBytes()` instead of a per-suite copy. The two escapers were compared exhaustively over every sequence of length 1..7 from {00,01,02,03,04,FF} plus 200000 random sequences before the swap, with zero mismatches.

## Compatibility

Conformant bitstreams decode to the same values, with one intended change: an SPS that carries `timing_info` with `fixed_frame_rate_flag` clear now reports its derived frame rate instead of 0.
`H264SPS::GetFps()` and `H265SPS::GetFps()` are read only by `GetInfoString()` today, which nothing calls, so no live path changes behavior.

### DIFF
--- a/src/modules/bitstream/h264/h264_parser.cpp
+++ b/src/modules/bitstream/h264/h264_parser.cpp
@@ -644,9 +644,10 @@ bool H264Parser::ParseVUI(NalUnitBitstreamParser &parser, H264SPS &sps)
 
 				if (fixed_frame_rate_flag)
 				{
-					sps._fps = (num_units_in_tick != 0)
-								   ? (time_scale / (2 * num_units_in_tick))
-								   : 0;
+					// Doubling in 32-bit arithmetic turns `num_units_in_tick` `0x80000000` into a zero divisor.
+					const auto divisor = static_cast<uint64_t>(num_units_in_tick) * 2;
+
+					sps._fps		   = (divisor != 0) ? static_cast<unsigned int>(time_scale / divisor) : 0;
 				}
 			}
 		}
--- a/src/modules/bitstream/h264/h264_parser.cpp
+++ b/src/modules/bitstream/h264/h264_parser.cpp
@@ -622,9 +622,7 @@ bool H264Parser::ParseVUI(NalUnitBitstreamParser &parser, H264SPS &sps)
 
 			if (timing_info_present_flag)
 			{
-				uint32_t num_units_in_tick, time_scale;
-				uint8_t fixed_frame_rate_flag;
-
+				uint32_t num_units_in_tick;
 				// ITU-T H.264 E.1.1: both are u(32), not ue(v). Reading them as Exp-Golomb
 				// desynchronizes the bit position for everything that follows in the VUI.
 				if (!parser.ReadU32(num_units_in_tick))
@@ -632,23 +630,24 @@ bool H264Parser::ParseVUI(NalUnitBitstreamParser &parser, H264SPS &sps)
 					return false;
 				}
 
+				uint32_t time_scale;
 				if (!parser.ReadU32(time_scale))
 				{
 					return false;
 				}
 
+				uint8_t fixed_frame_rate_flag;
 				if (!parser.ReadBit(fixed_frame_rate_flag))
 				{
 					return false;
 				}
 
-				if (fixed_frame_rate_flag)
-				{
-					// Doubling in 32-bit arithmetic turns `num_units_in_tick` `0x80000000` into a zero divisor.
-					const auto divisor = static_cast<uint64_t>(num_units_in_tick) * 2;
+				// ITU-T H.264 E.2.1 derives the clock tick from `num_units_in_tick` and `time_scale` alone.
+				// `fixed_frame_rate_flag` only constrains how evenly the output times are spaced.
+				// Doubling in 32-bit arithmetic turns `num_units_in_tick` `0x80000000` into a zero divisor.
+				const auto divisor = static_cast<uint64_t>(num_units_in_tick) * 2;
 
-					sps._fps		   = (divisor != 0) ? static_cast<unsigned int>(time_scale / divisor) : 0;
-				}
+				sps._fps		   = (divisor != 0) ? static_cast<unsigned int>(time_scale / divisor) : 0;
 			}
 		}
 
--- a/src/modules/bitstream/h264/h264_parser_test.cpp
+++ b/src/modules/bitstream/h264/h264_parser_test.cpp
@@ -1,0 +1,165 @@
+//==============================================================================
+//
+//  OvenMediaEngine
+//
+//  Created by Hyunjun Jang
+//  Copyright (c) 2026 OvenMediaLabs. All rights reserved.
+//
+//==============================================================================
+
+//  Covers: H264Parser VUI timing_info handling, in particular the frame rate
+//          division when num_units_in_tick sits at the edge of u(32).
+//
+//  Bitstreams are hand-built with a BitWriter so the field values under test
+//  are known by construction. Syntax follows ITU-T H.264 (7.3.2.1.1 SPS,
+//  E.1.1 vui_parameters).
+
+// Unit tests
+// ----------
+// cmake -S . -B __temp_build/tests -G Ninja -DOME_BUILD_TESTS=ON
+// ninja -C __temp_build/tests ome_test_modules
+// ./__temp_build/tests/bin/ome_test_modules --gtest_filter='H264Parser.*'
+
+#include <base/ovlibrary/bit_writer.h>
+#include <gtest/gtest.h>
+#include <modules/bitstream/h264/h264_parser.h>
+
+#include <vector>
+
+namespace
+{
+	// ue(v) (ITU-T H.264 9.1)
+	void WriteUE(ov::BitWriter &w, uint32_t value)
+	{
+		uint64_t code_num = static_cast<uint64_t>(value) + 1;
+		int numbits		  = 0;
+		while ((code_num >> (numbits + 1)) != 0)
+		{
+			numbits++;
+		}
+
+		if (numbits > 0)
+		{
+			w.WriteBits(numbits, 0);
+		}
+		w.WriteBits(numbits + 1, code_num);
+	}
+
+	// rbsp_trailing_bits(): stop bit '1' then zero-pad to a byte boundary
+	void WriteTrailing(ov::BitWriter &w)
+	{
+		w.WriteBits(1, 1);
+		while ((w.GetBitCount() % 8) != 0)
+		{
+			w.WriteBits(1, 0);
+		}
+	}
+
+	std::vector<uint8_t> ApplyEmulationPrevention(const std::vector<uint8_t> &rbsp)
+	{
+		std::vector<uint8_t> out;
+		out.reserve(rbsp.size() + 4);
+
+		size_t zeros = 0;
+		for (uint8_t byte : rbsp)
+		{
+			if ((zeros >= 2) && (byte <= 0x03))
+			{
+				out.push_back(0x03);
+				zeros = 0;
+			}
+
+			out.push_back(byte);
+			zeros = (byte == 0x00) ? (zeros + 1) : 0;
+		}
+
+		return out;
+	}
+
+	// A 640x320 Baseline SPS whose VUI carries the given timing_info
+	std::vector<uint8_t> MakeSpsNal(uint32_t num_units_in_tick, uint32_t time_scale)
+	{
+		ov::BitWriter w(64);
+
+		w.WriteBits(8, 66);	 // profile_idc: Baseline, so no chroma_format_idc block
+		w.WriteBits(8, 0);	 // constraint flags + reserved
+		w.WriteBits(8, 31);	 // level_idc
+		WriteUE(w, 0);		 // seq_parameter_set_id
+		WriteUE(w, 4);		 // log2_max_frame_num_minus4
+		WriteUE(w, 0);		 // pic_order_cnt_type
+		WriteUE(w, 4);		 // log2_max_pic_order_cnt_lsb_minus4
+		WriteUE(w, 1);		 // max_num_ref_frames
+		w.WriteBits(1, 0);	 // gaps_in_frame_num_value_allowed_flag
+		WriteUE(w, 39);		 // pic_width_in_mbs_minus1
+		WriteUE(w, 19);		 // pic_height_in_map_units_minus1
+		w.WriteBits(1, 1);	 // frame_mbs_only_flag
+		w.WriteBits(1, 1);	 // direct_8x8_inference_flag
+		w.WriteBits(1, 0);	 // frame_cropping_flag
+
+		w.WriteBits(1, 1);	// vui_parameters_present_flag
+		w.WriteBits(1, 0);	// aspect_ratio_info_present_flag
+		w.WriteBits(1, 0);	// overscan_info_present_flag
+		w.WriteBits(1, 0);	// video_signal_type_present_flag
+		w.WriteBits(1, 0);	// chroma_loc_info_present_flag
+
+		w.WriteBits(1, 1);					 // timing_info_present_flag
+		w.WriteBits(32, num_units_in_tick);	 // num_units_in_tick, u(32)
+		w.WriteBits(32, time_scale);		 // time_scale, u(32)
+		w.WriteBits(1, 1);					 // fixed_frame_rate_flag
+
+		w.WriteBits(1, 0);	// nal_hrd_parameters_present_flag
+		w.WriteBits(1, 0);	// vcl_hrd_parameters_present_flag
+		w.WriteBits(1, 0);	// pic_struct_present_flag
+		w.WriteBits(1, 0);	// bitstream_restriction_flag
+
+		WriteTrailing(w);
+
+		const std::vector<uint8_t> rbsp(w.GetData(), w.GetData() + w.GetDataSize());
+
+		std::vector<uint8_t> nal;
+		nal.push_back(0x67);  // forbidden_zero_bit(0) + nal_ref_idc(3) + nal_unit_type(7, SPS)
+
+		const auto ebsp = ApplyEmulationPrevention(rbsp);
+		nal.insert(nal.end(), ebsp.begin(), ebsp.end());
+
+		return nal;
+	}
+}  // namespace
+
+TEST(H264Parser, DerivesTheFrameRateFromTimingInfo)
+{
+	const auto nal = MakeSpsNal(1, 60);
+
+	H264SPS sps;
+	ASSERT_TRUE(H264Parser::ParseSPS(nal.data(), nal.size(), sps));
+
+	EXPECT_EQ(sps.GetWidth(), 640U);
+	EXPECT_EQ(sps.GetFps(), 30U);
+}
+
+// Doubling 0x80000000 in 32-bit arithmetic wraps the divisor to 0, which used to raise SIGFPE and
+// terminate the process. The value is legal u(32) syntax, so the parser has to survive it.
+TEST(H264Parser, SurvivesANumUnitsInTickThatWrapsTheDoubledDivisor)
+{
+	for (uint32_t time_scale : {0U, 1U, 53U, 60U, 0xFFFFFFFFU})
+	{
+		const auto nal = MakeSpsNal(0x80000000U, time_scale);
+
+		H264SPS sps;
+		ASSERT_TRUE(H264Parser::ParseSPS(nal.data(), nal.size(), sps))
+			<< "time_scale: " << time_scale;
+
+		// Any u(32) time_scale over a divisor of 2^32 truncates to 0
+		EXPECT_EQ(sps.GetFps(), 0U) << "time_scale: " << time_scale;
+	}
+}
+
+TEST(H264Parser, ReportsZeroFpsWhenNumUnitsInTickIsZero)
+{
+	const auto nal = MakeSpsNal(0, 60);
+
+	H264SPS sps;
+	ASSERT_TRUE(H264Parser::ParseSPS(nal.data(), nal.size(), sps));
+
+	EXPECT_EQ(sps.GetFps(), 0U);
+}
--- a/src/modules/bitstream/h264/h264_parser_test.cpp
+++ b/src/modules/bitstream/h264/h264_parser_test.cpp
@@ -77,7 +77,8 @@ namespace
 	}
 
 	// A 640x320 Baseline SPS whose VUI carries the given timing_info
-	std::vector<uint8_t> MakeSpsNal(uint32_t num_units_in_tick, uint32_t time_scale)
+	std::vector<uint8_t> MakeSpsNal(uint32_t num_units_in_tick, uint32_t time_scale,
+									uint8_t fixed_frame_rate_flag = 1)
 	{
 		ov::BitWriter w(64);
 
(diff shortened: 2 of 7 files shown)


src/modules/bitstream/h265/h265_parser.h
```diff
@@ -159,7 +159,13 @@ class H265SPS
 
     float GetFps() const
     {
-        return _vui_parameters._time_scale / _vui_parameters._num_units_in_tick;
+		// The tick count stays 0 when the SPS carries no VUI
+		if (_vui_parameters._num_units_in_tick == 0)
+		{
+			return 0.0f;
+		}
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/modules/bitstream/nalu/nal_unit_bitstream_parser.cpp
```diff
@@ -44,15 +44,22 @@ bool NalUnitBitstreamParser::ReadUEV(uint32_t &value)
...
-        if (ReadBits(zero_bit_count, rest) == false)
-        {
-            return false;
-        }
+		// 31 leading zeros already span the whole `uint32_t` range.
+		// A longer code cannot be represented, and the shift itself would be undefined.
+		if (zero_bit_count > 31)
+		{
+			return false;
+		}
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/modules/bitstream/nalu/nal_unit_bitstream_parser_test.cpp
```diff
@@ -2,12 +2,56 @@
...
+		{
+			write_bit(0);
+		}
+
+		write_bit(1);
+
+		for (int i = leading_zero_bits - 1; i >= 0; i--)
+		{
+			write_bit(static_cast<int>((rest >> i) & 1));
+		}
```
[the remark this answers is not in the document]
dimiden: Fixed.

src/modules/bitstream/nalu/nal_unit_bitstream_parser_test.cpp
```diff
@@ -36,3 +82,39 @@ TEST(NalUnitBitstreamParser, KeepsATrailingEmulationPreventionByte)
...
+
+// 32 leading zeros put codeNum at 2^32 - 1 + rest, past what the caller can hold
+TEST(NalUnitBitstreamParser, RejectsAnExpGolombCodeOutOfTheUint32Range)
+{
+	const auto bitstream = MakeUevBitstream(32, 1);
+	NalUnitBitstreamParser parser(bitstream.data(), bitstream.size());
+
+	uint32_t value = 0;
+	EXPECT_FALSE(parser.ReadUEV(value));
+}
```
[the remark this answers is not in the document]
dimiden: Checked against Rec. ITU-T H.264 (01/2012). Clause 9.1 sets no global bound on codeNum, but the widest `ue(v)` elements are capped at `2^32 - 2` in their own semantics -- for example `pan_scan_rect_id`: "shall be in the range of 0 to 2^32 - 2, inclusive" (also `scene_id`, `second_scene_id`, `snapshot_id`, `progressive_refinement_id`, `tone_map_id`). No syntax element in the spec is defined up to `2^32 - 1`, so rejecting the whole 32-zero length loses nothing conformant, and `0xFFFFFFFE` is the right boundary to assert.

Fixed.

src/modules/bitstream/nalu/nal_unit_bitstream_parser.cpp
```diff
@@ -44,15 +44,23 @@ bool NalUnitBitstreamParser::ReadUEV(uint32_t &value)
...
 
-        value = (1 << zero_bit_count) - 1 + rest;
-    }
-    else
+		uint32_t rest;
+		if (ReadBits(static_cast<uint8_t>(zero_bit_count), rest) == false)
```
getroot, line 56: The same `uint32_t` -> `uint8_t` narrowing that this line removes is still present in `Skip()` at line 87, and there it is already live:

```cpp
bool NalUnitBitstreamParser::Skip(uint32_t count)
{
    uint64_t dummy;
	return ReadBits(count, dummy);   // BitReader::ReadBits(uint8_t bits, T &value)
}
```

`h265_parser.cpp:2051` calls `parser.Skip(88)`, which is taken whenever an HEVC `profile_tier_level()` carries `sub_layer_profile_present_flag[i] == 1`. 88 survives the narrowing but then fails `bits > sizeof(value) * 8` in `ReadBitsInternal` (the dummy is a `uint64_t`, so the limit is 64). That is `OV_ASSERT2(false)` -> `::abort()` in a DEBUG build, and a plain `false` in release, so the VPS/SPS is rejected and the stream never starts.

`Skip(256)` is the other half of the same defect: it narrows to 0, `ReadBitsInternal` returns true on the `bits == 0` path without advancing, and the caller keeps parsing 256 bits behind where it thinks it is. That is exactly the silent desync this PR fixes in `ReadUEV()`.

src/modules/bitstream/nalu/nal_unit_bitstream_parser.cpp
```diff
@@ -44,15 +44,23 @@ bool NalUnitBitstreamParser::ReadUEV(uint32_t &value)
...
     {
-        uint32_t rest;
-        if (ReadBits(zero_bit_count, rest) == false)
-        {
-            return false;
-        }
+		// codeNum tops out at 2^32 - 2 with 31 leading zeros, which is the range every ue(v)
+		// syntax element is defined over. A longer code is out of range, and the shift below
+		// would be undefined.
+		if (zero_bit_count > 31)
```
getroot: This check runs only after the leading-zero loop has consumed the entire run, so a hostile NAL still costs one `ReadBit()` per zero bit before it is rejected. Every iteration is a `ReadBits<uint8_t>(1, ...)` plus the virtual `NextPosition()` with its emulation-prevention test, so an SPS carrying 8 Mbit of `0x00` before its first `1` bit burns on the order of 0.1 s on the MediaRouter inbound worker for a single NAL. Repeating that starves the worker.

Testing inside the loop bounds the work at 32 bits:

```cpp
if (bit == 0)
{
    if (++zero_bit_count > 31)
    {
        return false;
    }
}
```

That also covers the second half of the problem: `zero_bit_count` is a plain `int` incremented with no bound, which is undefined behavior at 2^31 zero bits.

src/modules/bitstream/nalu/nal_unit_bitstream_parser.cpp
```diff
@@ -44,15 +44,23 @@ bool NalUnitBitstreamParser::ReadUEV(uint32_t &value)
...
-    else
+		uint32_t rest;
+		if (ReadBits(static_cast<uint8_t>(zero_bit_count), rest) == false)
+		{
+			return false;
+		}
+
+		value = (1U << zero_bit_count) - 1 + rest;
+	}
+	else
```
getroot: This `else` is dead. `value` is already assigned 0 at the top of the function (line 25), so the branch cannot change anything. Removing it leaves `if (zero_bit_count > 0)` as the only writer and drops the second place that states the out parameter's default.

src/modules/bitstream/h264/h264_parser.cpp
```diff
@@ -644,9 +644,10 @@ bool H264Parser::ParseVUI(NalUnitBitstreamParser &parser, H264SPS &sps)
 
 				if (fixed_frame_rate_flag)
```
getroot: Since the expression below is being rewritten anyway: this `if (fixed_frame_rate_flag)` gate discards a frame rate that is fully derivable. An SPS with `timing_info_present_flag = 1`, `num_units_in_tick = 1001`, `time_scale = 60000`, `fixed_frame_rate_flag = 0` parses without error, but `_fps` stays 0 and `src/providers/rtsp/sdp.cpp:192` then publishes `SetFrameRate(0)`.

H.264 E.2.1 derives the clock tick from `num_units_in_tick / time_scale` regardless of that flag. The flag only asserts that output times are evenly spaced, and ffmpeg's h264 parser likewise ignores it when setting the frame rate. Encoders that will not promise constant spacing are common, so the `_fps` assignment belongs outside the flag test.

src/modules/bitstream/h265/h265_parser.h
```diff
@@ -159,7 +159,13 @@ class H265SPS
...
     float GetFps() const
     {
-        return _vui_parameters._time_scale / _vui_parameters._num_units_in_tick;
+		// `timing_info` is optional, so the tick count can still be 0 here
+		if (_vui_parameters._num_units_in_tick == 0)
+		{
+			return 0.0f;
+		}
+
+		return _vui_parameters._time_scale / _vui_parameters._num_units_in_tick;
```
getroot: Both operands are `uint32_t`, so the division truncates before the conversion to `float`. `vui_time_scale = 60000` over `vui_num_units_in_tick = 2002` returns 29.0f, and `GetInfoString()` (line 204, `"Fps(%.2f)"`) prints `29.00` instead of `29.97`. While this statement is already being touched, one cast on the numerator fixes it:

```cpp
return static_cast<float>(_vui_parameters._time_scale) / _vui_parameters._num_units_in_tick;
```

src/modules/bitstream/h265/h265_parser_test.cpp
```diff
@@ -1183,3 +1183,13 @@ TEST(H265Parser, SliceHeaderRejectsExcessiveOffsetLen)
 	slice = BuildWppIdrSlice(/*num_entry_point_offsets=*/2, /*offset_len_minus1=*/H265_MAX_OFFSET_LEN_MINUS1 + 1);
 	EXPECT_FALSE(H265Parser::ParseSliceHeader(slice.data(), slice.size(), shd, record));
 }
+
+// An SPS with no VUI leaves num_units_in_tick at 0, and the frame rate is a plain integer
+// division by it
+TEST(H265Parser, ReportsZeroFpsWhenTheSpsCarriesNoVui)
```
getroot, line 1199: This test default-constructs `H265SPS` rather than parsing one, so it pins only the member initializer `_num_units_in_tick = 0` at `h265_parser.h:71`, not the behavior the name describes. If `ProcessVuiParameters()` later left a stale non-zero tick count behind, or if an SPS set `vui_timing_info_present_flag = 1` with `vui_num_units_in_tick = 0` (assigned unvalidated at `h265_parser.cpp:2251`), this test still passes while `GetFps()` divides by zero.

`BuildSps()` in this file already writes `vui_parameters_present_flag = 0` (line 244), so the no-VUI case is one `ParseSPS()` call away. The zero-tick timing_info case needs a small extension to that helper, and it is the case the guard actually protects.

src/modules/bitstream/h264/h264_parser_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include <base/ovlibrary/bit_writer.h>
+#include <gtest/gtest.h>
+#include <modules/bitstream/h264/h264_parser.h>
+
+#include <vector>
+
+namespace
+{
+	// ue(v) (ITU-T H.264 9.1)
+	void WriteUE(ov::BitWriter &w, uint32_t value)
```
getroot: `WriteUE()`, `WriteTrailing()` and `ApplyEmulationPrevention()` are already defined in `src/modules/bitstream/h265/h265_parser_test.cpp` (lines 33, 56, 71), and both files link into the same `ome_test_modules` target. A fix to the Exp-Golomb writer or to the emulation-prevention rule now has to land twice, or the two suites silently disagree about what bytes a NAL contains.

For the escaping specifically there is already a production implementation, `NalUnitInsertor::EmulationPreventionBytes()` (`nal_unit_insertor.cpp:227`), which `h264_sps_rewriter_test.cpp:94` and `h264_sei_inserter_test.cpp:416` both call from tests. Moving the three helpers into a shared test header, and using the production escaper for the third, keeps one definition of each.

getroot (commented):
A few notes on the two fixes and the new tests. The `ReadUEV()` narrowing is fixed here but the identical one in `Skip()` is still live on the HEVC `profile_tier_level()` path, and the new range check is placed after the zero run has already been consumed.

src/modules/bitstream/nalu/nal_unit_bitstream_parser.cpp
dimiden: Confirmed. `Skip(88)` returned false with 0 bits consumed, so an HEVC `profile_tier_level()` with `sub_layer_profile_present_flag[i] == 1` never parsed, and `Skip(256)` returned true with 0 bits consumed. Not from this PR, but same function and same defect class, so taking it here to stabilize the part in one go.

`Skip()` now consumes the count in 64 bit steps. Added `Skip()` coverage for 1/8/64/65/88/255/256/264/1024 asserting `BitsConsumed()`, plus an HEVC SPS with a sub-layer profile_tier_level that aborts on the pre-fix parser in a DEBUG build.

src/modules/bitstream/nalu/nal_unit_bitstream_parser.cpp
dimiden: Measured at `-O2` with an 8 Mbit zero run in one NAL: before, `ReadUEV()` consumed `8388609` bits and took 24 ms; after, it consumes 32 bits and returns immediately. Same `false` either way.

Moved the check into the loop as suggested and dropped the now-redundant one after it. That also removes the unbounded `int` increment, since the count never gets past 32.

src/modules/bitstream/nalu/nal_unit_bitstream_parser.cpp
dimiden: Right, dead since 154a4495c (2023-06): the top-level `value = 0` came in first with 1d99c9db9, then that commit wrapped the tail in `if/else` and repeated the assignment. Removed.

src/modules/bitstream/h264/h264_parser.cpp
dimiden: Agreed on the spec point, confirmed against Rec. ITU-T H.264 (01/2012) E.2.1: the clock tick is defined by `num_units_in_tick` and `time_scale`, and `fixed_frame_rate_flag` only constrains the spacing of the HRD output times. Moved the assignment out of the gate, with a regression test for `1001/60000` and the flag clear.

One correction on the impact: `src/providers/rtsp/sdp.cpp` is not in the build. `src/providers/CMakeLists.txt` adds `rtspc` but not `rtsp`, that directory has no CMakeLists, and no `sdp.cpp.o` is produced. Its `payload.SetFrameRate()` call would not compile either, since `MediaTrack` has only `SetFrameRateByConfig()`. So the only live reader of `GetFps()` today is `GetInfoString()`, which nothing calls.

Taking it anyway: the value is wrong for most streams, since the flag is clear by default for x264 and openh264, and the first caller that wires `GetFps()` to a track would get 0.

src/modules/bitstream/h265/h265_parser.h
dimiden: Fixed with the cast. Added `H265Parser.DerivesANonIntegerFrameRateFromTheVui` for 2002/60000, which reports 29.00 without it.

`BuildSps()` now takes an optional `VuiTimingInfo` so a VUI with timing_info can be built. The zero-tick case you raised separately reuses it.

Line reference has shifted to 168 since the guard went in above it.

src/modules/bitstream/h265/h265_parser_test.cpp
dimiden: Right, that test only pinned the member initializer.
Replaced it with two that go through `ParseSPS()`: one with no VUI, and one with `vui_timing_info_present_flag = 1` and `vui_num_units_in_tick = 0`, which is the case the guard protects since `h265_parser.cpp:2251` stores the value without a range check.

The helper extension landed with the truncation fix, so `BuildSps()` already takes an optional `VuiTimingInfo`.

Line references have shifted: the test is at 1249 now, and `vui_parameters_present_flag` is written conditionally.

src/modules/bitstream/h264/h264_parser_test.cpp
dimiden: Moved `WriteUE()`, `WriteSE()`, `WriteTrailing()` and `ToBytes()` into `nalu/nal_unit_test_helpers.h`, and both suites now use it.
`ApplyEmulationPrevention()` in that header is a thin wrapper over `NalUnitInsertor::EmulationPreventionBytes()`, so there is one escaper again.

Checked the two escapers agree before swapping: exhaustive over every sequence of length 1..7 from {00,01,02,03,04,FF} (335922 cases) plus 200000 random sequences weighted 60% zero bytes, zero mismatches.
The only textual difference was `zero_count == 2` against `zeros >= 2`, and neither counter can pass 2 because both reset after escaping.

src/modules/bitstream/h265/h265_parser.h
```diff
@@ -159,7 +159,13 @@ class H265SPS
...
     {
-        return _vui_parameters._time_scale / _vui_parameters._num_units_in_tick;
+		// `timing_info` is optional, so the tick count can still be 0 here
+		if (_vui_parameters._num_units_in_tick == 0)
+		{
+			return 0.0f;
+		}
+
+		return static_cast<float>(_vui_parameters._time_scale) / _vui_parameters._num_units_in_tick;
     }
```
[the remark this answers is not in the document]
dimiden: Right, and the body was stale beyond that. Rewrote it for the eight commits added during review.

Three more things in it were wrong: it said `GetFps()` is reachable through `H265SPS::ToString()`, which does not exist (it is `GetInfoString()`); it claimed 31 leading zeros span the whole `uint32_t` range, which the code comment has since corrected to `2^32 - 2`; and it claimed conformant streams keep the same fps, which stopped being true once the `fixed_frame_rate_flag` gate came out.

src/modules/bitstream/h264/h264_parser.cpp
```diff
@@ -622,32 +622,32 @@ bool H264Parser::ParseVUI(NalUnitBitstreamParser &parser, H264SPS &sps)
...
 				if (!parser.ReadU32(time_scale))
 				{
 					return false;
 				}
 
+				uint8_t fixed_frame_rate_flag;
 				if (!parser.ReadBit(fixed_frame_rate_flag))
 				{
 					return false;
 				}
```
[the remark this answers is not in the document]
dimiden: No warning here. `ReadBit()` takes the flag by non-const reference, so the compiler counts it as used: `-Wall -Wextra -Wunused -Wunused-variable -Wunused-but-set-variable` on both clang++ and g++ report nothing for `fixed_frame_rate_flag`.

`-Werror` does break this file, but on pre-existing findings unrelated to this PR: `unused parameter 'header'` at h264_parser.cpp:1312 and several in the ovlibrary headers. The project does not build with `-Werror`.

Leaving the read as it is. The bit still has to be consumed to keep the position in sync, which is what the surrounding fields do too.

src/modules/bitstream/nalu/nal_unit_test_helpers.h
```diff
@@ -0,0 +1,80 @@
...
+//  OvenMediaEngine
+//
+//  Created by Hyunjun Jang
+//  Copyright (c) 2026 OvenMediaLabs. All rights reserved.
+//
+//==============================================================================
+
+//  Bitstream writers shared by the parser test suites, so that every suite
+//  agrees on what bytes a NAL unit holds. Escaping is not duplicated here:
+//  tests call `NalUnitInsertor::EmulationPreventionBytes()` directly.
```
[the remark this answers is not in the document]
dimiden: Fixed. The comment was written before the wrapper existed and I did not update it when the `std::vector<uint8_t>` signature made one necessary.

dimiden:
Reply to Copilot's comment:

Checked this one and it does not hold.

`nullptr + 0` is well-defined in C++17. `[expr.add]` was amended by P0137R1 to say the result is a null pointer value, so the "even when the range length is 0" premise does not apply at the standard this project builds with. Ran exactly that code under `-fsanitize=undefined,address,pointer-overflow` at `-std=c++17` on both clang++ and g++: no diagnostics. `pointer-overflow` fires on a non-zero offset from null, not on zero. The `std::vector` range constructor is fine too, since `first == last`.

Execution also never reaches that line with an empty RBSP. `std::vector<uint8_t>::data()` is null when empty, and `ov::Data`'s constructor rejects a null pointer at data.cpp:29-34 with `OV_ASSERT2(false); return;`. So the real constraint on an empty RBSP is the `ov::Data` contract, not the pointer arithmetic.

And no caller passes one: both suites always write `rbsp_trailing_bits()`, so the RBSP is never empty.

Leaving it as is. A guard would be harmless but would encode a misreading of C++17 while still hiding the `ov::Data` constraint.

getroot (approved):
LGTM

